PR #31, opened 2026-08-25
fix: wrong-results bugs in semantic cache, merge join, and COUNT(col)

semihalev:
Fixes three verified wrong-results bug classes, all pre-existing on main and shipping in v0.4.0.

## Semantic cache (`src/executor/semantic_cache.rs`)

**String IN lists compared as identical.** `extract_in_values` dropped non-numeric literals, so `col IN ('a','b')` and `col IN ('x','y')` both extracted to empty numeric sets, compared equal, and served each other's cached results. Extraction now bails on any non-numeric element so the cache falls back to normal execution.

**Boundary operator confusion.** A cached `col > v` answered a new `col >= v` as Identical (and `col < v` answered `col <= v`), silently dropping the boundary row from results. The combined match arms are split; equal-value cross-operator pairs no longer subsume.

## Merge join (`src/executor/operators/merge_join.rs`)

Both-NULL join keys compared `Equal` (an ordering artifact) and produced matches, so `NULL = NULL` joined rows. Every other join operator rejects NULL keys. The Equal arm now treats NULL-keyed rows as unmatched; LEFT/FULL OUTER still emit their null-padded rows, and unmatched-right bookkeeping is unaffected.

## Aggregation (`src/executor/aggregation.rs`)

`COUNT(col)` mapped to the same fast-path accumulator as `COUNT(*)` and counted NULLs (the enum comment even documented the conflation). User-visible through inline HAVING, e.g. `HAVING COUNT(nullable_col) = 0` never matched all-NULL groups. `COUNT(col)` now maps to a null-aware `CountColumn` variant across all fast-path accumulation, HAVING, and emission sites; `COUNT(*)` is unchanged.

## Verification

- Each regression test was run against the unfixed code and confirmed to fail, then confirmed to pass with the fix (unit tests for cache and join, integration tests for COUNT).
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`, and the affected test targets are green.

One reviewed finding was deliberately excluded: the hot-only `get_index_on_column` guard after seal. Extensive empirical probing (IN lists, IN subqueries, GROUP BY on an indexed column over 2000 sealed rows) could not reproduce a wrong result on main, so the consumers evidently fall back correctly; adding the guard without a repro would only cost post-seal index performance. It stays on the investigation list.

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -117,11 +117,12 @@ impl SimpleHavingFilter {
 /// Supports COUNT, SUM, AVG, MIN, MAX (no DISTINCT, FILTER, ORDER BY, or expressions)
 #[derive(Clone)]
 enum SimpleAgg {
-    Count,      // COUNT(*) or COUNT(col)
-    Sum(usize), // SUM(col) - stores column index
-    Avg(usize), // AVG(col) - stores column index
-    Min(usize), // MIN(col) - stores column index
-    Max(usize), // MAX(col) - stores column index
+    Count,              // COUNT(*)
+    CountColumn(usize), // COUNT(col) - counts only non-NULL values
+    Sum(usize),         // SUM(col) - stores column index
+    Avg(usize),         // AVG(col) - stores column index
+    Min(usize),         // MIN(col) - stores column index
+    Max(usize),         // MAX(col) - stores column index
 }
 
 /// Try to parse a simple HAVING clause for inline filtering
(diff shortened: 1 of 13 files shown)


semihalev:
Independent adversarial review (fresh-context agent, instructed to refute) found real issues; addressed in 264ccc8:

- **New regression caught**: after the CountColumn split, `HAVING COUNT(*)` bound to the first COUNT aggregate regardless of column and read the COUNT(col) slot, silently dropping result rows. Fixed: COUNT(*) binds only to a COUNT(*) aggregate. Regression test verified to fail against 90a5fb3.
- `COUNT(DISTINCT col)` could stream through the derived-table fast path with no dedup; DISTINCT now bails that path (the other two fast paths already dedup).
- The index-driven streaming GROUP BY path mapped COUNT(col) to a row counter; COUNT(col)/DISTINCT now bail it (defensive: current planning intercepts those shapes earlier, but the path could not have answered them correctly).
- IN-list subsumption ignored the NOT flag (IN cache could answer NOT IN and vice versa) and treated equal-length duplicate lists as identical; both fixed with tests.
- Pre-existing FULL/RIGHT merge-join bug: the left-exhaustion path started the unmatched-right scan at right_idx, dropping right rows passed over before a match; now scans from 0 (matched rows excluded via right_matched). Test verified to fail against the previous commit.

Full gate (clippy + both suites) re-running on the branch head.

PR #32, opened 2026-08-25
fix: fail loud on swallowed durability errors (rename tombstones, PK cleanup, manifest load)

semihalev:
Fixes three classes of swallowed errors on the storage write/open paths. **One of these is a deliberate semantic change that deserves explicit review — see the last section.**

## rename_table: failed tombstone move reported success

Both `rename_table` implementations moved the legacy `snapshots/<t>/tombstones.dat` with `let _ = fs::rename(...)` after the volume-dir rename. On failure the rename still returned `Ok`, and a legacy-format database would resurrect its deleted rows on the next restart (modern manifests carry tombstones inside `volumes/<t>/manifest.bin` and move atomically with the volume dir, so this is legacy-format exposure only). The tombstone move now happens first (its revert is a plain rename-back), and any disk failure reverts the full in-memory rename and returns an error, matching the function's existing all-or-nothing contract.

## Cold-row DELETE: discarded hot PK index cleanup

Two DML paths did `let _ = self.hot.delete_by_row_ids(&[row_id])` while tombstoning a cold row. A failure left the hot PK index pointing at a tombstoned row: the next INSERT reusing that primary key hits a permanent duplicate-key error with nothing visible to repair. The result is now propagated (the adjacent `try_claim_row` on the same path always was).

## Open: warn-and-skip on unreadable manifests — SEMANTIC CHANGE

`load_manifests_from_volumes` logged a warning and skipped any table whose `manifest.bin` failed to load (corrupt, truncated, or written by a newer format version). The database then opened with that table silently missing — and, because the table's segment manager is empty, the orphan reaper treats the table's intact `.vol` files as garbage and deletes them. This is not hypothetical: pointing a v0.4.0 binary at a directory with a newer-format manifest was observed to permanently destroy the entire database (all `.vol` files unlinked, manifests rewritten as empty) within one open+checkpoint+close cycle.

The function now returns an error and the open fails closed. Four existing `durability_test` cases that codified the old contract ("corrupt manifest: open anyway, count >= 0") are updated to assert the open fails. The trade-off: genuine single-table corruption now blocks the open entirely instead of degrading; given the reaper interaction makes the degraded mode actively destructive, failing closed is the safer default. If a rescue mode is wanted (explicit flag/PRAGMA to open degraded with the reaper disabled), that can be a follow-up.

This change should also ship in a v0.4.1 patch release: it is the protection that makes future on-disk format bumps safe for users who roll back binaries.

## Verification

- Both new regression tests verified to fail against the unfixed code, then pass with it: a blocked tombstone move makes `ALTER TABLE RENAME` error and fully revert (table intact before and after reopen), and a manifest with an unsupported version byte makes `Database::open` fail with an error naming the manifest.
- `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`, and both full suites green (4870/4870 default and test-filedb).


semihalev:
Independent adversarial review (fresh-context agent) re-examined all three fixes; addressed in 0b577d2:

- **Tombstone-move fix withdrawn as a phantom**: the review established `snapshots/<t>/tombstones.dat` is never written or read anywhere in this codebase (a Go-era leftover); real tombstones live inside `volumes/<t>/manifest.bin` and already move atomically with the volume-dir rename. Both rename_table implementations are restored to their previous shape and the fabricated test is removed. (The dead tombstones.dat references in rename_table and record_truncate_table remain in place; flagging them here rather than deleting unrelated dead code.)
- The `delete_by_row_ids` propagation passed review clean: the callee's only fallible ops are same-txn-idempotent claims already guarded one line above.
- The fail-closed manifest change passed review with its premise **strengthened**: with warn-and-skip, `load_standalone_volumes_no_schema_check` deletes every intact .vol file not present in the freshly created empty manifest. Review confirmed the error path is clean (no half-initialized engine, no lock/thread leak, checkpoint.meta untouched). Per review suggestions, the error message now names the recovery path (--reset-volumes --restore) and the four updated durability tests assert the manifest exists before corrupting it.

semihalev:
P1 addressed (missing manifest still deletes live volumes): the standalone-volume loader now leaves any manifest-less table directory completely untouched - no empty segment manager, no orphan reaping. The empirical repro (volumes deleted on open after removing manifest.bin) fails against the previous commit and passes now. One deliberate deviation from the suggested strict fail-closed: a missing manifest also legitimately occurs when a crash lands between the first volume write and the first manifest write, and WAL replay covers that case; failing the open there would brick plain crash recovery. The protection is delete-never rather than open-never. If strict fail-closed is preferred regardless, it is a two-line change. Full gate green (4869/4869 both suites).

PR #33, opened 2026-08-25
chore: address clippy 1.98 lints

semihalev:
CI's Lint job installs the latest stable toolchain, now 1.98.0, whose clippy flags 21 pre-existing sites across 10 files (question_mark, chunks_exact-with-constant, manual slice fill, map values iteration). Every open PR fails Lint on these regardless of its own changes.

All fixes are mechanical rewrites with no behavior change, following the repo's precedent from the 1.95 lint pass. clippy --all-targets --all-features -D warnings is clean on 1.98, and both full suites pass (4868/4868 default and test-filedb).

Merge this first; the open fix PRs carry this commit and their diffs shrink back to their own changes once it lands.

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -4804,8 +4804,11 @@ impl Executor {
         let mut has_float = false;
         let mut has_value = false;
 
-        // Unroll loop by 4 for better CPU pipelining
-        let (chunks, remainder) = rows.as_chunks::<4>();
+        // Unroll loop by 4 for better CPU pipelining.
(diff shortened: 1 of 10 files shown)


semihalev:
P1 addressed (MSRV break): as_chunks reverted to chunks_exact with a targeted #[allow(clippy::chunks_exact_to_as_chunks)] and an MSRV note; the documented Rust 1.70 floor holds. Open question: add rust-version = "1.70" to Cargo.toml (and optionally a 1.70 CI job) so cargo enforces the MSRV and this class cannot regress silently?

PR #34, opened 2026-08-25
perf(wal): fsync at commit/DDL boundaries instead of per entry

semihalev:
## The problem

In `sync_mode=full`, every WAL entry forced its own buffer flush **and fsync**. An N-row transaction paid N+1 fsyncs where one establishes the same guarantee: the commit-marker flush carries all of the transaction's buffered DML bytes, and recovery discards uncommitted transactions regardless. DDL still syncs immediately.

## Measured (file DB, sync_mode=full)

- 1,000-row transaction: **3,742 ms -> 10 ms**
- 100 single-row autocommits: 744 ms -> 375 ms (each previously paid two fsyncs: DML entry + commit marker)

## Independent review shaped the final design

An adversarial review of the first version found a real HIGH flaw: with buffer drains happening outside the lock (`mem::take`, then write), a concurrent committer could carry another transaction's DML bytes, lose the race to the file mutex, and fsync its own marker first, so an acknowledged commit could lose rows on power failure. The final design serializes drains under the `wal_file` lock (the same `wal_file -> buffer` order `truncate_wal` already uses), which restores the invariant that a commit's sync covers every byte appended before it. Appenders block only during the buffered `write_all`, never during the fsync, so the measured numbers are unchanged.

Two more review findings folded in: a failed WAL write now leaves the bytes in the buffer for retry instead of dropping them, and `flush()` maintains `current_file_position`, which otherwise drifts and delays WAL rotation now that mid-transaction bytes can reach it.

## Semantics and docs

- `Normal` mode is bit-for-bit equivalent to before (verified by review).
- `full` documentation updated: "fsync at every commit and DDL" — the user-visible durability contract (acknowledged commits survive power loss) is unchanged.
- The unit test that codified per-entry sync now asserts the new contract.

## Verification

- `cargo fmt`, clippy 1.98 `-D warnings`, both full suites green (4868/4868), including the 156-test durability suite and WAL truncation/persistence targets.
- Includes the clippy 1.98 chore commit (#33); the diff shrinks to the WAL change once #33 merges.

### DIFF
--- a/docs/_docs/getting-started/connection-strings.md
+++ b/docs/_docs/getting-started/connection-strings.md
@@ -68,7 +68,7 @@ file:///path/to/data?sync_mode=normal&checkpoint_interval=60
 
 | Option | Values | Default | Description |
 |--------|--------|---------|-------------|
-| `sync` / `sync_mode` | none/off, normal, full (or 0, 1, 2) | normal | WAL synchronization mode. none/off=no fsync (data durable at checkpoint), normal=fsync every 1s, full=fsync every write |
+| `sync` / `sync_mode` | none/off, normal, full (or 0, 1, 2) | normal | WAL synchronization mode. none/off=no fsync (data durable at checkpoint), normal=fsync every 1s, full=fsync at every commit and DDL |
 | `checkpoint_interval` | Integer (seconds) | 60 | Time between automatic checkpoint cycles |
 | `compact_threshold` | Integer (min 2) | 4 | Sub-target volumes per table before merging |
 | `keep_snapshots` | Integer | 3 | Backup snapshots to retain per table |
@@ -111,7 +111,7 @@ file:///data/mydb?cleanup=off
 |------|-------|-------------|
 | none | 0 | No fsync. Data is durable only after checkpoint writes volumes to disk. |
 | normal | 1 | Fsync every 1 second (configurable via `sync_interval_ms`). DDL fsyncs immediately. Comparable to SQLite WAL + synchronous=NORMAL. |
-| full | 2 | Fsync on every write operation. Maximum durability. |
+| full | 2 | Fsync at every commit and DDL. Maximum durability: every acknowledged commit survives power loss. |
 
 ## Usage Examples
 
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -4804,8 +4804,11 @@ impl Executor {
         let mut has_float = false;
         let mut has_value = false;
 
-        // Unroll loop by 4 for better CPU pipelining
-        let (chunks, remainder) = rows.as_chunks::<4>();
+        // Unroll loop by 4 for better CPU pipelining.
+        // as_chunks needs Rust 1.88; the documented MSRV is 1.70.
+        #[allow(clippy::chunks_exact_to_as_chunks)]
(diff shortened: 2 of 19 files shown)


semihalev:
P1 addressed (failed commit marker re-written by a later flush): a WAL write failure now best-effort truncates the partial suffix off the fd (tracked pre-write position) and POISONS the WAL - every subsequent append/flush fails until the database is reopened, and reopen recovers cleanly with unacknowledged transactions discarded. A failed commit fsync poisons too (the written marker's durability is unknowable afterwards). Poison was chosen over stripping the failed transaction's bytes from the shared buffer: the buffer can hold multiple transactions and surgical removal on the error path is fragile; happy to switch if strip is preferred. Contract change: three failpoint tests that asserted writes-recover-after-disarm (which only worked via the unsafe blind retry) now assert the poison-until-reopen contract; the key assertion was verified to fail against the previous commit. Full gate green (clippy 1.98, 4868/4868 both suites, failpoint suite 11/11).

semihalev:
Both round-2 P1s addressed:

1. **Fsync-failure marker persisted by close()**: the WAL now tracks `synced_position` (last successful fsync offset, advanced under the wal_file lock, reset on rotation, updated by truncate_wal). Any poison event truncates the file back to this durable floor - stronger than guarding close() alone, since kernel writeback could persist unsynced marker bytes without any fsync. A poisoned close() skips both flush and sync. In Full mode everything above the floor is unacknowledged by construction. The floor also exposed a pre-existing hole: DDL entries were fsynced but their commit markers were not, so acked DDL was not actually durable within the Normal-mode sync interval; DDL transaction markers now sync unconditionally.

2. **Poison check outside the lock**: re-checked after acquiring wal_file (pre-lock check kept as fast path), so a flush waiting on the lock cannot write the stale buffer after a concurrent poison. truncate_wal, which drains the buffer under the same lock, now refuses to run poisoned.

New regression test (sync failpoint -> commit fails -> close -> reopen -> row must not exist) verified to FAIL against the previous commit - the old code did resurrect the row via close-time fsync. Failpoint suite 12/12, clippy 1.98 clean, both full suites 4868/4868.

semihalev:
Round-3 P1s addressed (all three):

1. **truncate_wal poison TOCTOU**: the check is repeated after acquiring the wal_file lock (pre-lock check kept as fast path), same pattern as flush_and_maybe_sync.
2. **truncate_wal buffer drain**: no more mem::take - bytes stay in the buffer, a write failure goes through poison_and_truncate (poison + truncate to the durable floor), and the drain's sync now uses sync_with_file so success advances the floor and failure poisons. A later commit marker can no longer ack an incomplete transaction.
3. **rotate_wal counter reset race**: current_file_position and synced_position resets moved inside the file-swap critical section, so an interleaving commit's fsynced floor can no longer be zeroed and later truncated away.

On the push note: the PR head was already 1b06f7fa when checked via gh (the local/remote mismatch was likely a mid-push read). This round is pushed on top.

Verification: clippy 1.98 clean, failpoint suite 12/12, WAL unit tests 28/28, durability+truncation+persistence 171/171, both full suites 4868/4868. The two lock-ordering races (1, 3) are not deterministically testable; they were verified by inspection against the locking design. Finding 2's invariant is covered by the existing resurrection regression test.

PR #35, opened 2026-08-26
fix: fail closed when cold volumes cannot be reloaded

semihalev:
The worst remaining verified bug from the engine audit: a persistent cold-volume reload failure (I/O error, fd exhaustion, missing file) was handled by silently excluding the segment. SELECT returned short results with Ok status, and UPDATE/DELETE skipped the segment's rows while reporting success. The same conflation existed across the whole access family: ensure_volume returned None for both 'removed by compaction' (legitimately skippable) and 'reload failed' (must not be), and the point-lookup helpers turned reload failures into 'row not found' during PK/unique checks.

## What changed (~60 sites)

- get_volumes_newest_first -> Result; persistently-cold segments are an error naming the segments, not a retain-drop.
- ensure_volume -> Result<Option>: Ok(None) strictly means removed-by-concurrent-compaction.
- The point-lookup family (get_cold_row, get_cold_row_normalized, retries, check_value_exists_*, find_row_id_by_values_*, get_authoritative_value) propagates errors; retries fail on a still-cold volume instead of touching missing column data.
- SegmentedTable: find_segment_row, find_segment_row_id_by_values, create_segment_scanners_filtered, collect_cold_rows and friends are fallible; Option fast paths bail to the general path, whose plumbing now fails closed; the parallel closures signal via reload/bail flags checked on every collection path.
- Trait level: get_active_row_ids and fetch_rows_by_ids_into became fallible; collect_hot_row_ids_into stays infallible via direct hot-side impls (18 call sites untouched).
- engine: compaction candidates skip cleanly (count check aborts the merge); get_row_fetcher closure returns Result; HNSW rebuild propagates.

## Independent review round (findings fixed in the second commit)

- segments_snapshot still carried the old retain-drop (making a new guard dead code); point lookups could return 'row not found' or a STALE overlapping version after a reload failure. Now fails closed.
- The lazy window-partition path treated a reload failure as an empty partition and silently dropped it from results; get_rows_for_partition_value now returns Result<Option> and the window path propagates.
- create_backup_snapshot's ? bypassed the writer.fail() temp-file cleanup; it now follows the standard fail-and-break arm.
- Hardened a pre-existing get_partition_count double-call unwrap the conversion could newly trigger.

**Deliberate tradeoff for review**: engine open now fails if an HNSW-indexed table has an unreadable cold volume at recovery (previously the ANN index silently missed those vectors). Fail-closed at open is consistent with the manifest policy from #32; say the word if degraded-open is preferred there.

## Verification

- New regression test drives a volume hot->warm->cold through the real eviction machinery with no reload source, and asserts ensure_volume / get_volumes_newest_first / get_cold_row all fail closed. (The pre-fix silent drop cannot be compile-checked against the new signatures; it was verified by inspection and the empirical probes recorded in review.)
- clippy 1.98 -D warnings clean; both full suites 4886/4886 before and after the review-repair commit.

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -6797,14 +6797,16 @@ impl Executor {
             }
         };
 
-        // Check if column has an index (required for fast path)
-        if table.get_partition_count(&column_name).is_none() {
-            *compiled_guard = CompiledExecution::NotOptimizable(self.engine.schema_epoch());
-            return None;
-        }
-
-        // Get the count
-        let count = table.get_partition_count(&column_name).unwrap();
+        // Check if column has an index (required for fast path). Reuse the
+        // value: a second call could return None (e.g. seal overlap or a
+        // cold-volume reload failure) and an unwrap would panic.
+        let count = match table.get_partition_count(&column_name) {
+            Some(c) => c,
+            None => {
+                *compiled_guard = CompiledExecution::NotOptimizable(self.engine.schema_epoch());
+                return None;
+            }
+        };
 
         // Cache the compiled state
         let compiled_cd = CompiledCountDistinct {
--- a/src/executor/context.rs
+++ b/src/executor/context.rs
@@ -325,7 +325,7 @@ pub fn cache_exists_index(key: String, index: std::sync::Arc<dyn Index>) {
 }
 
 /// Type alias for row fetcher function used in EXISTS/COUNT optimization.
-pub type RowFetcher = Box<dyn Fn(&[i64]) -> crate::core::RowVec + Send + Sync>;
+pub type RowFetcher = Box<dyn Fn(&[i64]) -> crate::core::Result<crate::core::RowVec> + Send + Sync>;
 
 /// Type alias for row counter function used in COUNT(*) optimization.
 /// This only counts visible rows without cloning their data.
--- a/src/executor/dml.rs
+++ b/src/executor/dml.rs
@@ -202,7 +202,7 @@ impl Executor {
                             .collect();
 
                         let mut row_ids: Vec<i64> = table
-                            .get_active_row_ids()
+                            .get_active_row_ids()?
                             .into_iter()
                             .filter(|id| !excluded.contains(*id))
                             .collect();
--- a/src/executor/index_optimizer.rs
+++ b/src/executor/index_optimizer.rs
@@ -1109,7 +1109,7 @@ impl Executor {
             Box::new(ConstBoolExpr::true_expr());
 
         // Fetch rows by row_ids - returns RowVec directly
-        let mut rows = table.fetch_rows_by_ids(&all_row_ids, filter.as_ref());
+        let mut rows = table.fetch_rows_by_ids(&all_row_ids, filter.as_ref())?;
 
         // Apply remaining predicate if any
         if let Some(ref remaining) = remaining_predicate {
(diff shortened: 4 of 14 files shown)


semihalev:
Both P1s addressed:

1. **Test race on GLOBAL_EVICTION_EPOCH**: both eviction-driving tests now serialize on a shared mutex AND use base-relative epochs (loaded from the global at test start), so neither ordering can corrupt the other's idle-cycle accounting. Reproduced the failure with a threaded whole-module cargo test run before the fix; five consecutive runs are 12/12 after.

2. **Partial DML in an open transaction**: update / delete / update_by_row_ids / delete_by_row_ids now call preflight_cold_access() BEFORE the first hot-buffer mutation - ensure_columns plus fail-closed on any still-cold segment - so a (persistent) reload failure surfaces before the statement touches anything. Honest residual: a volume evicted between preflight and the cold iteration can still error mid-statement, but that requires eviction AND a fresh I/O failure inside one statement window; noted in the code and commit. The complete fix is statement-level savepoints, which I'd propose as separate work. The regression test asserts the preflight fails on an unreloadable table.

Full gate re-run: clippy 1.98 clean, both suites 4886/4886.

semihalev:
Atomicity P1 fully closed with the snapshot-carrying approach you suggested: all four DML entry points capture the verified all-warm segments_snapshot BEFORE the first hot-buffer mutation and use that same snapshot for the entire statement (where-clause paths via volumes_newest_first_from_snapshot, by-row-ids paths via a snapshot-scoped find_segment_row_in). Eviction copy-on-writes new maps and new volume Arcs, so the snapshot's volumes keep their column data for the statement's duration - a mid-statement reload can never be needed, so the reload-failure window is zero, not merely narrow. Segments compacted away mid-statement are found in the snapshot's older volumes, which is exactly the statement's consistent view. The interim preflight helper is removed; the regression test asserts the statement snapshot itself fails closed.

Cost note (unchanged since the preflight round, now stated on the PR): DML on a table with cold segments warms the whole cold set up front; a zone-map-targeted warm-up is possible future work.

Gate: clippy 1.98 clean, manifest module threaded 5x12/12, both full suites 4886/4886.

semihalev:
Snapshot self-containment P1 addressed: statement_snapshot() now captures the newest-first segment ids, the segment map, and the committed tombstone view together under ONE manifest lock hold (fail-closed on cold, one retry for the eviction race), and the statement reads nothing live afterwards - where-clause loops iterate StatementSnapshot::volumes_newest_first, mappings come from the snapshot's cs.mapping (no live get_volume_mapping in the DML loops), and the by-row-ids paths use find_segment_row_in over the snapshot's ids/map/tombstones. The only live read left is the pending-tombstone check, deliberately: that is this transaction's own in-flight state and must reflect the statement's own earlier deletes.

Regression test replays your deterministic scenario: capture snapshot, replace_segments_atomic compacts segment 1 into 2, the snapshot must still serve segment 1 (the live-ordering version returned zero volumes there).

Gate: clippy 1.98 clean, manifest module threaded 3x13/13, both full suites 4887/4887. (Head note: 0312a077 was confirmed as the GitHub PR head via gh at review time - the fetch likely raced the push; current head includes this fix.)

semihalev:
Unique-check P1 addressed: check_cold_unique_for_update now receives the StatementSnapshot from both UPDATE call sites and runs its lookup entirely against the snapshot's parts (find_row_id_by_values_in over seg_ids_newest_first / segs / tombstones). get_authoritative_value was the last internal spot re-capturing live state; it now takes the caller's (seg_ids, segs) view, so the authoritative-value verification stays inside the same snapshot too. INSERT-path constraint checks deliberately keep the live snapshot - they run before any mutation.

Regression test replays your probe exactly: with the live map cold and unreloadable, the live lookup errors while the statement-snapshot lookup still returns Some(5). Gate: clippy 1.98 clean, manifest module threaded 3x14/14, both full suites 4888/4888.

PR #36, opened 2026-08-26
fix: collision-proof the compiled-program caches

semihalev:
Both compiled-program caches - the global LRU in compile_expression_cached and CompiledEvaluator's per-instance local_cache - were keyed by a single unverified 64-bit FxHash of the expression AST. On a key collision the cache silently returns and executes the WRONG compiled program: no error, just wrong results for every execution served from the aliased entry. FxHash's weak diffusion makes structured collisions realistic on real SQL shapes, not merely birthday-bound.

The key is now a pair of hashes from independent algorithms (FxHash plus fixed-seed ahash), so aliasing requires a simultaneous collision in both - effectively impossible. Cost: one extra hash walk on the compile-miss path only; hit-path lookups are a tuple compare instead of a u64 compare.

The second cache (local_cache) was found during implementation - same pattern, same fix. A unit test pins key stability (fixed seeds) and that distinct expressions/column sets differ in BOTH components.

Verification: clippy 1.98 -D warnings clean, module tests 21/21, both full suites 4886/4886.

### DIFF
--- a/src/executor/expression/evaluator_bridge.rs
+++ b/src/executor/expression/evaluator_bridge.rs
@@ -180,6 +180,16 @@ fn hash_expression<H: std::hash::Hasher>(expr: &Expression, hasher: &mut H) {
             in_hash.not.hash(hasher);
             hash_expression(&in_hash.column, hasher);
             in_hash.values.len().hash(hasher);
+            // Hash the actual set contents: length alone made `a IN {1}`
+            // and `a IN {2}` share a key, and no key derivation can help
+            // when its input omits the values. Canonical (sorted) order
+            // keeps the hash set-order-independent while every algorithm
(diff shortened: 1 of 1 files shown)


semihalev:
P1 addressed - and the finding was better than my fix: the dual-algorithm key cannot help when its INPUT omits the values. The InHashSet arm hashed only values.len(); it now hashes the actual set contents in canonical (sorted) order so each algorithm sees the full material directly (deliberately NOT combined via a single inner hash, which would reintroduce the same class). The audit found the same len-only deficiency in ValuesSource.rows, also fixed. Your repro is now the regression test (two RowFilters; the second must not be served the first's program; a=2 must match a IN {2}) and was verified to FAIL against the len-only hashing. Full gate: clippy clean, both suites 4887/4887.

PR #39, opened 2026-08-30
perf(executor): cut per-statement fixed costs on the cached SELECT path

semihalev:
Wave A of the executor performance campaign (from the 72-finding agent audit). Targets the fixed cost every cached/prepared statement pays before real work starts.

## Changes

1. **One transaction-state capture per statement.** `execute_cached` / `execute_with_cached_plan` capture the active-transaction id once and pass it down. Before, a cached SELECT could take the `active_transaction` mutex up to 6 times (PK probe, COUNT(DISTINCT) probe, COUNT(*) probe, execute_statement injection, plus execution-time acquisitions). The compiled probes are now lock-free and gated by the caller (same contract as the existing `_with_params` variants); the probe battery is deduplicated into one `try_compiled_fast_paths` helper used by all three entry points. `execute_statement` no longer unconditionally clones the `ExecutionContext` in autocommit, and the redundant uncompiled PK-detection pass is skipped when the compiled battery already ran (any None from the compiled probe implies None from the uncompiled one).
2. **Negative statistics caching** (`planner.rs`). A fresh cache entry with `row_count == 0` is now the authoritative "no stats collected" verdict. Before, every SELECT-with-WHERE on a non-ANALYZEd table paid `begin_transaction` + `list_tables` + two system-table existence scans, per statement, forever. The only stats writer (`execute_analyze`) already invalidates the entry per table (verified: statistics.rs:94); the 300s TTL bounds staleness for external changes. The LRU `touch()` is now an atomic stamp under the read lock instead of a write-lock round trip per cache hit.
3. **Classification attached to the cached plan** (`OnceLock<Arc<QueryClassification>>` in `CachedQueryPlan`/`CachedPlanRef`). Classification is a pure function of the immutable AST, so repeated executions skip the full-AST structural hash and the process-global classification-cache mutex. First execution still populates through the global cache (cross-plan dedup preserved).
4. **Column names as the schema's cached Arc.** `execute_simple_table_scan` used `column_names_owned().to_vec()`, a deep clone of every column name per SELECT; now the cached `CompactArc` flows through, and `ScannerResult` stores the Arc. One site (`execute_temporal_query`, AS OF path) intentionally keeps its clone: not a hot path.
5. **Literal LIMIT/OFFSET fast path.** `LIMIT 10` no longer goes through expression compile + global program cache + VM execution per statement.

Deferred to Wave D by design: the `subquery.to_string()` cache keys (audit items at index_optimizer.rs:907 / subquery.rs:1704). That is a cache-identity change; per the PR #36 lesson it needs dual-hash keys plus a collision discrimination test, and it belongs with the subquery-cache cluster.

## Measurements

Release build, `memory://`, 10K-row table, best of 3 runs (micro-bench in session scratchpad, not committed):

| Scenario | main | this PR | speedup |
|---|---|---|---|
| prepared point PK SELECT | 1464 ns | 915 ns | **1.60x** |
| same-SQL point SELECT (plan cache) | 1547 ns | 999 ns | 1.55x |
| WHERE + LIMIT 10 (literal) | 2122 ns | 1423 ns | 1.49x |
| COUNT range, no ANALYZE | 109.2 us | 77.0 us | 1.42x |
| COUNT range, ANALYZE'd | 102.4 us | 80.4 us | 1.27x |

## Safety notes for review

- The probe battery runs only when no explicit transaction is active; INSERT's compiled path is exempt and handles the transaction itself (unchanged behavior).
- Skipping the uncompiled PK probe after the compiled battery is safe because both share the same detection outcome: NotOptimizable epochs, param-type mismatches, and epoch-changed recompiles all end in the same fallback.
- Negative stats entries can only go stale if something writes sys stats without invalidating; the only writer is ANALYZE, which invalidates.

## Verification

fmt + clippy `-D warnings` clean; both suites green, 4890/4890 each (default and `test-filedb`).

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -6572,16 +6572,8 @@ impl Executor {
         stmt: &SelectStatement,
         compiled: &RwLock<CompiledExecution>,
     ) -> Option<Result<Box<dyn QueryResult>>> {
-        // Quick reject: must not be in an explicit transaction (for simplicity)
-        {
-            let active_tx = match self.active_transaction.try_lock() {
-                Ok(guard) => guard,
-                Err(_) => return None,
-            };
-            if active_tx.is_some() {
-                return None;
-            }
-        }
+        // Caller (try_compiled_fast_paths) guarantees no explicit transaction
+        // is active; this path reads committed state only.
 
         // Try read lock first - check if already compiled
         {
@@ -6850,16 +6842,8 @@ impl Executor {
         stmt: &SelectStatement,
         compiled: &RwLock<CompiledExecution>,
     ) -> Option<Result<Box<dyn QueryResult>>> {
-        // Quick reject: must not be in an explicit transaction (for simplicity)
-        {
-            let active_tx = match self.active_transaction.try_lock() {
-                Ok(guard) => guard,
-                Err(_) => return None,
-            };
-            if active_tx.is_some() {
-                return None;
-            }
-        }
+        // Caller (try_compiled_fast_paths) guarantees no explicit transaction
+        // is active; this path reads committed state only.
 
         // Try read lock first - check if already compiled
         {
--- a/src/executor/dml_fast_path.rs
+++ b/src/executor/dml_fast_path.rs
@@ -184,7 +184,9 @@ impl Executor {
         let (pk_value, pk_value_source) = match val_expr {
             Expression::IntegerLiteral(lit) => (lit.value, PkValueSource::Literal(lit.value)),
             Expression::FloatLiteral(lit) => {
-                let v = lit.value as i64;
+                // Only lossless float keys qualify: truncating 5.5 to 5
+                // would mutate row 5 where correct execution matches nothing.
+                let v = Self::lossless_float_key(lit.value)?;
                 (v, PkValueSource::Literal(v))
             }
             Expression::Parameter(param) => {
@@ -195,7 +197,7 @@ impl Executor {
                     let value = ctx.get_named_param(name)?;
                     let pk_value = match value {
                         Value::Integer(i) => *i,
-                        Value::Float(f) => *f as i64,
+                        Value::Float(f) => Self::lossless_float_key(*f)?,
                         _ => return None,
                     };
                     (
@@ -214,7 +216,7 @@ impl Executor {
                     }
                     let pk_value = match &params[param_idx] {
                         Value::Integer(i) => *i,
-                        Value::Float(f) => *f as i64,
+                        Value::Float(f) => Self::lossless_float_key(*f)?,
                         _ => return None,
                     };
                     (pk_value, PkValueSource::Parameter(param_idx))
@@ -235,7 +237,7 @@ impl Executor {
         match source {
             PkValueSource::NamedParameter(name) => match ctx.get_named_param(name)? {
                 Value::Integer(i) => Some(*i),
-                Value::Float(f) => Some(*f as i64),
+                Value::Float(f) => Self::lossless_float_key(*f),
                 _ => None,
             },
             _ => Self::extract_pk_value_from_params(source, ctx.params()),
@@ -253,7 +255,7 @@ impl Executor {
                 }
                 match &params[*idx] {
                     Value::Integer(i) => Some(*i),
-                    Value::Float(f) => Some(*f as i64),
+                    Value::Float(f) => Self::lossless_float_key(*f),
                     _ => None,
                 }
             }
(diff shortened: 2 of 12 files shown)


semihalev:
Review round pushed (7a92ad8d). Sources: Semih's reviewer (P2), two independent adversarial agent passes, and the discrimination tests written for the fixes.

**Fixed in this round**
1. Cached INSERT double mutex (reviewer P2): INSERT dispatches before the txn-state capture at all three entry points; single acquisition restored.
2. NotOptimizable value-poisoning (adversarial pass 1): prepared `WHERE id = $1` first executed with a text/NULL parameter permanently killed the fast path once the uncompiled probe was skipped. Detection split into structural vs value; the compiled PkLookup now survives wrongly-typed executions. Sequence test added.
3. Cross-handle ANALYZE blindness (both passes): `stats_epoch` on MVCCEngine, bumped by ANALYZE, staling every planner's cached verdict including the negative one. Mechanism unit test added.
4. Classification-key collision (adversarial pass 2): the key ignored the FROM clause; `SELECT * FROM a` vs `... JOIN b` could trade classifications, and the new plan-pinned OnceLock would have frozen a collision. Key now hashes join shape/types/table names. Unit test proven red pre-fix.

**Wrong-results bugs found on main during this review (fixed here, red-first tests):**
- PK fast path ignored `LIMIT 0`/`OFFSET`: `SELECT * FROM t WHERE id = 5 LIMIT 0` returned the row. Literal `LIMIT >= 1` keeps the fast path.
- Fast paths truncated fractional float keys: `UPDATE t SET ... WHERE id = 5.5` mutated row 5. Fast paths now require lossless conversion.

**Deferred, will be its own PR:** the float truncation also lives in the standard path (index-probe/columnar predicate extraction; `SELECT COUNT(*) FROM t WHERE n < 5.5` with n=5 returns 0). Storage `comparison.rs` and the VM arms are correct; the extraction layer is not. Failing tests are staged.

**Also noted from bisecting the bench-compare regressions on one machine:** `UPDATE by ID` +23% is attributable to #36's dual-hash on the DML compiled path (Wave B target); the `BETWEEN +30%` from BENCHMARKS.md does not reproduce as a commit-attributable change (all three commits sit in the same 3.5-4.1us band), so the reference column likely predates the current machine state.

Gates: fmt + clippy -D warnings clean; both suites 4897/4897 (7 new tests). Bench after this round: prepared point PK 862ns (1.70x vs main), WHERE+LIMIT literal 1238ns (1.71x), COUNT range ~68us.

semihalev:
Third round pushed. All three findings addressed:

1. **[P1] 2^63 boundary**: `lossless_float_key` now range-checks `[-2^63, 2^63)` before the cast (rejects NaN/infinities in the same check); the literal arm's inline copy routes through the same helper. Unit tests cover 2^63, 1e19, exact i64::MIN, NaN, both infinities. The end-to-end boundary test joins the staged float-truncation tests: with the fast path correctly rejecting, the standard path's index probe still saturates, which is the same deep-layer family as the fractional cases and ships in the dedicated float PR.
2. **[P1] classification key**: the hash walker now covers every child the classifier reads (IN-list/ExpressionList elements, LIKE left/pattern/escape/operator, CASE value, DISTINCT expr, scalar-subquery columns, ALL/ANY left, InHashSet), and the key is a dual independent 64-bit pair computed by one shared input function, per the program-cache precedent. Your exact scenario is the regression test: `IN (1, 2)` vs `IN (1, RANDOM())` was red on 7a92ad8d and is green now, including the end-to-end nondeterminism flags.
3. **[P2] epoch TOCTOU**: the epoch is captured before the system-table read and stamped into the entry, so an ANALYZE committing mid-read leaves the entry stale on the next access instead of fresh-stamped for the TTL.

Gates: fmt + clippy -D warnings clean; both suites 4899/4899.

semihalev:
Fourth round pushed. The remaining P1 is closed:

- The four subquery arms (`EXISTS`, `ScalarSubquery`, `SubquerySource`, `AllAny`) now hash columns + FROM structure + WHERE through one `hash_subquery_structure` helper, and the FROM walker mirrors `collect_tables_recursive` exactly: table names with aliases, bare identifiers, derived-source aliases (plus body), and function table sources. The alias that decides local-vs-outer resolution is now always part of the key.
- Regression test is your exact pair: `EXISTS (SELECT 1 FROM inner_t o WHERE o.id = 1)` vs `... FROM inner_t i WHERE o.id = 1`. Red on b9ae14ef, green now, and the end-to-end `where_has_correlated_subqueries` flags are asserted (false/true respectively).

Gates: fmt + clippy -D warnings clean; both suites 4900/4900.

PR #40, opened 2026-08-30
fix: float constants against INTEGER columns compare numerically

semihalev:
Found while hardening the PK fast path in #39: the fast paths truncated float keys, and it turned out the standard path does too, at several independent layers. This PR fixes the family engine-wide.

## Symptoms (all red-first tests on main)

- `WHERE n = 5.4` matched n = 5; `WHERE n != 5.5` matched nothing
- `WHERE n < 5.5` returned zero rows for n = 5
- `WHERE n IN (5.5, 6.5)` matched n = 5
- `WHERE n BETWEEN 5.1 AND 5.9` matched n = 5
- `SELECT COUNT(*) WHERE n > 499.5` counted through the truncated bound
- `UPDATE t SET ... WHERE id = 5.5` mutated row 5; DELETE likewise
- `WHERE id = 9223372036854775808.0` (2^63) matched the i64::MAX row

## Root causes and fixes, by layer

1. **Plan-time coercion (the main root)**: `pushdown` coerced WHERE constants with `into_coerce_to_type`, the INSERT-style storage coercion, turning `n = 5.5` into `n = 5` before anything else ran. WHERE constants now convert Float to Integer only losslessly and otherwise stay floats.
2. **Compiled storage filters**: the typed variants were chosen by constant type only, so a Float filter against an INTEGER column never matched anything (`n < 5.5` returned zero rows once layer 1 stopped truncating). Selection is now schema-aware, and fractional ranges rewrite to exact integer bounds (`n < 5.5` is `n <= 5`, `n > -5.5` is `n >= -5`, BETWEEN gets `[ceil, floor]`), which keeps these index- and zone-map-friendly. Fractional equality falls to the dynamic path, which compares numerically.
3. **Between/Range/InList storage expressions**: integer cells now compare float bounds and float list values numerically instead of through `as_int64`.
4. **Exact extreme comparison**: `i as f64` rounds above 2^53, so `Integer(i64::MAX) == Float(2^63)` was true across PartialEq, Value::compare, the storage comparison, and the fused VM ops. New `core::value::cmp_i64_f64` compares exactly over the full range (unit-tested at the boundaries); all cross-type numeric comparison sites now use it, including 49 fused VM op arms in both directions. Arithmetic casts are untouched.

## Verification

- 6 end-to-end regression tests, each proven red on main before the fix, plus unit tests for `cmp_i64_f64` boundaries
- fmt + clippy -D warnings clean
- Both suites green: 4895/4895 default, 4895/4895 test-filedb; the new integration test also passes under CI-mode `cargo test`
- Perf numbers on the hot comparison paths follow as a comment. The integer-vs-integer fast arms are untouched; the exact comparison only runs on cross-type (float-vs-integer) arms.

### DIFF
--- a/src/core/value.rs
+++ b/src/core/value.rs
@@ -390,7 +390,21 @@ impl Value {
             return self.compare_same_type(other);
         }
 
-        // Cross-type numeric comparison (integer vs float)
+        // Cross-type numeric comparison (integer vs float), exact at the
+        // extremes where `i as f64` rounds
+        match (self, other) {
+            (Value::Integer(i), Value::Float(f)) => {
+                if let Some(ord) = cmp_i64_f64(*i, *f) {
+                    return Ok(ord);
+                }
+            }
+            (Value::Float(f), Value::Integer(i)) => {
+                if let Some(ord) = cmp_i64_f64(*i, *f) {
+                    return Ok(ord.reverse());
+                }
+            }
+            _ => {}
+        }
         if self.data_type().is_numeric() && other.data_type().is_numeric() {
             let v1 = self.as_float64().unwrap();
             let v2 = other.as_float64().unwrap();
@@ -892,7 +906,7 @@ impl PartialEq for Value {
             // Cross-type numeric comparison: Integer vs Float
             // This is critical for queries like WHERE id = 5.0 or WHERE price = 100
             (Value::Integer(i), Value::Float(f)) | (Value::Float(f), Value::Integer(i)) => {
-                *f == (*i as f64)
+                cmp_i64_f64(*i, *f) == Some(std::cmp::Ordering::Equal)
             }
             (Value::Text(a), Value::Text(b)) => a == b,
             (Value::Boolean(a), Value::Boolean(b)) => a == b,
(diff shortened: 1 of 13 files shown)


semihalev:
Perf sanity on the hot comparison paths (release, same machine as the earlier three-commit bisect, best of 3):

| Bench | main band (bisect) | this PR |
|---|---|---|
| BETWEEN (non-indexed) | 3.5-4.1 us | 2.90 us |
| IN list (7 values) | ~2.4-2.5 us | 2.23 us |
| SELECT by ID | ~0.11-0.13 us | 0.130 us |
| Math expressions | ~15.5-16 us | 14.4 us |
| SELECT by index (range) | ~26-30 us | ~30 us |

No regression signal; integer-vs-integer fast arms are untouched and the exact comparison only runs on cross-type arms. (The benchmark's own BETWEEN uses integer literals, so its improvement is machine-state variance, not this PR.)

semihalev:
Review round pushed (d3ffdba6). Your four findings plus the independent adversarial pass landed together:

1. **[P1] Ord**: now `cmp_i64_f64`; Eq/Ord/PartialOrd invariant tests added at the extremes (MAX vs 2^63 both directions, exact MIN, NaN ordered last both ways).
2. **[P1] IN boundary**: the hashed set only accepts floats with an exact i64 equivalent; linear fallback and the float-cell path use `i64_eq_f64`. Regression tests cover `IN (9223372036854775808.0, ...)`, `NOT IN`, and the 2^53+1 rounding case.
3. **[P1] BETWEEN**: fixed at the depth you pointed at — each bound now compares IN ITS OWN TYPE (`int_cell_lower_ok`/`float_cell_upper_ok` family). My earlier joint `as_float64` conversion rounded an Integer(i64::MAX) bound to 2^63; your MAX/MAX-1 scenario is a regression test now, plus the float-cell-vs-integer-upper-bound mirror. The VM's mixed-bound Between/NotBetween arms and the `*_int` fused helpers are converted too.
4. **[P2] NaN `<>`**: `None => operator == Ne` in both `evaluate` and `evaluate_fast`, consistent with the VM wrappers.

Also in this push from the adversarial pass: exact join comparators (hash-table bit-cast equality and utils' epsilon equality removed; merge-join ordering now matches Value::compare), reverse-direction coercion (large INTEGER literal vs FLOAT column no longer rounds at plan time; compiled filters schema-aware in that direction), volume-scanner typed pre-filter lossless guard, RangeExpr float-bound flag, and exact semantic-cache numeric identity (the 1e-10 tolerance and as-f64 integer literals could let >2^53 predicates share a cache identity; e2e test included).

One discovery worth its own issue: cross-type hash equi-joins (`a.i = b.f`) never match on main even for representable pairs (42 vs 42.0) — the join hash table buckets Integer and Float keys separately. Documented in the join test; separate fix planned.

Gates: fmt + clippy -D warnings clean; both suites 4901/4901; the float test file passes under CI-mode `cargo test` as well.

semihalev:
Both findings closed in the latest push:

1. **[P1] i64::MIN panic**: reproduced red-first (`attempt to negate with overflow` with a prepared `i64::MIN` parameter against a FLOAT column). Fixed with your suggested shape: `core::value::lossless_f64_from_i64` checks `cmp_i64_f64(i, i as f64) == Equal`, applied at all three sites (pushdown coercion, compiled filters, volume scanner). Side benefit: the old `abs() < 2^53` check wrongly REJECTED i64::MIN, which is exactly representable; the new check accepts it, and the regression test asserts the MIN parameter finds its row.
2. **[P2] RangeExpr**: per-bound `min_is_float`/`max_is_float` flags replace the joint routing; `check_integer` and `check_float` compare each bound in the bound's own type, mirroring BetweenExpr. Your scenario is a unit test: `[Integer(i64::MAX), Float(2^63)]` rejects MAX-1 and accepts MAX; a 2^63 float cell is rejected by an `Integer(i64::MAX)` upper bound; fractional bounds stay empty. Both `evaluate` and `evaluate_fast`.

Gates: fmt + clippy -D warnings clean; both suites 4903/4903; the float test file passes under CI-mode `cargo test`.

PR #41, opened 2026-08-30
perf(dml): cut fixed costs on the prepared INSERT path

semihalev:
Wave B1 of the executor perf campaign (audit big-wins 2 and 3): the hottest OLTP write path stops paying parse/hash/lock costs per row.

## Changes

1. **Direct parameter resolution.** Every `$n` in a prepared INSERT went through `compile_expression_cached` on every execution (per row for multi-row VALUES): a full AST hash, the global PROGRAM_CACHE mutex, and a VM run, to fetch a value already sitting in `ctx.params()`. `try_extract_simple_value` now resolves positional and named parameters directly; complex expressions still take the VM.
2. **CHECK constraints pre-compile once per plan** into `CompiledInsert` (`SharedProgram`, the same pattern `execute_update` and the upsert path already use) instead of `format!("SELECT {check}")` + tokenizer + parser + compile per inserted row per constrained column. Truthiness rules, NULL-passes, and skip-on-parse-failure semantics preserved exactly; both compiled-path validation sites (VALUES loop and INSERT..SELECT) converted. `CompiledInsert` gets a manual `Debug` since programs aren't `Debug`.
3. **DEFAULT expressions evaluate once per statement** in the plain `execute_insert` paths (three sites), matching the compiled path's per-plan template. Previously each row re-parsed the default SQL.

## Measurements

Release, `memory://`, best of 3 (micro-bench in session scratchpad):

| Scenario | main | this PR | speedup |
|---|---|---|---|
| prepared INSERT, 2 CHECK constraints | 1941 ns | 629 ns | **3.09x** |
| prepared INSERT, plain | 721 ns | 621 ns | 1.16x |
| prepared INSERT, 4-row VALUES | 1642 ns | 1426 ns | 1.15x |

A table with CHECK constraints now inserts at nearly the same rate as one without.

## Verification

- New tests: parameter round-trips including NULL and text, CHECK violations still raised through the compiled plan (error names the column; later executions still enforce), NULL passes CHECK, DEFAULT templates for multi-row and `DEFAULT` keyword inserts
- fmt + clippy `-D warnings` clean; both suites green 4906/4906; the new test file passes under CI-mode `cargo test`

### DIFF
--- a/src/executor/dml.rs
+++ b/src/executor/dml.rs
@@ -158,6 +158,73 @@ fn try_extract_simple_value(expr: &Expression, ctx: &ExecutionContext) -> Option
     None
 }
 
+/// Evaluate one default slot for a row: run the per-row program when the
+/// default is non-literal, else clone the constant template value.
+fn eval_default_slot(
+    vm: &mut super::expression::ExprVM,
+    exec_ctx: &super::expression::ExecuteContext,
+    program: &Option<(super::expression::SharedProgram, DataType)>,
+    template_val: &Value,
+) -> Value {
+    match program {
+        Some((program, target)) => match vm.execute_cow(program, exec_ctx) {
+            Ok(v) => v.into_coerce_to_type(*target),
+            Err(_) => Value::null_unknown(),
+        },
+        None => template_val.clone(),
+    }
+}
+
+/// Per-statement default machinery: constant defaults land in the value
+/// template; non-literal defaults compile ONCE and execute PER ROW, so
+/// DEFAULT (RANDOM()) differs per row. Parse and compile failures degrade
+/// the slot to NULL, preserving evaluate_default_expr's swallowed-error
+/// behavior for defaults.
+#[allow(clippy::type_complexity)]
+fn build_default_slots(
+    schema: &crate::core::Schema,
+) -> (
+    Vec<Value>,
+    Vec<Option<(super::expression::SharedProgram, DataType)>>,
+) {
+    use super::expression::compile_expression;
+    let n = schema.columns.len();
+    let mut values = Vec::with_capacity(n);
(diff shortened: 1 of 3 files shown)


semihalev:
Both blockers closed (plus the adversarial pass findings, which matched yours):

1. **[P1] Volatile DEFAULTs**: exactly your prescription — parse/compile hoists once per statement, the PROGRAM runs per row. `build_default_slots` gives literal defaults a constant template slot and non-literal ones a compiled program executed per row; all three plain paths, the compiled path, and the `DEFAULT` keyword share the machinery. Bonus: this also fixes the compiled path's pre-existing per-PLAN freezing, so a prepared `DEFAULT (RANDOM())` now differs across executions too. Red-first test asserts `COUNT(DISTINCT r)` equals the row count across plain multi-row, prepared, and DEFAULT-keyword inserts.
2. **[P1] CHECK compile failure**: parse-failure skip preserved; compile failure now fails the statement with `CHECK constraint on column 'x' failed to compile: ...` — in the compiled-INSERT build and in the upsert build, which had the same pre-existing `.ok()?` swallow. The adversarial pass found the realistic trigger: `RENAME COLUMN` does not rewrite `check_expr`, so the swallow would have silently disabled the constraint. Red-first test: an uncompilable CHECK rejects prepared inserts.

Also from the adversarial pass: `CompiledInsert.all_column_types` was dead after the wave and is removed; noted as follow-ups (not this PR): the compiler's `LoadParam` u16 truncation for `$65538+` in giant multi-row inserts, and `RENAME COLUMN` should rewrite or reject stale `check_expr` text.

Gates: fmt + clippy clean; both suites 4908/4908; CI-mode `cargo test` passes. Perf retained: prepared INSERT 584-649ns, with CHECK 615-649ns (3.1x vs main), 4-row VALUES ~1.46us.

PR #42, opened 2026-08-30
perf(dml): expression SET values on the PK fast-update path

semihalev:
Wave B2 of the executor perf campaign: `UPDATE t SET n = n + 1 WHERE id = $1` — the most common UPDATE shape — fell off the PK fast path because SET only supported literal/parameter sources, and the full path recompiles the SET expression through the global program cache on every execution (AST double-hash + mutex). This is also where the +23% UPDATE-by-ID regression from the cache-key collision-proofing (#36) lived.

## Changes

- `UpdateValueSource::Expression(SharedProgram)`: row expressions compile once per plan against the schema columns; the minimal-update setter executes them per update against the current row.
- Semantics: all SET expressions read the ORIGINAL row (`SET a = b, b = a` swaps — probed identical to the full path), parameters inside expressions (`SET a = a + $1`) resolve through the execution context, NULL propagates.
- Safety: subquery-bearing expressions are rejected by a conservative walker and take the full path — the expression compiler emits placeholder subquery ops that only the full planner can resolve; the existing `bug_03` regression test caught exactly this during development. All existing fast-path guards are untouched (CHECK-constrained columns, FK tables, RETURNING, PK-column updates still fall back), and the `_with_params` probe declines Expression sources.

## Measurements

| Bench | main | this PR |
|---|---|---|
| UPDATE by ID (`SET balance = balance + 1`) | 0.70-0.76 us | **0.59-0.65 us** |
| prepared `SET a = a + 1 WHERE id = $1` (micro) | full path | 707-730 ns (fast path) |

Back at the pre-#36 level; the shape no longer touches the global program cache per execution.

## Scope note

The remaining audit items for this wave (post-SELECT WHERE re-evaluation in UPDATE — needs an MVCC-visibility argument, per-row update Vec hoist, upsert full-row clone, DELETE batching for RETURNING/FK) follow as a separate PR to keep this one reviewable.

## Verification

- Both suites 4914/4914 (default + test-filedb); CI-mode `cargo test` passes on the new file
- New tests: incremental reads through the cached plan, multi-assignment swap, parameters inside SET expressions, NULL propagation, CHECK enforcement via the full path, scalar-subquery SET correctness
- fmt + clippy -D warnings clean

### DIFF
--- a/src/executor/dml.rs
+++ b/src/executor/dml.rs
@@ -86,7 +86,7 @@ fn conflict_matches_target(conflict_target: &[Identifier], schema: &Schema, erro
 
 /// Validate type coercion didn't silently fail.
 /// Returns an error if a non-null value became null during coercion.
-fn validate_coercion(
+pub(crate) fn validate_coercion(
     original: &Value,
     coerced: &Value,
     column_name: &str,
@@ -2218,6 +2218,13 @@ impl Executor {
                         .collect()
                 };
 
+            // Column names for coercion errors, captured before the setter
+            // borrows the table mutably
+            let update_col_names: Vec<String> = compiled_updates
+                .iter()
+                .map(|(idx, _, _, _)| schema.columns[*idx].name.to_string())
+                .collect();
+
             // Pre-compile CHECK constraints for columns being updated (once, not per row)
             let compiled_check_exprs: Vec<(usize, String, String, SharedProgram)> =
                 compiled_updates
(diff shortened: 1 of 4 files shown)


semihalev:
Review round pushed. Your blocker plus the adversarial pass findings:

1. **[P1] VECTOR(n) dimension validation** (your blocker): `column_vector_dims` now travels in `CompiledUpdateColumn` and `validate_coercion` runs at all three application sites — expression setter, the pre-existing constant sources (`SET v = $1` had the same hole), and the `_with_params` probe (declines on validation failure so the ctx path raises the full error). Red-first test: `CAST('[1,2]' AS VECTOR)` into `VECTOR(3)` errors on both the expression and parameter routes, row unchanged.
2. **[P2] VM-error divergence** (adversarial pass): the FULL path's SET application swallowed VM errors with `if let Ok`, so `SET a = a + 1, b = 5` at `a = i64::MAX` half-applied the row (b set, a skipped) inside a transaction while the fast path errored in autocommit. The full path now propagates, matching the fast path: statement errors, nothing applied.
3. **[P2] Silent non-null-to-NULL** (adversarial pass): the full path also stored `SET a = 'abc'` (INTEGER) as NULL; it now runs the same `validate_coercion` as INSERT and the fast path. Cross-path consistency test covers both scenarios in autocommit and explicit transactions.
4. **[P3]** subquery walker now inspects `LIKE ... ESCAPE`, `FILTER`, and function-argument `ORDER BY`; `IntervalLiteral` is a safe leaf so `SET ts = ts + INTERVAL '1 day'` stays fast-pathed.

Gates: fmt + clippy clean; both suites 4916/4916; CI-mode `cargo test` passes.

PR #43, opened 2026-08-30
perf(dml): pre-compile the memory WHERE for UPDATE and DELETE scans

semihalev:
Wave B2.5 of the executor perf campaign: partial-pushdown UPDATE/DELETE paid a per-row AST double-hash + program-cache lookup for the memory WHERE via `CompiledEvaluator::evaluate_bool`. The clause now compiles once per statement; each row runs `vm.execute_bool` only, and UPDATE shares a single ExecuteContext between the WHERE check and the SET programs.

## Deliberate non-changes (the audit's caution was warranted)

- **The setter's WHERE re-evaluation stays.** `select_row_ids_for_dml` filters through the full SELECT executor, which runs on a separate snapshot (the statement's transaction is not installed in autocommit), so the per-row re-check against the statement's own snapshot is load-bearing for cross-snapshot consistency. This PR only makes it cheap.
- **Subquery-bearing clauses stay on the evaluator path** — same guard as the SET fast path in #42: the expression compiler emits placeholder subquery ops that cannot run standalone. Clauses that fail to compile also fall back; behavior is byte-identical (`execute_bool` is exactly what the evaluator ran underneath).
- DELETE compiles with the same prefixed column list its evaluator uses (`alias.column` resolution preserved exactly).

## Measurements

| Bench | main | this PR |
|---|---|---|
| 10K-row UPDATE, memory-filter WHERE | 3.72 ms | **3.29 ms** (1.13x) |
| point UPDATE shapes | unchanged | unchanged (different path) |

Per-row WHERE cost drops ~43ns/row on scan-heavy DML.

## Scope

Remaining small audit items for this family (upsert row clone without conflict, DELETE per-row storage calls for RETURNING/FK) are queued behind Wave C — diminishing returns relative to what C holds.

## Verification

Both suites 4916/4916; fmt + clippy -D warnings clean. Existing update/delete/correlated-WHERE tests exercise both the program path and the evaluator fallback.

### DIFF
--- a/src/executor/dml.rs
+++ b/src/executor/dml.rs
@@ -2257,23 +2257,36 @@ impl Executor {
             // Extract params before the closure so they can be captured
             let params = ctx.params();
             let named_params = ctx.named_params();
+            // Pre-compile the memory WHERE once: evaluate_bool re-hashes
+            // the AST per row through the program cache. Subquery-bearing
+            // clauses stay on the evaluator (compiled placeholder subquery
+            // ops cannot run standalone), as do clauses that fail to
+            // compile.
+            let memory_where_program = memory_where_clause
+                .as_ref()
+                .filter(|expr| !Self::expr_contains_subquery(expr))
+                .and_then(|expr| compile_expression(expr, &column_names).ok());
             let mut setter = |mut row: Row| -> Result<(Row, bool)> {
-                // If we need in-memory WHERE filtering, check the condition first
-                if needs_memory_filter {
-                    evaluator.set_row_array(&row);
-                    if let Some(ref where_expr) = memory_where_clause {
-                        match evaluator.evaluate_bool(where_expr) {
-                            Ok(true) => {}
-                            _ => return Ok((row, false)),
-                        }
-                    }
-                }
-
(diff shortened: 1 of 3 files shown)


semihalev:
Both findings closed:

1. **[P1] Transaction id**: both `ExecuteContext`s now carry `.with_transaction_id(ctx.transaction_id())`. In UPDATE the shared context also hands the id to the SET programs, aligning them with the WHERE (the old code's SET context lacked it too). Red-first tests: `UPDATE/DELETE ... WHERE id + 0 = ... AND CURRENT_TRANSACTION_ID() > 0` inside explicit transactions — both silently skipped every row before the fix.
2. **[P2] Registry**: both residual programs compile via `compile_expression_with_context(..., &self.function_registry)`. The discriminating test registers a custom `ABS` (always −1) on an executor via `with_function_registry` and asserts the residual re-check rejects a row the builtin-ABS selection phase passed — proven red with the global-registry compile, green with the fix.

Gates: fmt + clippy clean; both suites 4919/4919; CI-mode `cargo test` passes.

PR #44, opened 2026-08-30
perf(executor): join reorder in place, top-K ORDER BY LIMIT; fix radix NULL order

semihalev:
Wave C1 of the executor perf campaign — plus a NULL-ordering bug the new tests caught on main.

## Perf

1. **JOIN + ORDER BY reorders in place** (the audit big-win): the join path deep-cloned every `(i64, Row)` into a new vector; it now uses the same cycle permutation as the single-table path, and with LIMIT/OFFSET only the needed rows move. 10K-row join sort: 2.63ms → 2.52ms.
2. **ORDER BY + LIMIT top-K**: `select_nth_unstable_by` partitions the needed prefix in O(n), then only that prefix sorts. The NULL-aware multi-column comparator is factored into one helper shared by selection and sort (and no longer duplicated).
3. **Timestamp parse memo**: text-to-timestamp parses in comparisons memoize per thread (failed parses too). Honest scoping note: the audit's headline scenario (`WHERE ts > 'string'` re-parsing per row) turned out to be already neutralized by #40's plan-time coercion; the memo remains as cheap insurance for residual VM-evaluated shapes.

## Correctness — user-visible change, decision yours

The ordering tests exposed that **adding LIMIT flipped NULL placement**: `ORDER BY v DESC` returned `[5, 3, NULL]` without LIMIT but `[NULL, 5]` with `LIMIT 2` — the radix path encoded NULLs as sentinel keys (ASC first / DESC last, SQLite-style) while every comparator path follows the SQL standard (ASC last / DESC first).

Which side is intended? The evidence conflicted: two coverage tests documented the radix behavior as "default in Stoolap", but the **AST contract** (`OrderByExpression.nulls_first`: "None = default — NULLS LAST for ASC, NULLS FIRST for DESC in SQL standard"), the parser comment, and three comparator implementations all say SQL standard. I aligned the radix path to the AST contract: NULL rows are partitioned and placed per the default (also fixing the sentinel collision where real `i64::MIN`/`i64::MAX` values tied with NULLs), the multi-column composite radix falls back to comparison sort when NULLs are present, and the two coverage tests now assert the contract.

One more data point from CI: the differential oracle's feature-gated NULL-ordering query also pinned SQLite's default and failed. It now asserts the two explicit `NULLS FIRST`/`NULLS LAST` forms, which both engines execute identically, so the oracle stays strong on ordering while the defaults are allowed to differ (PostgreSQL has the same default as the AST contract).

**If you prefer the SQLite convention instead** (NULLs smallest everywhere), say so — then the comparators, AST doc, and parser comment change and the radix stays. Either way this belongs in the v0.4.1 release notes as a behavior note for no-LIMIT ORDER BY with NULLs.

## Verification

- New tests: top-K path compared row-by-row against the full sort (ties, OFFSET, NULLS), join reorder integrity in both branches (permutation and take-path)
- Both suites 4921/4921; CI-mode `cargo test` passes on the new file; fmt + clippy clean

## Adversarial review round (002e2511)

The reviewer could not refute the four changes above (cycle permutation direction, mem::take safety, select_nth semantics, radix partition all verified), but found a family of **pre-existing LIMIT/OFFSET bugs on the exact lines this PR touches**. All fixed with red-first tests in `tests/order_limit_test.rs`:

1. **P1: INL join ORDER BY + OFFSET applied OFFSET twice.** The join block applied `skip(offset).take(limit)` inline, then returned `limit_offset_applied = false`, so the standard path wrapped the rows in `TopNResult` and drained OFFSET again (`LIMIT 10 OFFSET 5` over 30 rows returned 5 wrong rows). The block now reports the flag truthfully when the inline sort + LIMIT/OFFSET fully covered the statement. Side effect: the redundant second sort is gone, and the join ORDER BY tests now pin the inline permutation for real.
2. **JOIN + LIMIT + OFFSET without ORDER BY undercounted.** All three early-termination sites (batch INL, JoinExecutor, streaming hash join) collected only LIMIT rows while OFFSET is skipped after collection (returned 5 rows instead of 10). The collection budget is now LIMIT+OFFSET; a non-evaluable OFFSET disables the pushdown.
3. **INL inline sort produced garbage order for `ORDER BY <select-alias>`.** Key-evaluation errors were swallowed into NULL keys (broken on main too). Errors now defer ordering to the standard path, and aggregation/window queries no longer enter the inline sort/truncation at all.

Also from the review: the timestamp memo is now a fixed two-slot round-robin array (miss cost: two comparisons, no Vec shift).

**Follow-ups the review flagged (pre-existing, not touched here):**
- `try_order_by_index_optimization` returns B-tree order = NULLS FIRST always and silently ignores explicit `NULLS FIRST`/`NULLS LAST` on the index-ordered fast path.
- Window int/float sort fast paths use MAX sentinels (real `i64::MAX`/`f64::MAX` values tie with NULLs) and window ORDER BY ignores `nulls_first` entirely.

Both suites 4925/4925 after the round; CI-mode passes; fmt + clippy clean.

### DIFF
--- a/src/core/value.rs
+++ b/src/core/value.rs
@@ -414,7 +414,7 @@ impl Value {
         // Timestamp ↔ Text: try parsing the text side as a timestamp
         match (self, other) {
             (Value::Timestamp(ts), Value::Text(s)) => {
-                if let Ok(parsed) = parse_timestamp(s) {
+                if let Some(parsed) = parse_timestamp_for_compare(s) {
                     return Ok(ts.cmp(&parsed));
                 }
             }
@@ -423,12 +423,12 @@ impl Value {
             {
                 // SAFETY: Json data is always valid UTF-8
                 let s = std::str::from_utf8(&data[1..]).unwrap_or("");
-                if let Ok(parsed) = parse_timestamp(s) {
+                if let Some(parsed) = parse_timestamp_for_compare(s) {
                     return Ok(ts.cmp(&parsed));
                 }
             }
             (Value::Text(s), Value::Timestamp(ts)) => {
-                if let Ok(parsed) = parse_timestamp(s) {
+                if let Some(parsed) = parse_timestamp_for_compare(s) {
                     return Ok(parsed.cmp(ts));
                 }
             }
@@ -437,7 +437,7 @@ impl Value {
             {
                 // SAFETY: Json data is always valid UTF-8
                 let s = std::str::from_utf8(&data[1..]).unwrap_or("");
-                if let Ok(parsed) = parse_timestamp(s) {
+                if let Some(parsed) = parse_timestamp_for_compare(s) {
                     return Ok(parsed.cmp(ts));
                 }
             }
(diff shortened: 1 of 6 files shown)


semihalev:
Both P1s fixed in c8878f75, each with a red-first test in tests/order_limit_test.rs:

1. **Positional ORDER BY**: a bare integer literal in ORDER BY now defers the entire inline join sort to the standard path's ordinal resolver (same treatment as SELECT aliases). Test: `ORDER BY 1 DESC LIMIT 10 OFFSET 5` on an INL join returned the wrong prefix, now resolves to the first output column.
2. **DESC NULL double-reverse**: the hand-rolled closure comparator is gone; the inline sort now calls `compare_by_sort_keys_impl`, the exact comparator the standard path uses, so NULL placement never passes through the DESC reversal. Test pins default DESC (NULLS FIRST) plus explicit `NULLS LAST` and `NULLS FIRST` under DESC. The orphaned `compare_values` import is removed.

Gates: both suites 4927/4927, CI-mode 8/8 on the new file, differential oracle 9/9, fmt + clippy clean.

PR #46, opened 2026-08-31
perf(executor): whole-program cow dispatch, parallel filter path collect

semihalev:
Wave C3 of the executor perf campaign: the execute_cow bail and the parallel memory-filter drain.

## 1. execute_cow whole-program dispatch (vm.rs, program.rs)

The borrowed-value interpreter ran a program until the first op its loop does not handle (the fused `Ne/Ge/Le/IsNull/Like/InSet/Between` comparisons, arithmetic, `Dup`, `NativeFn1`...), then restarted the WHOLE program in the owned VM. Every cow-handled op before the bail point evaluated twice per row: `WHERE a = 1 AND b >= 2` paid the equality twice.

`Program` now precomputes a `cow_supported` flag at build time (mirroring the cow loop's op set), and `execute_cow` dispatches unsupported programs to the owned VM before touching the stack. The cow catch-all stays in place, so drift between the flag and the loop only costs performance, never correctness.

**Why not add the missing ops to the cow loop instead?** I tried that first and measured it: **6-10% SLOWER** on integer-heavy filters. The fused ColumnConst ops read the row directly and never borrow stack values, so the borrowed interpreter's per-op overhead buys nothing there; on main the bail at op 0 was effectively free because the owned VM is faster for those shapes. The dispatch flag keeps every shape on its faster interpreter.

## 2. Parallel memory-filter path double materialization (query.rs)

The no-LIMIT memory-filter path drained an already-materialized scanner row by row (two virtual calls per row) into a `RowVec::new()` with no capacity hint. The fix is gated on `storage_expr.is_none()`: with no storage filter there is no index opportunity, so `table.collect_all_rows` returns the RowVec directly; with a storage filter, `scan()` keeps its PK/index fast paths and the drain loop pre-sizes from `estimated_count`. (The adversarial round caught the first version of this gate: keying on identity projection instead of the filter dropped the MVCC PK/index lookup inside `scan()` and turned partial-pushdown point queries without LIMIT into full scans. Fixed; a 200K-row probe confirms the point shape is back to 1.1us/op.) Side note: the collect branch restores real row ids where the old drain produced id 0 for every row (`take_row_with_id` has no MVCCScanner override); the reviewer verified nothing on this SELECT-only path consumes the id.

## 3. C2 review carry-over

`evaluate_program` gets a `debug_assert` rejecting join-mode evaluators, converting the C2 adversarial reviewer's "comment-only contract" finding into a loud failure in debug builds.

## Numbers (selbench, alternating rounds vs main, 10K rows)

| bench | main | this PR |
|---|---|---|
| mem_filter_mixed (4 conjuncts, residual forces VM) | 345-355us | 316-330us (~7%) |
| parallel_memfilter_scan (no LIMIT, non-pushable WHERE) | 648-722us | 609-618us (6-15%) |
| mem_filter_eq_prefix / order_by_expr_sort | | slightly better, within noise band |

## Adversarial review round

No correctness refutation. The reviewer mechanically diffed the flag's allow-list against the cow loop's match arms (48 vs 48, identical) and compared all 48 shared op implementations line by line: semantically identical. One correction to this PR's own reasoning, worth recording: "bail programs always produced the owned result anyway" is false when short-circuit or CASE control flow skips the unsupported op on a given row; those rows previously got the cow result. Correctness therefore rests on the verified op-level parity, not on the bail argument. Two honest behavior notes from the round: (1) a non-deterministic scalar function in the prefix of a bailing program used to run twice per row and now runs once (a fix); (2) rows whose short-circuit skips the unsupported op used to stay in the borrowed interpreter and now pay owned-VM loads, while the opposite rows stop paying the double prefix; net is workload-dependent, program-wise the double evaluation is gone.

## Verification

- Both suites 4927/4927, differential oracle 9/9, fmt + clippy clean; 140 targeted tests by the reviewer on both backends.
- The dispatch flag changes which interpreter runs, never what it computes (op-level parity verified arm by arm).

### DIFF
--- a/src/executor/expression/evaluator_bridge.rs
+++ b/src/executor/expression/evaluator_bridge.rs
@@ -2049,6 +2049,10 @@ impl<'a> CompiledEvaluator<'a> {
     /// without storing a per-row copy in the evaluator and without
     /// re-hashing the expression. Not for join mode (single row only).
     pub fn evaluate_program(&mut self, program: &SharedProgram, row: &Row) -> Result<Value> {
+        debug_assert!(
+            self.current_row2.is_none() && self.columns2.is_none(),
+            "evaluate_program does not support join mode"
+        );
         let mut ctx = ExecuteContext::new(row);
         if !self.params.is_empty() {
             ctx = ctx.with_params(&self.params);
--- a/src/executor/expression/program.rs
+++ b/src/executor/expression/program.rs
@@ -57,6 +57,11 @@ pub struct Program {
     /// Whether this program contains subqueries
     has_subqueries: bool,
 
+    /// Whether every op is handled by ExprVM::execute_cow, so the
+    /// borrowed interpreter never bails mid-program into a full owned
+    /// re-execution
+    cow_supported: bool,
+
     /// Source expression string (for debugging)
     #[cfg(debug_assertions)]
     source: Option<String>,
@@ -83,12 +88,15 @@ impl Program {
             )
         });
 
+        let cow_supported = Self::compute_cow_supported(&ops);
+
         Self {
             ops,
             max_stack_depth,
             needs_outer_context,
             needs_second_row,
             has_subqueries,
+            cow_supported,
             #[cfg(debug_assertions)]
             source: None,
         }
@@ -111,12 +119,15 @@ impl Program {
             )
         });
 
+        let cow_supported = Self::compute_cow_supported(&ops);
+
         Self {
             ops,
             max_stack_depth,
             needs_outer_context,
             needs_second_row,
             has_subqueries,
+            cow_supported,
             #[cfg(debug_assertions)]
             source: None,
         }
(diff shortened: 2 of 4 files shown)


semihalev:
Adversarial round done, one real catch fixed in df293522:

- **P2 (introduced by this PR's first version): the collect_all_rows swap dropped MVCC PK/index fast paths.** The identity-projection gate was tautologically true on this path, so partial-pushdown point queries without LIMIT (`WHERE id = 5 AND <unpushable>`) went from O(1) lookup to a full scan. The gate is now `storage_expr.is_none()` (no filter = no index opportunity = collect wins; filter present = scan() keeps its lookups + pre-sized drain). 200K-row probe: point shape back to 1.1us/op; parallel-scan win (~7%) holds.
- Flag vs cow-loop diffed mechanically: 48 vs 48 ops, identical; all 48 shared implementations compared line by line, semantically identical. The reviewer corrected the PR's own justification (short-circuit rows could previously get the cow result), recorded in the body; correctness rests on the verified op parity, with the catch-all as drift safety net.
- debug_assert unreachable from current callers; Program flag cannot go stale (only optimize() mutates ops and recomputes).

Gates after the round: both suites 4927/4927, oracle 9/9, fmt + clippy clean.

Issue #1, opened 2025-12-13
DOC: incorrect example in the readme


cakedev0:
Hi, I'm not sure, but I think this example in the readme isn't correct / isn't written as intended:

```SQL
-- Bitmap: Low-cardinality columns, efficient AND/OR
CREATE INDEX idx_status ON orders(status) USING BITMAP;
SELECT * FROM orders WHERE status = 'pending' AND priority = 'high';
```

I think the index is meant to be `orders(status, priority)`

semihalev:
Updated README. Thanks.

PR #47, opened 2026-08-31
perf(cte): share recursive working-set rows across iterations

semihalev:
Wave D1 of the executor perf campaign: recursive CTE row sharing. Small diff (20 lines), measured story below.

## The change (cte.rs, recursive loop only)

Each recursive iteration paid three full row-set copies: `working_rows.clone()` into the temp registry, `new_rows.clone()` for the accumulator extension, and the recursive member's `FROM r` reference cloning the working set out of the registry again. Now:

- the working set **moves** into the temp registry (`mem::take`; it is rebuilt from `new_rows` right after),
- the anchor and each iteration's new rows convert to **Shared** storage once, so the accumulator extension and the per-iteration registry reference are Arc bumps per row.

**Recursive fan-out probe (40 iterations x 1000 rows with heap Text): 4.5-5.2ms -> 2.7-2.9ms (~1.6-1.8x).** CTE self-join shape at parity.

## What the audit expected vs what measurement showed

The audit's big-win #1 targeted the single-reference `WITH big AS (...) SELECT ... FROM big WHERE ... LIMIT n` shape (registry clone per reference + no LIMIT short-circuit). Tracing on current main shows that shape **never reaches the CTE registry**: no store, no `execute_query_on_cte_result` call; the planner inlines it. The registry engages only for join, multi-reference, and recursive shapes. So:

- No LIMIT short-circuit was added (the audited path is unreachable for the common shape; I wrote it, proved it dead by trace, removed it).
- Converting **all** stored CTE rows to Shared in `CteRegistry::store` was tried and measured: the CTE self-join shape regressed 5-8% (a per-row Arc allocation at store time with no payback for single-use rows, plus downstream copy-on-write). Reverted; the conversion lives only in the recursive path where the same rows are provably referenced multiple times.

## Verification

- Both suites 4927/4927, recursive_cte_test 19/19, differential oracle 9/9, fmt + clippy clean.
- Pure sharing change: row values are never mutated through the shared handles on this path (the accumulator and registry are read-only consumers; any downstream writer goes through Row's copy-on-write make_mut).

### DIFF
--- a/src/executor/cte.rs
+++ b/src/executor/cte.rs
@@ -425,12 +425,19 @@ impl Executor {
             anchor_columns
         };
 
-        let mut all_rows = Self::materialize_result(result)?;
+        let all_rows = Self::materialize_result(result)?;
 
         // If no anchor rows, return empty result
         if all_rows.is_empty() {
             return Ok((columns, all_rows));
         }
+
+        // Shared rows make every accumulator/working-set/registry copy
+        // an Arc bump instead of a deep row clone
+        let mut all_rows: RowVec = all_rows
+            .into_iter()
+            .map(|(id, row)| (id, row.into_shared()))
+            .collect();
 
         // Current working set (rows from previous iteration)
         let mut working_rows = all_rows.clone();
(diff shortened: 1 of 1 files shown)


semihalev:
Adversarial round clean: no correctness findings. The reviewer ran a differential A/B (origin/main binary vs this branch over ~18 adversarial recursive-CTE statements including write-through probes, multi-member renumbering, UNION/INTERSECT/EXCEPT dedup, 2000-deep recursion): byte-identical output including error messages. All six attack vectors refuted (mem::take window, Shared-row escape, registry aliasing, id renumbering, dedup storage-independence, anchor conversion scope).

One P3 nit fixed in fab84435: the mem::take placeholder checked a buffer out of the thread-local RowVec pool per iteration; now a pool-free empty vec. Gates: both suites 4927/4927, recursive tests 19/19, clippy clean.

PR #49, opened 2026-08-31
perf(subquery): memoized IN set on cache hits; fix result-cache rollback poisoning

semihalev:
Wave D2 of the executor perf campaign: subquery cache costs. Two items, two honest deferrals.

## 1. Memoized IN-subquery hash set (context.rs, subquery.rs)

The non-correlated IN-subquery cache stored `Vec<Value>`; every hit cloned the entire vector (`get_cached_in_subquery`), and the InHashSet consumer then rebuilt a `ValueSet` plus a fresh `CompactArc` per statement. The cache entry now carries a lazily built `CompactArc<ValueSet>`: the first set-request after (re)population builds it once inside the entry, and every later hit is an Arc bump. The `Vec` stays for the callers that genuinely need values (index probing, multi-column IN).

**Repeated IN-subquery statement, 2000 values: 224-227us -> 200-201us (~12%).**

Staleness: DML invalidation pops the whole cache entry and repopulation always writes `(tables, values, None)`, so the memoized set can never outlive its values. New test `in_subquery_cache_test` pins repeat + INSERT/DELETE invalidation + NOT IN, and runs under CI-mode `cargo test` too.

## 2. EXISTS row_counter: tried, reverted by the adversarial round

The no-additional-predicate EXISTS path was switched to the COUNT twin's visibility row counter, but the reviewer showed the counter's cold fallback guards on `has_committed_row` while the fetcher guards on actual visibility: during a concurrent commit of a sealed-row UPDATE/DELETE this widens an existing race into a wrong-EXISTS window. The hot-path bench was flat (~630us both sides), so the fetcher stays. The guard divergence itself exists on main's COUNT path today; follow-up filed.

## 3. Pre-existing P1 fixed in-round: result-cache rollback poisoning

`BEGIN; INSERT; SELECT ... IN (subquery); ROLLBACK` cached the uncommitted view in the IN/scalar/semi-join caches, and ROLLBACK never invalidated, so later autocommit queries returned rolled-back data until unrelated DML. All three caches now bypass read and write inside explicit transactions (`ctx.transaction_id()` gate). Red-first test pins insert and delete rollbacks across all three cache shapes.

## Deferred, with reasons

- **`to_string()` cache keys -> dual hash** (scalar/IN subquery caches): needs a full-fidelity `SelectStatement` hasher where any missed field is a silent cache collision = wrong results. Deserves its own PR with a collision-discrimination test, same as the #36/#39 rounds.
- **Audit big-win #10 (uncorrelated subquery under correlated outer re-executes per row)**: the audit's own verifier flagged the proposed gate as unsound — `is_subquery_correlated` ignores unqualified identifiers by design, but the compiler resolves them against the outer row (`resolve_outer_column` fallback), so gating on it alone could cache a de-facto correlated subquery. Needs a schema-aware resolvability check; design-first item.

## Verification

- Both suites 4929/4929 (new tests included), differential oracle 9/9, fmt + clippy clean; rebased on #48.

### DIFF
--- a/src/executor/context.rs
+++ b/src/executor/context.rs
@@ -111,8 +111,13 @@ pub fn cache_scalar_subquery(key: String, tables: SmallVec<[CompactArc<str>; 2]>
 // Stores (tables_referenced, result) for table-based invalidation.
 // LRU-bounded to prevent unbounded memory growth.
 
-/// Cached IN subquery entry: (tables_referenced for invalidation, result values)
-type InSubqueryCacheEntry = (SmallVec<[CompactArc<str>; 2]>, Vec<Value>);
+/// Cached IN subquery entry: (tables_referenced for invalidation, result
+/// values, lazily built shared hash set for single-column IN)
+type InSubqueryCacheEntry = (
+    SmallVec<[CompactArc<str>; 2]>,
+    Vec<Value>,
+    Option<CompactArc<crate::core::ValueSet>>,
+);
 
 thread_local! {
     static IN_SUBQUERY_CACHE: RefCell<LruCache<String, InSubqueryCacheEntry>> =
@@ -140,7 +145,7 @@ pub fn invalidate_in_subquery_cache_for_table(table_name: &str) {
         // Collect keys to remove (LruCache doesn't have retain)
         let keys_to_remove: Vec<String> = c
             .iter()
-            .filter(|(_, (tables, _))| tables.iter().any(|t| t.eq_ignore_ascii_case(table_name)))
+            .filter(|(_, (tables, _, _))| tables.iter().any(|t| t.eq_ignore_ascii_case(table_name)))
             .map(|(k, _)| k.clone())
             .collect();
         for key in keys_to_remove {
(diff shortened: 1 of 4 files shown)


semihalev:
Adversarial round outcome, both landed in c4639ee0:

1. **EXISTS row_counter change reverted.** The reviewer showed the counter's cold-fallback guard (`!has_committed_row`) diverges from the fetcher's visibility guard, widening a concurrent-commit window on sealed-row updates into a wrong-EXISTS window. The hot bench was flat, so there was nothing worth that risk. (The same guard divergence exists on main's COUNT twin today; filed as a follow-up.)
2. **Pre-existing P1 fixed in-round: subquery result-cache rollback poisoning.** `BEGIN; INSERT; SELECT ... IN (subquery); ROLLBACK` left the IN/scalar/semi-join caches holding the uncommitted view; every later autocommit query returned rolled-back data until unrelated DML. All three caches now bypass read+write inside explicit transactions; red-first test pins all three shapes.

The memoized IN set (~12%) survives untouched: the reviewer empirically verified old-path vs memoized-path parity including NULL-in-set, NULL probe, NOT IN, duplicates, and empty set, plus the cache-entry lifecycle and that the new test genuinely routes through the switched site.

Also surfaced by the review, pre-existing on main (follow-up list): a concrete repro of the known hot-only index leak through this EXISTS probe after CHECKPOINT (index returns 0 rids for sealed rows); `NOT EXISTS` + `COUNT(*)` with an index erroring 'Function not found: COUNT'; EXISTS probe blind to same-tx writes (known class).

Gates after the round: both suites 4929/4929, CI-mode, oracle 9/9, fmt + clippy clean.

semihalev:
Blocker fixed in c958fe33, red-first:

- New test `in_index_fast_path_not_poisoned_by_rollback` uses the exact shape you named (`SELECT id ... WHERE id IN (...)`, no aggregation, PK path) and was red on the previous commit (51 rows after ROLLBACK).
- `try_optimize_in_subquery` now gates both the lookup and the write on `ctx.transaction_id().is_none()`; the in-transaction path skips the cache-key `to_string` entirely.
- Swept every remaining consumer of the result caches this time: scalar read + all three writes are under the gated key, semi-join read/write under the cacheable flag, batch-aggregate cache is per-top-level-query (cleared at query.rs:273), so no cross-statement lifetime.

Title updated (EXISTS row-counter mention dropped since it was reverted). Gates: both suites 4930/4930, CI-mode 3/3, oracle 9/9, fmt + clippy clean.

PR #50, opened 2026-08-31
fix(join): cross-type numeric equi-join keys never matched in hash joins

semihalev:
Wave E1: the cross-type hash-join bucketing bug found by the #40 test round, promised its own PR.

## The bug

`SELECT COUNT(*) FROM a JOIN b ON a.k = b.k` with `a.k INTEGER` and `b.k FLOAT` returned **zero rows** through every hashed join path, while the same predicate matched correctly via comma-join + WHERE, `IN`, and `CAST`. Wrong results on main since the hash-join key layer existed.

## Three roots, one class

The equality side was already exact (`values_equal` uses `i64_eq_f64` since #40). The hash side never let equal keys meet:

1. **Single-Integer fast path** in `hash_row_keys`/`hash_keys_with` bypassed `hash_value` entirely: an Integer build key took `i * K` while the equal Float probe key took the hasher path. Removed; a 10K-row integer-key hash join measures within noise (~1%) of the fast path.
2. **`hash_table.rs::hash_value`** hashed a type discriminant + raw bits. Now delegates to `Value`'s `Hash` impl, which was engineered for exactly this constraint (`Integer(5)` and `Float(5.0)` hash equal, whole-float normalization, exact above 2^53).
3. **`utils::hash_value_into`** (the parallel join path, 10K+ build rows) was an independent copy of the same discriminated hasher; same delegation.

## Verification

- New `cross_type_join_test` (7 tests, CI-mode too): hash path (shuffled keys), sorted-input path, multi-key, LEFT JOIN, fractional float non-match, NULL keys never match, parallel-threshold path (12K rows), and the 2^63 boundary (`9223372036854775807.0` must NOT join `i64::MAX`).
- Both suites 4939/4939, differential oracle 9/9, fmt + clippy clean.
- Perf: integer-key 10K hash join 942-976us (main) vs 948-990us (branch), within noise.

## Adversarial review round (95853516)

The round refuted the first commit's "zero other change" claim twice; both addressed:

- **GROUP BY leak contained**: `hash_value_into` also keys GROUP BY, and the storage grouped-aggregate fast path keys floats by raw bits, so the first commit made GROUP BY path-dependent. Now only `hash_composite_key` (parallel join) delegates; GROUP BY is byte-identical to main. Follow-up decision filed: GROUP BY splits `1` vs `1.0` while DISTINCT merges them (pre-existing inconsistency, SQLite merges; needs a storage fast-path keying redesign).
- **0.0/-0.0 cliff fixed**: shared buckets exposed the streaming (bit-exact) vs parallel (epsilon) verifier disagreement across the 10K-row threshold. Both now use exact IEEE equality + the NaN convention, which also fixes the parallel path dropping Infinity self-joins. Red-first tests pin both shapes.

## Observed, unchanged (follow-up note)

`utils::values_equal` still compares Float-Float with `f64::EPSILON`, inconsistent with the exact-comparison direction of #40, but unreachable through hashed paths (epsilon-close floats hash into different buckets before and after this change). Flagged rather than touched.

### DIFF
--- a/src/executor/hash_table.rs
+++ b/src/executor/hash_table.rs
@@ -308,14 +308,6 @@ pub fn hash_keys_with<'a, F>(key_indices: &[usize], get_value: F) -> u64
 where
     F: Fn(usize) -> Option<&'a Value>,
 {
-    // Fast path for single integer key (most common case: PK joins)
-    if key_indices.len() == 1 {
-        if let Some(Value::Integer(i)) = get_value(key_indices[0]) {
-            // FxHash for single integer - very fast
-            return (*i as u64).wrapping_mul(0x517cc1b727220a95);
-        }
-    }
-
     let mut hasher = FxHasher::default();
 
     for &idx in key_indices {
@@ -336,14 +328,6 @@ where
 /// This is the same algorithm used in utils.rs but kept here to avoid circular deps.
 #[inline]
 pub fn hash_row_keys(row: &Row, key_indices: &[usize]) -> u64 {
-    // Fast path for single integer key (most common case: PK joins)
-    if key_indices.len() == 1 {
-        if let Some(Value::Integer(i)) = row.get(key_indices[0]) {
-            // FxHash for single integer - very fast
-            return (*i as u64).wrapping_mul(0x517cc1b727220a95);
-        }
-    }
-
     let mut hasher = FxHasher::default();
 
     for &idx in key_indices {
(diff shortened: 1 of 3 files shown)


semihalev:
Adversarial round refuted the 'zero other semantic change' claim twice; both addressed in 95853516:

1. **GROUP BY leak contained.** `hash_value_into` turned out to also key GROUP BY, and the storage grouped-aggregate fast path keys floats by raw bits, so the delegation made GROUP BY path-dependent (executor paths merged `1`/`1.0`, storage path split). Reverted `hash_value_into` to its discriminated body; only `hash_composite_key` (sole production consumer: the parallel join) delegates now. GROUP BY behavior is byte-identical to main. **Filed as a follow-up decision:** main's GROUP BY splits `Integer 1` vs `Float 1.0` while DISTINCT merges them (SQLite merges); unifying needs the storage fast-path keying redesigned, so it deserves its own design round.
2. **0.0/-0.0 result cliff fixed.** Equal keys now share buckets, which exposed the two join verifiers disagreeing (streaming bit-exact vs parallel epsilon) across the 10K-row threshold. Both now use exact IEEE equality + the NaN convention, matching `Value` PartialEq and the WHERE evaluator; this also fixes the parallel path dropping `Infinity` self-joins. Red-first tests pin `0.0 = -0.0` and Infinity joins.

Also catalogued by the review, pre-existing: `COUNT(DISTINCT float_col)` hashes by bits (counts 0.0 and -0.0 separately); on the follow-up list.

Gates: both suites 4939/4939, CI-mode 9/9, oracle 9/9, fmt + clippy clean.

semihalev:
Both coverage findings closed in 44aafb93:

1. **Parallel verifier now guarded**: `float_zero_and_infinity_parallel_path` uses 11K build rows (past the 10K threshold) and pins `0.0 = -0.0` + Infinity matches; discrimination proven by temporarily restoring the epsilon arm, which fails the test (Infinity pair dropped, 11001 vs 11002). `verify_composite_key_equality` additionally gets a direct unit test (zero signs, inf, NaN convention, epsilon-close non-match).
2. **Sorted test now reaches merge join**: bumped to 400+400 rows (800 total > MERGE_JOIN_MIN_ROWS 500, per-side above the 200-row nested-loop ceiling); merge selection verified with a temporary trace, then removed.

Gates: both suites 4941/4941, CI-mode 10/10, fmt + clippy clean.

PR #51, opened 2026-08-31
perf(join): reference comparators, shared NULL pads, leaner batch INL

semihalev:
Wave E2 of the executor perf campaign: per-row and per-comparison join costs.

## Changes

1. **Merge join comparators compare by reference** (merge_join.rs): both key Values were cloned per key column per comparison across the whole merge (O(N+M) comparisons), purely to default out-of-range indices to NULL; a match on `(left.get(li), right.get(ri))` handles that without cloning. NULLs-sort-last semantics preserved exactly.
2. **Shared NULL-pad caches** (hash_join.rs, merge_join.rs, nested_loop_join.rs): the cached NULL rows were Owned, so every unmatched OUTER-join row paid a CompactVec allocation + NULL memcpy despite the cache comment saying otherwise. The caches convert to Shared storage once; each clone is an Arc bump.
3. **Batch INL cleanup** (index_nested_loop.rs): the `I64Map<Row>` built from the fetched inner batch was iterated exactly once and never keyed into — deleted, the batch iterates directly (ids already deduplicated). `key_to_outer_indices` stores `SmallVec<[usize; 2]>`, keeping the common unique-key PK-FK case inline instead of allocating a one-element heap Vec per distinct row id.
4. **Lossless float PK keys in BOTH INL operators — a reproduced wrong-results fix.** The adversarial round constructed a reaching shape (the PK INL planner gate has no outer-column type check and the residual filter never carries the join equality): pre-PR, `ON r.val = i.id` with FLOAT `val` joined `5.5` to PK row 5, `1e19` to the `i64::MAX` row, and `NaN` to row 0. The batch operator was guarded in the first commit; the streaming (LIMIT-path) twin carried the identical truncating line and is guarded in the review round. Red-first test `float_key_pk_inl_join_no_truncation` pins both paths.
5. **Hash join build collection pre-sizes** from `estimated_rows()`.

## Numbers (selbench, two alternating rounds vs main)

| bench | main | this PR |
|---|---|---|
| left_join_sparse (10K left, sparse right, NULL pads) | 541-556us | 426-451us (**1.23-1.27x**) |
| merge_join_sorted_5k | 381-407us | 331-353us (**1.15x**) |
| join_order_by (batch INL, 10K) | 2070-2177us | 1866-2014us (~8-10%) |
| hash_join_int_10k | flat | flat (prealloc marginal) |

## Deferred (operator API changes, next join wave)

- Residual filters evaluate only after full composite materialization (join_executor.rs:913); needs RowRef-reading filters or a scratch-row protocol.
- `MaterializedOperator::from_arc` deep-clones genuinely shared Arc row sets (semantic-cache hits); needs an Arc-serving operator.

## Verification

- Both suites 4942/4942, differential oracle 9/9, fmt + clippy clean.
- No semantic changes intended: comparator NULL ordering preserved arm for arm, Shared NULL rows materialize identically through combine (established by the D1 round's Row API audit), INL result construction unchanged apart from iteration source.

### DIFF
--- a/src/executor/operators/hash_join.rs
+++ b/src/executor/operators/hash_join.rs
@@ -491,7 +491,9 @@ impl HashJoinOperator {
             JoinSide::Left => self.left_col_count,
             JoinSide::Right => self.right_col_count,
         };
-        let row = Row::from_values(vec![NULL_VALUE; count]);
+        // Shared storage: every later clone is an Arc bump, not a
+        // CompactVec allocation per unmatched row
+        let row = Row::from_values(vec![NULL_VALUE; count]).into_shared();
         self.cached_null_build = Some(row.clone());
         row
     }
@@ -507,7 +509,7 @@ impl HashJoinOperator {
             JoinSide::Left => self.right_col_count,
             JoinSide::Right => self.left_col_count,
         };
-        let row = Row::from_values(vec![NULL_VALUE; count]);
+        let row = Row::from_values(vec![NULL_VALUE; count]).into_shared();
         self.cached_null_probe = Some(row.clone());
         row
     }
@@ -604,8 +606,9 @@ impl Operator for HashJoinOperator {
             JoinSide::Right => &mut self.right,
         };
 
-        // Collect all build rows
-        let mut build_rows = Vec::new();
+        // Collect all build rows, pre-sized from the operator's estimate
+        // to avoid doubling reallocation on materialized inputs
+        let mut build_rows = Vec::with_capacity(build_op.estimated_rows().unwrap_or(0));
         while let Some(row_ref) = build_op.next()? {
             build_rows.push(row_ref.into_owned());
         }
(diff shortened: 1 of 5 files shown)


semihalev:
Adversarial round upgraded finding 4 and both halves are now closed in cbc306bd:

- **The 'defensive' float-PK guard is a reproduced P1 fix.** The reviewer built a reaching shape: the PK INL planner gate checks only that the inner column is the PK (no outer type check), and the residual filter carries only the inner WHERE, never the join equality, so truncated probes emit unchecked. Live pre-PR results: `5.5` joined PK row 5, `1e19` joined the `i64::MAX` row, `NaN` joined row 0.
- **The streaming INL twin still truncated** (same line, LIMIT path), so the same query returned different rows with vs without LIMIT after the first commit. Guarded identically; red-first test `float_key_pk_inl_join_no_truncation` pins both paths.

Other vectors refuted by the review: merge comparator equivalence arm by arm, Shared NULL pads safe through the whole Row API (copy-on-write, no pointer-identity consumers), `fetch_rows_by_ids` cannot return duplicate ids on either backend (dedup removal safe), prealloc fed only by exact estimators today.

Filed follow-ups (pre-existing): SecondaryIndex INL probes pass the raw outer Value — the PK-index impl rejects non-Integer keys outright, so `5.0` silently MISSES PK 5 on that path (reverse direction of the same family as #50); NestedLoopJoinOperator::estimated_rows returns l*r unclamped (latent prealloc risk).

Gates: both suites 4942/4942, CI-mode 11/11, oracle 9/9, fmt + clippy clean.

PR #52, opened 2026-08-31
perf(aggregation): raw-entry grouping, clone-free distinct tracking

semihalev:
Wave F1 of the executor perf campaign: aggregation per-row costs.

## Changes

1. **One raw-entry loop for slow-path grouping** (aggregation.rs): the all-simple-columns slow path (taken whenever any aggregate carries DISTINCT / FILTER / expression arguments) had four per-column-count branches, each cloning every group-key Value per input row (the 4+ column branch allocated a whole `Vec<Value>` per row), plus a `contains_key` + `entry` double hash per row when a LIMIT was pushed down. All four branches are now one hashbrown `raw_entry_mut` loop mirroring the fast path's existing pattern at line ~2466: hash borrowed row values, compare stored keys against row references, clone only on vacant insert. About 150 lines down to 50.
2. **DistinctTracker checks membership before cloning** (functions/aggregate/mod.rs): `check_and_add` cloned the Value per accumulated row and let `insert` discard the owned key on duplicates; `COUNT(DISTINCT status)` over 1M rows with 5 values built ~1M throwaway clones. One fix covers COUNT/SUM/AVG DISTINCT in both the dynamic and compiled aggregate paths.
3. **Hidden aggregates extract columns, not rows** (aggregation.rs): `ORDER BY SUM(x)` with `SUM(x)` not in SELECT cloned the entire aggregated result (`having_columns.clone(), having_rows.clone()`) to read back one column per hidden aggregate afterwards. The needed columns now extract into per-column buffers in one pass and the rows move, matching the non-hidden branch.

## Numbers (selbench, two alternating rounds vs main, 10K rows / 20 text groups)

| bench | main | this PR |
|---|---|---|
| group_by_distinct_slow (`COUNT(DISTINCT age) GROUP BY city`) | 451-527us | 387-390us (**1.16-1.36x**) |
| count_distinct_text / group_by_hidden_agg | | within noise on these shapes |

Honest scoping: the tracker win needs heavier text values than the 20-char bench strings (clone is an Arc bump there), and the hidden-aggregate win scales with group count (invisible at 20 groups). Both changes strictly remove work; the measured headline is the grouping rewrite.

## Adversarial review round (273b0729)

Two Sev-1 hash-consistency bugs, both fixed with red-first resize-crossing tests (`group_by_resize_test`):

- **Introduced (first commit)**: the raw-entry map's default hasher disagreed with the manual FxHasher hashes; hashbrown rehashes with the map's own hasher on resize, so >112 groups split (113 keys -> 226 rows). Map is Fx-built now and the manual hash length-prefixes exactly like `hash_one(&Vec<Value>)`.
- **Pre-existing on main**: `try_fast_aggregation`'s multi-column path had the mirror bug (Fx map, manual hash missing the slice length prefix): multi-column GROUP BY split past initial capacity — 500 (a,b) groups inserted twice returned 1000 rows with COUNT(*)=1, proven on the merge base. Same fix, so this PR also repairs a live wrong-results bug on main.

Neither was CI-visible: no existing test drove >112 groups through these paths. Sign-off note: the rewrite trades the old AHash (HashDoS-hardened) group maps for deterministic FxHash, following the existing try_fast_aggregation precedent; flag if you want seeded hashing back.

## Deferred (own PRs)

- ROLLUP/CUBE re-groups the full input once per grouping set (algorithm change: group once on the finest set, derive coarser sets).
- Derived-table streaming aggregation handles only Text keys; Integer keys re-execute the whole subquery and materialize (needs the I64Map branch mirrored).

## Verification

- Both suites 4944/4944, differential oracle 9/9, fmt + clippy clean.
- Semantic parity: the raw-entry comparator uses the same Value equality the old maps used (their Hash/Eq were Value's own), the limit check moved from pre-entry to the vacant arm with identical outcomes, and the hidden-aggregate buffers reproduce the old `having_rows.get(row_idx)` positional semantics including the out-of-range NULL fill.

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -2468,8 +2468,13 @@ impl Executor {
 
         for (_, row) in rows {
             // OPTIMIZATION: Hash directly from row references (no clone for hashing)
-            // FxHasher has zero initialization cost unlike AHash
+            // FxHasher has zero initialization cost unlike AHash.
+            // The manual hash MUST equal the map hasher's hash_one(&Vec<Value>),
+            // which length-prefixes the slice: on resize hashbrown rehashes
+            // every entry with the map's own hasher, and a mismatch scatters
+            // the table and splits groups
             let mut hasher = FxHasher::default();
+            hasher.write_usize(group_by_indices.len());
             for &idx in &group_by_indices {
                 if let Some(value) = row.get(idx) {
                     value.hash(&mut hasher);
@@ -3645,13 +3650,18 @@ impl Executor {
             // against borrowed row values, clone key Values only when a
             // new group is actually inserted (same pattern as
             // try_fast_aggregation above)
-            let mut keyed_groups: hashbrown::HashMap<Vec<Value>, Vec<usize>> =
-                hashbrown::HashMap::with_capacity(64);
+            // The map hasher and the manual hash must agree exactly
+            // (hash_one(&Vec<Value>) length-prefixes the slice): hashbrown
+            // rehashes with the map's own hasher on resize
+            type FxBuild = BuildHasherDefault<FxHasher>;
+            let mut keyed_groups: hashbrown::HashMap<Vec<Value>, Vec<usize>, FxBuild> =
+                hashbrown::HashMap::with_capacity_and_hasher(64, FxBuild::default());
             let num_cols = column_indices.len();
             let mut group_count: usize = 0;
 
             for (row_idx, (_, row)) in rows.iter().enumerate() {
                 let mut hasher = FxHasher::default();
+                hasher.write_usize(num_cols);
                 for &idx in &column_indices {
                     match row.get(idx) {
                         Some(value) => value.hash(&mut hasher),
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -31,7 +31,7 @@ use rustc_hash::{FxHashMap, FxHashSet, FxHasher};
 // SmallVec removed - Vec is faster due to spilled() check overhead in hot loops
 
 use crate::common::{CompactArc, CompactVec, I64Map, StringMap};
-use crate::core::{Result, Row, RowVec, Value, ValueMap};
+use crate::core::{Result, Row, RowVec, Value};
 use crate::functions::aggregate::CompiledAggregate;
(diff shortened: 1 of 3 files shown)


semihalev:
Adversarial round found two Sev-1 hash-consistency bugs; both fixed in 273b0729 with red-first resize-crossing tests:

1. **Introduced by this PR's first commit**: the raw-entry map used hashbrown's default hasher while lookups used manual FxHasher hashes. hashbrown rehashes with the map's OWN hasher on resize, so past 112 groups the table scattered and groups split (113 distinct keys returned 226 rows, deterministic). Map is now FxHasher-built and the manual hash length-prefixes exactly like `hash_one(&Vec<Value>)`.
2. **Pre-existing on main**: `try_fast_aggregation`'s multi-column path — the very pattern this PR copied — had the mirror-image bug: Fx map, but the manual hash omitted the slice length prefix. Multi-column GROUP BY with simple aggregates split past the initial capacity: **500 (a,b) groups inserted twice returned 1000 rows with COUNT(*)=1 on main**, proven on the merge base. Same fix.

Neither was visible to CI because no existing test drives more than 112 groups through these paths; `group_by_resize_test` now pins both (resize-crossing group counts with revisited keys).

Also flagged by the review for your sign-off: the rewrite replaces the old AHash-keyed maps (comments cited HashDoS resistance) with deterministic FxHash. Precedent exists (`try_fast_aggregation` has used Fx for the same reason raw_entry needs a compatible manual hash), but it reverses a documented choice — say the word if you want AHash-with-stored-state instead.

Gates after the round: both suites 4944/4944, CI-mode 2/2, oracle 9/9, fmt + clippy clean.

semihalev:
P2 fixed in bbf32f7a: `DistinctTracker` now uses hashbrown `get_or_insert_with` — one hash, one probe, clone only on actually-new values. Validated on both extremes as requested: high-cardinality unique text (`COUNT(DISTINCT name)`, 10K unique heap strings) is at parity with main across alternating rounds (no double-hash penalty), duplicate-heavy keeps the clone savings.

Bonus: the set is built with a seeded `ahash::RandomState`, which restores the HashDoS hardening for the distinct tracker that the body's sign-off note mentioned.

Gates: both suites 4944/4944, oracle 9/9, fmt + clippy clean.

PR #53, opened 2026-08-31
perf(window): O(n) PERCENT_RANK and CUME_DIST, hoisted constant arguments

semihalev:
Wave G1 of the executor perf campaign: window function costs.

## Changes

1. **PERCENT_RANK / CUME_DIST: O(n^2) -> O(n)** (window.rs, all three partition-compute variants). PERCENT_RANK rescanned every preceding row per row and CUME_DIST scanned forward through its peer group per row. Both now read the precomputes that already existed in the same callers: PERCENT_RANK = (rank-1)/(n-1) with rank from `precompute_rank_info_columnar` (its gate now includes PERCENT_RANK), CUME_DIST = `peer_group_ends[i]/n`. The two per-row scan helpers are deleted.
2. **Constant argument hoisting** for LEAD/LAG (offset), NTILE (bucket count), NTH_VALUE (n): these compiled `ExpressionEval` per row against empty columns. They resolve once before the row loop now. LAG/LEAD's default is different: it can reference row columns, so it compiles once and still **evaluates** per row, unconditionally, preserving the old error surface exactly.
3. **Accumulate by reference**: the three aggregate frame loops cloned every value (heap clone for Text) just to pass `&value`; they bind references now (static NULL + a `COUNT_ONE` const for `COUNT(*)`); `accumulate` clones internally only when it retains.
4. **Extended-row buffer moves** instead of double-cloning at the three expression-wrapping sites (`Row::from_compact_vec(std::mem::take(&mut ext_values))`).

## Numbers (selbench vs main, 10K rows)

| bench | main | this PR |
|---|---|---|
| window_percent_rank_10k (single partition) | **221.2ms** | **338us (~650x)** |
| window_lag_offset_10k (`LAG(v, 2, 0)`) | 1689us | 421us (**4.0x**) |
| window_max_text_10k (text agg over frames) | 605us | 567us (~7%) |
| window_row_number | 441us | 420us (~5%) |

## Adversarial review round (f34106c7)

No P1; a 4-battery differential oracle against main was byte-identical except two edge deltas:

- **Fixed**: hoisted constant-argument evaluation errored on empty tables where the per-row scheme vacuously succeeded; all three variants now early-return for empty partitions (red-first test).
- **Disclosed, kept**: PERCENT_RANK now always equals (RANK-1)/(n-1). Old and new provably coincide wherever the window sort groups equal values adjacently (every consistent comparator path); they differ only where the sort was already broken, and there the old value contradicted the engine's own RANK.
- **Filed pre-existing window sort bugs, now with repros**: integer sort fast path keys NULL and non-Integer values as i64::MAX (collides with real i64::MAX data), NaN treated Equal-vs-everything in mixed-type comparator, type drift past the 100-row sample, lossy `as f64` above 2^53.

## Deferred (G2)

- RANGE-offset frame bounds still scan the partition per row (binary search over the sorted keys + sliding accumulators for invertible aggregates); needs DESC-direction care.
- Partition map + sort are rebuilt once per window function sharing the same OVER clause; the cache mirrors order_by_cache but touches shared plumbing.
- Base-column materialization clones every value (signature change through five call sites).
- From the E1 review, still open: window ORDER BY ignores NULLS FIRST/LAST entirely, and the int/float sort fast paths use MAX sentinels that tie real MAX values with NULLs.

## Verification

- Both suites 4945/4945, differential oracle 9/9, window test batteries green, fmt + clippy clean.
- Parity notes: PERCENT_RANK's old "position of first row with equal ORDER BY value" is exactly rank_info's group_start + 1; CUME_DIST's forward peer scan is exactly the peer-group end; both no-ORDER-BY edge cases (all rows one peer group) produce 0.0 / 1.0 as before; the three compute variants received identical edits.

### DIFF
--- a/src/executor/window.rs
+++ b/src/executor/window.rs
@@ -38,6 +38,9 @@ use std::cmp::Ordering;
 use crate::common::{CompactVec, StringMap};
 use crate::core::row_vec::RowVec;
 use crate::core::value::NULL_VALUE;
+
+/// Stack constant for COUNT(*) accumulation by reference
+const COUNT_ONE: Value = Value::Integer(1);
 use crate::core::{Error, Result, Row, Value};
 
 /// Type alias for partition keys - stack-allocated for common case (up to 4 columns)
@@ -587,7 +590,9 @@ impl Executor {
                                 .unwrap_or(NULL_VALUE);
                             ext_values.push(wf_value);
                         }
-                        let ext_row = Row::from_compact_vec(ext_values.clone());
+                        // Move the buffer into the row instead of cloning it; the
+                        // next iteration's extend refills a fresh one
+                        let ext_row = Row::from_compact_vec(std::mem::take(&mut ext_values));
 
                         if let Ok(ext_result_values) = eval.eval_all(&ext_row) {
                             for (eval_idx, (item_idx, _)) in ext_expr_items.iter().enumerate() {
@@ -814,7 +819,9 @@ impl Executor {
                                     .unwrap_or(NULL_VALUE);
                                 ext_values.push(wf_value);
                             }
-                            let ext_row = Row::from_compact_vec(ext_values.clone());
+                            // Move the buffer into the row instead of cloning it; the
+                            // next iteration's extend refills a fresh one
+                            let ext_row = Row::from_compact_vec(std::mem::take(&mut ext_values));
                             eval.eval(&ext_row)?
                         }
                     };
@@ -1025,7 +1032,9 @@ impl Executor {
                                     .unwrap_or(NULL_VALUE);
                                 ext_values.push(wf_value);
                             }
-                            let ext_row = Row::from_compact_vec(ext_values.clone());
+                            // Move the buffer into the row instead of cloning it; the
+                            // next iteration's extend refills a fresh one
+                            let ext_row = Row::from_compact_vec(std::mem::take(&mut ext_values));
                             eval.eval(&ext_row)?
                         }
                     };
@@ -1964,7 +1973,6 @@ impl Executor {
         skip_sorting: bool,
     ) -> Result<(Vec<Value>, Vec<usize>)> {
         // Suppress unused variable warnings - these are needed for compute_lead_lag, compute_ntile, etc.
-        let _ = columns;
 
         // Empty fallback for when no precomputed values are provided
         let empty_order_by = ColumnarOrderByValues {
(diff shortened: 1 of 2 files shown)


semihalev:
Adversarial round outcome, landed in f34106c7:

- **No P1.** The reviewer ran a 4-battery differential oracle (old vs new engine, ~13K output lines across sequential/parallel/streaming/lazy paths): byte-identical except the two disclosed deltas below.
- **Fixed: empty-table hoisted evaluation.** `SELECT LEAD(v, grp) OVER (ORDER BY id) FROM empty_t` errored on the first commit where it vacuously succeeded before; all three variants now early-return for empty partitions. Red-first test pins LEAD/NTILE/NTH_VALUE on empty tables.
- **Disclosed, kept: PERCENT_RANK is now always (RANK-1)/(n-1).** The old scan and the new precompute provably coincide wherever the window sort groups equal values adjacently, which holds on every consistent comparator path. They differ only where the sort itself is already broken, and there the old PERCENT_RANK contradicted the engine's own RANK while the new one matches it definitionally.
- **Filed (pre-existing, now with repros)**: the window integer sort fast path keys NULL and any non-Integer value as i64::MAX, colliding with genuine i64::MAX data (RANK/ordering wrong on both builds); compare_values_fast treats NaN as Equal-vs-everything in mixed-type ORDER BY expressions; type drift past the 100-row detection sample; lossy `as f64` above 2^53. These join the existing window follow-up list (NULLS FIRST/LAST ignored, MAX-sentinel ties).

Gates after the round: both suites 4945/4945, CI-mode 26/26, oracle 9/9, fmt + clippy clean.

semihalev:
Both P2s closed in c8b09ef4, red-first where testable:

1. **Volatile arguments evaluate per row again.** New `ConstArg`: compile once; memoize the result only when the expression is deterministic, with the function registry's `is_deterministic` as the source of truth and anything unrecognized conservatively treated as volatile. LEAD/LAG offsets, NTILE bucket counts, and NTH_VALUE's n follow this in all three partition-compute variants, so RANDOM-style arguments keep the old per-row behavior while the per-row compile stays hoisted.
2. **NTH_VALUE's n is fully lazy behind the frame check** (compile included). Your exact shape is the red-first test: single-row partition + `ROWS BETWEEN 1 FOLLOWING AND 1 FOLLOWING` + invalid argument returns NULL where the previous commit errored at compile. Deterministic n memoizes at the first non-empty frame; volatile n resolves per row like the old code.

Perf held: PERCENT_RANK ~650x unchanged, LAG(v, 2, 0) at 3.7x (the small give-back from the memo check is the price of volatility correctness).

Gates: both suites 4946/4946, CI-mode 27/27, oracle 9/9, fmt + clippy clean.

PR #54, opened 2026-09-01
fix(window): sort keys stop tying NULLs with MAX values

semihalev:
G2-correctness: the three window ORDER BY sort bugs the #53 adversarial round reproduced.

## Fixes (window.rs, all with red-first RANK tests)

1. **NULL vs MAX sentinel collision.** The integer/float sort fast paths keyed NULLs as `i64::MAX` / `f64::MAX`, so real MAX values tied with NULLs: `ORDER BY k` over `(NULL, i64::MAX, NULL, 5, i64::MAX, NULL)` produced RANK 1..6 instead of `[1, 2, 2, 4, 4, 4]`. NULL indices now stay **out of band** and are written after the values ascending / before them descending (exactly the placement the old sentinels produced), while the value sort keeps its bare-u64 keys, so the hot comparison is unchanged.
2. **Type drift past the detection sample.** A value of a different type appearing after the 100-row sample got the sentinel key and corrupted the ordering. The sample is a hint now; the typed extractors return false on the first non-matching value and the generic sort takes over.
3. **Comparator total order.** `compare_values_fast` mapped NaN to Equal-vs-everything (an inconsistent comparator handed to sort_unstable) and compared Integer vs Float through a lossy `as f64`. NaNs now compare equal to each other and after every real (matching the float key path and the NULLS-after-reals convention), and mixed numeric compares use the exact `cmp_i64_f64`.

## Perf

Value-class sorting is unchanged (bare u64 keys, no tuple): benches with no NULL keys measure at parity with main within the noise band. The out-of-band NULL block skips sorting NULL peers entirely.

## Verification

- Both suites 4949/4949, differential oracle 9/9, CI-mode 3/3, fmt + clippy clean.
- New `window_sort_keys_test`: i64::MAX-vs-NULL RANK, f64::MAX-vs-NULL RANK, and ordering integrity when the type drifts past the sample.

### DIFF
--- a/src/executor/window.rs
+++ b/src/executor/window.rs
@@ -3436,8 +3436,9 @@ impl Executor {
             keyed.sort_unstable_by_key(|&(_, key)| std::cmp::Reverse(key));
         }
 
-        // Write back: NULLs last ascending, first descending (the
-        // placement the old sentinel keys produced)
+        // Write back: NULLs strictly last ascending, first descending.
+        // For floats this also fixes the old sentinel's misplacement of
+        // NULLs before +inf and NaN
         let mut pos = 0;
         if !ascending {
             for &idx in &null_indices {
@@ -3526,8 +3527,9 @@ impl Executor {
             keyed.sort_unstable_by_key(|&(_, key)| std::cmp::Reverse(key));
         }
 
-        // Write back: NULLs last ascending, first descending (the
-        // placement the old sentinel keys produced)
+        // Write back: NULLs strictly last ascending, first descending.
+        // For floats this also fixes the old sentinel's misplacement of
+        // NULLs before +inf and NaN
         let mut pos = 0;
         if !ascending {
             for &idx in &null_indices {
--- a/src/executor/window.rs
+++ b/src/executor/window.rs
@@ -3376,7 +3376,10 @@ impl Executor {
             (Value::Null(_), Value::Null(_)) => Ordering::Equal,
             (Value::Null(_), _) => Ordering::Greater, // NULLs last
             (_, Value::Null(_)) => Ordering::Less,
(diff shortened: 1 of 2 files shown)


semihalev:
Adversarial round could not refute the fix (10 CLI repros incl. 60K-row parallel paths, 360 targeted tests green). Two low-severity items landed in 0eec28a7:

- **Drift test now drives the bail path**: the old test sampled Text and took the generic sort on both builds; the new shape (120 integers, Float 0.5 past the 100-row sample, RANK assertions) fails pre-fix per the reviewer's CLI proof and passes post-fix.
- **Comment + disclosure**: the float write-back comment claimed exact sentinel parity, but the new placement deliberately puts NULLs after +inf/NaN, fixing the old sentinel's misordering (NULLs used to sort before infinity). Counts as a third fixed bug; belongs in the v0.4.1 release notes with the others.

Also catalogued pre-existing (unchanged): the heterogeneous numeric-vs-Text comparator fallback is still intransitive, and the differential oracle has no window coverage.

Gates: default suite green, CI-mode 3/3, fmt + clippy clean.

semihalev:
P2 closed in 2ee8797b, red-first: the exact cycle you named (`2 / 10 / '15'` via CASE, 150 rows) put bucket-0 rows at rank 4 pre-fix; the fallback now uses `Value::cmp` after the NULL arms (type rank, then within-type; numerics before text, no stringly comparison). The fast numeric/NaN arms match Value's own numeric ordering exactly (`cmp_i64_f64`, NaN-last), so the combined relation is a genuine total order. Test pins the block ranks [1, 51, 101].

Gates: both suites 4950/4950, CI-mode 4/4, oracle 9/9, fmt + clippy clean.

PR #55, opened 2026-09-01
perf(window): share the partition map and sort across one OVER clause

semihalev:
Wave G2 of the executor perf campaign: share partition work across window functions.

## Change

Each window function called `build_partition_map` itself and then sorted every partition by its ORDER BY, so `SELECT ROW_NUMBER() OVER w, RANK() OVER w, SUM(x) OVER w, LAG(x) OVER w` paid four hash-grouping passes over all rows and four full sets of partition sorts.

Partitions now resolve **once per distinct OVER clause**: `shared_sorted_partitions` builds the map (or reuses the planner's pre-grouped map), sorts each partition once, and stores an `Arc<Vec<Vec<usize>>>` in a per-statement cache keyed by a semantic `(PARTITION BY, ORDER BY)` string. The three per-partition compute variants take an `already_sorted` flag so they skip the redundant sort. The aggregate window path uses the same cache.

## Numbers (selbench vs main, 10K rows / 80 partitions)

| bench | main | this PR |
|---|---|---|
| window_shared_over_x4 (4 functions, one OVER) | 1421-1541us | 889-927us (**1.60-1.66x**) |
| percent_rank / lag / max_text / row_number (single function) | | parity |

## Scoping

- The **presorted-by-index** aggregate path still builds its own map: the shared entry is sorted by definition, and that path exists precisely to avoid sorting.
- The **streaming-LIMIT** and **lazy-partition** paths keep local caches: their row indices are partition-local, so a statement-wide cache would key identical OVER clauses onto different index spaces.
- Partition *iteration order* changes from HashMap order to the collected Vec order. Both are arbitrary and no window result depends on it (each partition writes only to its own row indices).

## Verification

- Both suites 4950/4950, window batteries (function/advanced/coverage/sort-keys) 92/92, differential oracle 9/9, fmt + clippy clean.

### DIFF
--- a/src/executor/window.rs
+++ b/src/executor/window.rs
@@ -394,9 +394,13 @@ impl Executor {
 
         // OPTIMIZATION: LIMIT pushdown for PARTITION BY queries
         // Only safe when there is no top-level ORDER BY (otherwise the sort must see all
-        // rows before LIMIT can be applied) and when all window functions share the same
-        // PARTITION BY so one partition map can serve them all.
-        if stmt.order_by.is_empty() && stmt.offset.is_none() {
+        // rows before LIMIT can be applied), when all window functions share the same
+        // PARTITION BY so one partition map can serve them all, and when no OVER clause
+        // is volatile (RANDOM and friends must draw per window function).
+        let all_over_clauses_shareable = window_functions
+            .iter()
+            .all(|wf| self.over_clause_is_shareable(wf));
+        if all_over_clauses_shareable && stmt.order_by.is_empty() && stmt.offset.is_none() {
             if let Some(limit_expr) = &stmt.limit {
                 let has_partition_by = window_functions
                     .iter()
@@ -4342,6 +4346,7 @@ impl Executor {
 
 #[cfg(test)]
 mod tests {
+
     use super::*;
     use crate::storage::mvcc::engine::MVCCEngine;
     use std::sync::Arc;
--- a/src/executor/window.rs
+++ b/src/executor/window.rs
@@ -78,15 +78,20 @@ impl ConstArg {
     }
 }
 
-/// Conservative determinism check: anything unrecognized counts as
-/// volatile and re-evaluates per row
+/// Conservative determinism check: does this expression yield the same
+/// value every time it is evaluated against the same row? Column
+/// references and bound parameters qualify; RANDOM and friends do not,
+/// and anything unrecognized counts as volatile.
 fn expr_is_deterministic(expr: &Expression, registry: &FunctionRegistry) -> bool {
     match expr {
         Expression::IntegerLiteral(_)
         | Expression::FloatLiteral(_)
         | Expression::StringLiteral(_)
         | Expression::BooleanLiteral(_)
-        | Expression::NullLiteral(_) => true,
+        | Expression::NullLiteral(_)
+        | Expression::Identifier(_)
+        | Expression::QualifiedIdentifier(_)
+        | Expression::Parameter(_) => true,
         Expression::Prefix(prefix) => expr_is_deterministic(&prefix.right, registry),
         Expression::Infix(infix) => {
             expr_is_deterministic(&infix.left, registry)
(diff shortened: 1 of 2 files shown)


semihalev:
Both P1s fixed in 52cd67de, red-first with your exact repros:

- **Parameter collision**: `PARTITION BY v % ?` twice with bindings (2, 3) returned the v%2 partitioning for both; now `[1,1,2,2,3,3]` and `[1,1,1,2,2,2]`.
- **Separator injection**: `PARTITION BY "a,b"` vs `PARTITION BY a, b` shared a key; now the (a,b) pairs each form their own partition.

Root cause was keying on rendered SQL. Both caches now key structurally: partition expressions via `compute_expression_hash` (which folds parameter index and name, exactly as the compiler and evaluator bridge do) and ORDER BY items via (expression hash, ascending, nulls_first). This also closes the same `?` collision in the **pre-existing** `order_by_cache` you identified on main.

P3s also addressed: the stale doc comment and `#[inline]` that had attached themselves to the new function are back on `order_by_cache_key`, the duplicated `allow` attributes are gone, and the aggregate path now borrows the shared partitions instead of cloning each index vector (trimming the single-WF tax you measured).

Perf after the round: shared-OVER x4 still ~1.7x (1759-2034us -> 1019-1059us), single-function shapes at parity.

Also filed from your incidental finding: `mvcc/table.rs:3681` `(limit + offset) * 2 + 100` overflows on the no-LIMIT sentinel and panics in debug on the pre-sorted window path (pre-existing, both branches).

Gates: both suites 4952/4952, CI-mode 2/2, oracle 9/9, fmt + clippy clean.

semihalev:
All three points closed in 9ac0d48b:

- **P1 (hash used as identity)**: the cache now keys on **(expression hash, rendered expression)** per item and a hit requires both to match. Hash alone can collide, rendered SQL alone drops positional parameter indices; together they cover both classes. Your literal-collision shape is in the suite as a regression test (it passed on the previous commit too, so your probe likely hit a different path, but the defense is unconditional now).
- **P2 (volatile partitions)**: OVER clauses containing non-deterministic expressions bypass the cache on read and write. Red-first test: two `PARTITION BY RANDOM() < 0.5` windows over 200 rows must not produce identical row numbers.
- **Side note (all_partitions_match)**: the streaming-LIMIT gate now uses the same structural identity, so different `?` bindings are no longer treated as one partitioning.

Worth flagging from this round: the determinism walker I reused (written for constant arguments) treated column references as volatile, which silently disabled sharing altogether. Tests stayed green because the behavior was correct, only the optimization was dead. **The benchmark caught it** — column refs, qualified names and bound parameters are deterministic per row, and the constant-argument path is unaffected since it compiles against no columns. Shared-OVER x4 is back to 1.67-1.72x with single-function shapes at parity.

Gates: both suites 4954/4954, oracle 9/9, fmt + clippy clean.

semihalev:
Both points closed in 8d0e1335:

1. **LIMIT streaming path guarded** — you were right that it bypasses `shared_sorted_partitions` entirely. The streaming gate now also requires every OVER clause to be shareable, so volatile partitions fall back to the normal path. Discrimination proven: with the guard temporarily removed the new test fails, with it restored it passes.
2. **Collision test sharpened** with the i64::MIN row as you described. Two findings from doing it:
   - **Pre-existing panic**: inserting `i64::MIN` as a PRIMARY KEY panics with `i64::MIN cannot be used as a key in I64Map (reserved as empty sentinel)` — valid SQL, hard panic. The test uses a non-PK column instead; filing this separately.
   - **The two literals do not actually collide in this build**: I measured the hashes (`4106119473615583154` vs `9561797097096359709`) and confirmed by temporarily switching the key back to hash-only, where the test still passes. So the test pins partition semantics, not collision discrimination. The rendered-form half of the key stays as unconditional defense since a hash is not an identity regardless.

Gates: both suites 4955/4955, CI-mode 5/5, oracle 9/9, fmt + clippy clean.

PR #57, opened 2026-09-01
fix: close the remaining i64::MIN and unbounded-LIMIT panics

semihalev:
Follow-up to #56. The adversarial round on that PR showed its framing was wrong: guarding INSERT closed one door of a class that stays open in four other places, all reachable from plain SQL. This closes the rest.

## Panics fixed (each with a red-first test)

1. **PK lookup of i64::MIN inside a transaction.** After the transaction had written any row, `SELECT/UPDATE/DELETE ... WHERE id = -9223372036854775808` panicked in `I64Map::get`. The PK fast path declines the sentinel, and the transaction-local accessors (`get`, `get_local_version`, `has_locally_seen`) answer "not found" without touching the map, so callers that bypass the fast path are safe too. Notably #56's error message made this worse in effect: a user told the value is rejected on INSERT would reasonably expect lookups to be handled.
2. **`IN (-9223372036854775808)`.** Building the hashed integer set inserted the sentinel. It is tracked out of band beside the set, the same shape `pk.rs` already uses. Covered for IN, NOT IN, mixed lists, and the PK index-probe path.
3. **`GROUP BY` over a non-PK INTEGER column holding i64::MIN.** Both grouping fast paths crashed. The storage path routes the sentinel to the generic group map (its Float branch already did exactly this), and the executor's single-column fast path declines so the general path handles it.
4. **Large but legal LIMIT.** `SELECT * FROM t LIMIT 9223372036854775807` aborted the process at `Vec::with_capacity`. Five capacity sites clamp now (three in the version store, two window paths, the TopN heap). Note the heap site is the cautionary one: it already used `saturating_add` and then handed the saturated value straight to `with_capacity`.
5. **AUTO_INCREMENT past i64::MAX.** `fetch_add` panicked in debug; in release it wrapped the counter to i64::MIN, so every later row got a negative id and #56's guard then reported a "primary key out of range" error for a statement that supplied no primary key. The counter is fallible now and reports exhaustion against the column name.

## Review notes also addressed

- The primary-key rejection is an `invalid_argument` naming the value and column, not `Error::internal`.
- The presorted-window test now uses 200 rows: in release the old arithmetic wrapped to a batch size of 98, so the 4-row fixture passed there and only covered the debug panic, not the silent truncation.
- The persistent tests seal cold segments (`PRAGMA CHECKPOINT`) before asserting, so they exercise the segmented path rather than the hot table.

## Still open (deliberately)

`src/storage/index/pk.rs` carries `has_i64_min` machinery documented as existing because `I64Set` reserves the sentinel, so that layer believes i64::MIN is a legal row id while `prepare_insert` rejects it. Resolving that contradiction (either tolerate the sentinel in the row-id maps or delete the dead flag) is a design call, not a panic fix.

## Verification

Both suites 4962/4964, differential oracle 9/9, CI-mode 9/9, fmt + clippy clean.

### DIFF
--- a/src/common/i64_map.rs
+++ b/src/common/i64_map.rs
@@ -895,6 +895,10 @@ pub struct I64Set {
     slots: Box<[i64]>,
     len: usize,
     mask: usize,
+    /// i64::MIN is the slot array's empty sentinel, so it is stored out
+    /// of band. Callers pass user data (join keys, IN lists), so the set
+    /// must accept the full i64 range instead of panicking.
+    has_min: bool,
 }
 
 impl Clone for I64Set {
@@ -943,17 +947,18 @@ impl I64Set {
             slots: slots.into_boxed_slice(),
             len: 0,
             mask: cap - 1,
+            has_min: false,
         }
     }
 
     #[inline(always)]
     pub fn len(&self) -> usize {
-        self.len
+        self.len + usize::from(self.has_min)
     }
 
     #[inline(always)]
     pub fn is_empty(&self) -> bool {
-        self.len == 0
+        self.len == 0 && !self.has_min
     }
 
     #[inline(always)]
@@ -1008,7 +1013,11 @@ impl I64Set {
     /// Insert a value into the set. Returns true if the value was newly inserted.
     #[inline(always)]
     pub fn insert(&mut self, key: i64) -> bool {
-        check_key(key);
+        if key == EMPTY {
+            let inserted = !self.has_min;
+            self.has_min = true;
+            return inserted;
+        }
 
         if self.len * LOAD_FACTOR_DEN >= self.slots.len() * LOAD_FACTOR_NUM {
             self.grow();
@@ -1039,7 +1048,9 @@ impl I64Set {
 
     #[inline(always)]
     pub fn contains(&self, key: i64) -> bool {
-        check_key(key);
+        if key == EMPTY {
+            return self.has_min;
+        }
 
         let mask = self.mask;
         let mut idx = Self::hash(key) & mask;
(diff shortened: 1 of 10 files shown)


semihalev:
All four blockers fixed in c3b909ad. I changed approach: patching call sites one round at a time was clearly not converging, so the set half is fixed **inside the data structure**.

- **I64Set now stores i64::MIN out of band** (insert/contains/remove/len/is_empty/iter/drain/into_iter/clear all consistent). That fixes your point 4 at both sites (`index_optimizer` NOT IN exclusion set and `dml.rs` InHashSet) **without touching either call site**, and the old should-panic unit tests became a contract test for the new behaviour.
- **Batch INL** (point 3): a sentinel outer key can never match a primary key, so the probe is skipped before `I64Map::entry`.
- **Two LIMIT capacities** (points 1 and 2): the transaction-local scan path and the CTE indexed-join path are clamped.

I also swept the class rather than waiting for round five: every remaining `I64Map`/`I64Set` is keyed by row or transaction ids that never carry user data, and the volume grouping paths use `ValueMap`/`FxHashMap`, so they were never exposed.

On your point about being fed up with the sentinel: I counted it. **~75 production lines across 7 files exist only to defend the sentinel** (pk.rs alone has 20 `has_i64_min` branches), versus 13 lines of genuine i64/IEEE range math that should stay. My recommendation is to retire the sentinel the same way I just did for the set: a side slot inside `I64Map` plus an `Entry` variant, then delete all 75 lines including pk.rs's parallel machinery. Cost is one comparison that `check_key` already performs today. I can bring that as the next PR with row-id hot-path benchmarks attached.

Gates: both suites 4965/4967, CI-mode 13/13, oracle 9/9, fmt + clippy clean.

semihalev:
P2 fixed in 9d5a4885. Both bounds now include the pending sentinel, not just the upper one: once the slots are exhausted the sentinel is a *guaranteed* remaining item, so the lower bound was wrong too.

Worth flagging how the test evolved: my first version passed against the buggy code, because the bad state (slots exhausted, sentinel pending) only occurs when the last occupied slot happens to be the final one, which depends on hash placement. The committed test constructs that state directly and fails on the old code.

Gates: both suites 4966/4968, oracle 9/9, fmt + clippy clean.

PR #58, opened 2026-09-01
refactor(i64map): retire the i64::MIN sentinel and its defenses

semihalev:
You said you were fed up with the sentinel, so I counted it: **~75 production lines across 7 files existed only to defend `i64::MIN`**, against 13 lines of genuine i64/IEEE range math. Every panic round found another caller that had not been told. This retires the sentinel instead.

## The change

`I64Map` keeps `i64::MIN` in a side slot, exactly as `I64Set` already does since #57, so the map covers the full i64 range. Accessors, iterators, `retain`, `clear` and the whole `Entry` API (occupied and vacant, `insert` and `remove`) route through it. `check_key` and its panic are deleted.

**Hot path**: sentinel access lives behind `#[cold] #[inline(never)]` helpers, so the probe loops keep the exact shape they had when this branch was a panic. That detail is load-bearing, see the numbers.

## Defenses deleted

| Site | What went away |
|---|---|
| `index/pk.rs` | the `has_i64_min` flag and **all 20 branches** (contains/insert/remove/clear/overflow_empty/overflow_min/overflow_max and four collection sites): the index has one code path again |
| `expression/in_list.rs` | the parallel flag beside the hashed set |
| `operators/index_nested_loop.rs` | the sentinel-probe skip |
| `executor/aggregation.rs` | the fast path bailing out when it saw the value |
| `mvcc/version_store.rs` | both grouping branches routing the value away, including the float `-0.0` special case, plus three transaction-local short circuits |
| `mvcc/table.rs` | the `try_pk_lookup` rejection |

## What stays, deliberately

The primary-key rejection. The row-id maps can hold `i64::MIN` now, but the arena and the segment layer still use it as their "no row" marker, so it stays outside the addressable id space; the comment in `prepare_insert` says exactly that. Lifting it is a separate decision about the row-id space.

`CowHashMap` still panics on the sentinel. It has no production consumers (only its own tests), so I left it: either align it or delete it, your call.

## Numbers (best of five interleaved runs vs main)

| bench | main | this PR | delta |
|---|---|---|---|
| prepared_point_pk | 993ns | 973ns | -2.0% |
| hash_join_int_10k | 929us | 905us | -2.5% |
| prepared_update_pk | 751ns | 768ns | +2.3% |
| join_order_by | 2194us | 2243us | +2.3% |

All inside the noise band. Worth recording: **my first attempt inlined the side-slot access into the hot path and cost 11.8% on point lookups.** The cold helpers are what make this free, not the design by itself.

## Found on the way (pre-existing, not fixed here)

`SELECT SUM(id) FROM t GROUP BY v ORDER BY v` ignores the ORDER BY when the grouped column is not projected: with v in {7,1,7,5} it returns sums `[4,4,2]` for both ASC and DESC, where ASC should be `[2,4,4]`. The projected form is correct. It deserves its own red-first fix; I only decoupled a test of mine that had been leaning on the buggy order.

## Verification

Both suites 4965/4967, differential oracle 9/9, CI-mode 13/13, fmt + clippy clean. The map and set panic tests became contract tests: full round trip through the side slot, iterator and clone consistency, and the Entry API on both occupied and vacant sentinel entries.

### DIFF
--- a/src/common/i64_map.rs
+++ b/src/common/i64_map.rs
@@ -13,15 +13,12 @@
 // limitations under the License.
 
 //! High-performance i64 HashMap
-// - Uses i64::MIN as empty sentinel (row IDs and txn IDs are always >= 0)
+// - Uses i64::MIN as the empty marker inside the slot array; the entry
+//   for that key lives in a side slot, so the full i64 range is storable
 // - Direct key storage (no XOR transform)
 // - FxHash with pre-mixing (XOR>>16 before multiply) - 0 sequential collisions,
 //   65% reduction in strided key collisions
 // - Backward-shift deletion (no tombstones)
-//
-// SAFETY: i64::MIN CANNOT be used as a key - it is reserved as the empty sentinel.
-// Inserting i64::MIN will cause silent data corruption. This is enforced via
-// assertions in debug and release builds.
 
 use std::mem::MaybeUninit;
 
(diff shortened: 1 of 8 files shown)


semihalev:
Blocker fixed in $(git rev-parse --short HEAD), and it had a sibling.

- **Your finding**: `I64Map::drain()` moved only the slot array, so a map holding `i64::MIN` drained without that entry and was not empty afterwards. `Drain` now carries the side slot, yields it last, counts it in `size_hint`, and drops it when the iterator is dropped early. Red-first test covers the full drain and a drain dropped after one item; reverting the one-line take makes it fail.
- **Found while checking the sibling path**: `I64Set::drain()` has a pre-existing version of the same class. It set `len = 0` up front and cleared slots lazily, so a drain dropped half-way left the set reporting empty while `contains()` still matched every unvisited key. It hands its storage to the iterator now, the way the map's drain always did. Separate red-first test.

Gates: both suites 4967/4969, oracle 9/9, fmt + clippy clean.

PR #59, opened 2026-09-01
fix(aggregation): honour ORDER BY on a grouped column the SELECT drops

semihalev:
Found while retiring the sentinel (#58): an ORDER BY on a grouped column is silently dropped when that column is not in the SELECT list.

```sql
CREATE TABLE t (id INTEGER PRIMARY KEY, v INTEGER);
INSERT INTO t VALUES (1,7), (2,1), (3,7), (4,5);
SELECT SUM(id) FROM t GROUP BY v ORDER BY v;       -- returned [4,4,2], correct is [2,4,4]
SELECT SUM(id) FROM t GROUP BY v ORDER BY v DESC;  -- returned [4,4,2] as well
```

Projection removed `v` before the sort ran, so every sort key evaluated to NULL and the rows kept the order the grouping happened to produce. ASC and DESC returned the same thing, which is the tell.

## Fix

The aggregate path captures the grouped columns that ORDER BY names before projection consumes the rows, and appends them to the result. That is the same shape it already uses for aggregates appearing only in ORDER BY (the `hidden` aggregates). `execute_select` then trims the extra trailing columns after sorting through the existing `expected_columns` contract, which is how the non-aggregate path has always handled ORDER BY columns outside the SELECT list.

## Scope, honestly

- **Ordering by an aggregate was never broken.** I first thought `ORDER BY SUM(id) DESC` was affected too; it is not. My expectation was simply wrong: two groups tied on the sort key and I read the tie as a bug.
- The fix only carries plain column references (`ORDER BY v`, `ORDER BY t.v`). Expressions over grouped columns (`ORDER BY v + 0`) still fall to the generic path and remain unsorted; they need the same treatment for computed keys, which is a larger change and belongs in its own PR.

## Verification

New `group_by_order_test`: ascending and descending on an unprojected integer group column, the same on a TEXT column, and the projected form as a control. Both suites 4969/4971, differential oracle 9/9, CI-mode 3/3, fmt + clippy clean.


semihalev:
Both findings fixed in $(git rev-parse --short HEAD), plus the artifact you spotted:

- **P1 (qualified key)**: the collector reduced `ORDER BY t.v` to the bare name, so it never matched the aggregate rows, which hold the column as `t.v`. Both spellings are tried now, qualified first (the order `execute_select` itself resolves them in), and the column is published under the name the sort will look for. Tests cover `GROUP BY t.v ORDER BY t.v` and the mixed `GROUP BY v ORDER BY t.v DESC`.
- **P2 (aggregate + window)**: right, that route goes through `execute_aggregation_for_window` and then the window projection, which dropped the grouped column again. The window projection carries the same hidden sort keys now. Red-first test: `SELECT MIN(id), ROW_NUMBER() OVER () FROM t GROUP BY v ORDER BY v` returned [4,1,2] before, [2,4,1] after. The name-resolution rule is shared between the two paths rather than duplicated by hand.
- **`examples/gb_probe.rs`**: not deliberate, it was my diagnostic scratch file and `git add -A` picked it up. Removed. It is also what failed the License Check on the previous commit, since it had no license header.

Gates: both suites 4971/4973, CI-mode 5/5, oracle 9/9, fmt + clippy clean.

semihalev:
Both edge cases fixed in d211bc41:

- **P1 (bare over qualified)**: resolution is symmetric now — full name, then bare name, then a unique `<qualifier>.<name>` suffix, declining when several qualifiers match so an ambiguous `v` never silently picks one. While fixing it I collapsed the aggregate and window copies into **one shared resolver**, which is the durable answer to "the same logic must be fixed in window.rs too": there is no second copy left to drift.
- **P2 (alias shadow)**: an alias is never qualified, so the alias-name fallback now applies only when no column carries the qualified name. `SELECT MAX(id) AS v FROM t GROUP BY t.v ORDER BY t.v` sorts by the group key again.

Tests: reversed spelling, alias shadow, and the reversed spelling through the aggregate+window route.

Gates: both suites 4974/4976, CI-mode 8/8, oracle 9/9, fmt + clippy clean.

PR #60, opened 2026-09-01
fix(common): identify columns by content, guard retain and drain

semihalev:
First wave of the `src/common` audit: the three correctness findings. No performance work here, that comes in the following PRs.

## 1. `CompiledEvaluator` identified columns by a raw address (silent wrong results)

`columns_arc_id` was a single `usize` holding two different kinds of address:

- `init_columns(&[String])` stored `columns.as_ptr()`, the address of a **borrowed slice the evaluator does not retain** (it keeps a `to_vec()` copy). Once the caller's buffer is dropped, that address can be handed back by the allocator for something else.
- `init_columns_arc(CompactArc<Vec<String>>)` compared its `CompactArc` payload pointer against the same field, with no length guard.

A match on a recycled address returns early and leaves the **previous schema** installed. Column indices then resolve against the wrong table: wrong values, no error.

Fixed by comparing what actually identifies the columns:

- slice path: compare contents (`self.columns.as_slice() == columns`)
- Arc path: `CompactArc::ptr_eq`, where both handles are alive so pointer equality is exact identity and cannot ABA

The address cache is deleted. It was not paying for itself anyway: every production caller either creates a fresh evaluator (so the field is 0 and the fast path never hits) or calls `init_columns` once outside the row loop. On the hit path the content compare replaces N `String` allocations, so it can never be worse than the miss path it guards.

Red-first test (`test_compiled_evaluator_init_columns_reused_buffer`) reuses one buffer with different names, which keeps the address and the length identical. Before the fix:

```
assertion `left == right` failed
  left: ["a", "b"]
 right: ["z", "b"]
```

## 2. `CompactVec::retain` had no panic guard (double free)

The loop shifts kept elements down with `copy_nonoverlapping` and calls `set_len` only after it finishes. A predicate that panics leaves the length at its original value while the retained prefix has already been duplicated by the backshift, so those elements are dropped twice, and any element the `else` arm already dropped is dropped again.

`retain` now runs under an RAII guard that moves the untouched tail into place and restores an accurate length on every exit path, matching the shape of the existing guards in `Clone`, `extend` and `from_iter`.

Red-first test: 5 elements, predicate rejects one and panics on the next. Before the fix, 5 elements produced 6 drops.

Only `CompactVec<i64>` callers exist today (`storage/index/{hash,multi_column}.rs`), and `i64` has no drop glue, so this was not reachable in production. It is public API.

## 3. `CompactVec::drain` left the length untouched (double free under `mem::forget`)

`drain()` built the `Drain` without lowering `len`; only `Drain::drop` set it to `start`. `mem::forget` is safe code, so a partially consumed `Drain` that is forgotten leaves the vector owning elements that were already moved out.

`drain()` now shortens the vector before yielding anything, which is also what `std::vec::Drain` does. `Drain::start` became unnecessary and is removed.

Red-first test: drain from index 2 of a 4-element vector, consume one, `mem::forget` the drain. Before the fix `vec.len()` was still 4.

`drain` has no production callers today. Also public API.

## Not in scope

`I64Map::drain` and `I64Set::drain` are correct on `main` already (fixed in #58); the audit initially flagged them because the agent read a stale branch. The capacity they throw away is a real perf finding and is queued for a later PR.

## Gates

- `cargo nextest run`: 4978 passed
- `cargo nextest run --features test-filedb`: 4980 passed
- `cargo +nightly miri nextest run --lib -E 'test(/^common::compact_vec::/)'`: 17 passed, which covers the new `retain` guard under Stacked Borrows

### DIFF
--- a/src/common/compact_vec.rs
+++ b/src/common/compact_vec.rs
@@ -417,13 +417,18 @@ impl<T> CompactVec<T> {
                     }
                 }
                 guard.write_idx += 1;
+                guard.read_idx += 1;
             } else {
+                // Advance past the slot before dropping it: if `T::drop`
+                // panics, the guard must not count this element as untouched
+                // tail and hand the dropped value back to the vector.
+                let idx = guard.read_idx;
+                guard.read_idx += 1;
                 // SAFETY: the element is initialized and is not moved anywhere.
                 unsafe {
-                    ptr::drop_in_place(ptr.add(guard.read_idx));
+                    ptr::drop_in_place(ptr.add(idx));
                 }
             }
-            guard.read_idx += 1;
         }
     }
 
@@ -1053,13 +1058,19 @@ impl<'a, T> ExactSizeIterator for Drain<'a, T> {
 
 impl<'a, T> Drop for Drain<'a, T> {
     fn drop(&mut self) {
-        // Drop any remaining elements not yet consumed
-        while self.current < self.end {
-            // SAFETY: Elements [current..end] are initialized but not yet consumed.
+        let remaining = self.end - self.current;
+        if remaining > 0 {
+            let first = self.current;
+            self.current = self.end;
+            // Drop the rest as a slice: slice drop glue keeps dropping the
+            // remaining elements when one destructor panics, where an
+            // element-by-element loop would strand them. The vector no longer
+            // owns this range, so nothing else would ever free them.
+            // SAFETY: [first..end] are initialized and owned by this drain.
             unsafe {
-                ptr::drop_in_place(self.vec.ptr.as_ptr().add(self.current));
+                let start = self.vec.ptr.as_ptr().add(first);
+                ptr::drop_in_place(ptr::slice_from_raw_parts_mut(start, remaining));
             }
-            self.current += 1;
         }
         // The length was already lowered to `start` by `drain()`, so there is
         // nothing to restore here.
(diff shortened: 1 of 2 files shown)


semihalev:
Both findings were right, fixed in 2147d601.

**[P1] `retain` re-owned a dropped slot.** `read_idx` was advanced after `drop_in_place`, so a panicking `T::drop` left it pointing at the slot it had just dropped; the guard counted that slot as untouched tail, copied it down and handed the dropped value back to the vector. The index is now advanced before the drop. This is the same ordering `std::vec::Vec::retain` uses, with the same reason in its comment ("Advance early to avoid double drop if `drop_in_place` panicked").

**[P2] `Drain::drop` stranded the rest after a panic.** The element-by-element loop stopped at the first panicking destructor, and since `drain()` had already lowered the length to `start`, nothing else would ever free the remainder. The range is now dropped through slice drop glue, which keeps dropping the remaining elements when one destructor panics. Confirmed empirically by the new test rather than assumed.

Both have regression tests that count drops with a destructor that panics exactly once, so a double drop shows up in the counter instead of aborting the process:

| test | before | after | expected |
|---|---|---|---|
| `test_retain_destructor_panic_does_not_double_drop` | 6 drops | 5 | 5 elements, one drop each |
| `test_drain_drop_finishes_after_destructor_panic` | 3 drops | 5 | nothing stranded |

Gates re-run on the new head:

- `cargo nextest run`: 4980 passed
- `cargo nextest run --features test-filedb`: 4982 passed
- `cargo +nightly miri nextest run --lib -E 'test(/^common::compact_vec::/)'`: 19 passed

PR #61, opened 2026-09-01
perf(storage): mix the float bit pattern before it becomes a group key

semihalev:
Rewritten after review. The first version changed `I64Map`'s hash and that was the wrong place; this one leaves the hash untouched and fixes the key.

## Why the hash was the wrong place to change

The bucket is the low bits of `hash(key)`, and in `Z/2^64` a multiply is lower triangular: `(a*b) mod 2^m` depends only on `a mod 2^m`. Two consequences follow, and they pull against each other:

- Multiplying by an odd constant is a **permutation of `Z/2^m`**, so a dense key range never collides, at any capacity and any load factor. That is not luck, and it is what row ids, transaction ids and every other counter in the engine rely on.
- Entropy above the mask is **invisible**. `f64::to_bits()` of a round value is exactly that shape, so all such keys share one bucket.

Any mixer that pulls the high bits down destroys the first property. Measured on dense keys with realistic slot sizes:

| dense-key lookup | current hash | strong (SplitMix) hash |
|---|---|---|
| 10k keys, 75% load | 0.522 ns | 1.437 ns |
| 1M keys, 75% load | **3.003 ns** | **8.662 ns** |

A perfect hash touches one cache line per lookup; a strong one touches 2.4. That is a 2.9x regression on the version store's row id map, which is not a trade worth making for a grouping key.

I also measured a probe matrix over stride families, capacities and loads before concluding this. Worst average probes over each family (deterministic, no timing involved):

| cap / load / family | current hash | rotate mixer (v1 of this PR) | split mixer | SplitMix |
|---|---|---|---|---|
| 4096, 75%, odd strides | 51.49 | 11.92 | 1.00 | 2.77 |
| 4096, 75%, even strides | 8.50 | 12.89 | 24.50 | 2.77 |
| 4096, 75%, powers of two | 13.81 | 1.71 | 1536.50 | 2.78 |
| 65536, 75%, powers of two | 3.50 | 1.67 | 24576.50 | 2.55 |

Every cheap fixed mixer has a family it loses to. The reviewer's `i*1483` case is in there: the rotate mixer this PR originally shipped gives 1069.64 average and 3071 maximum probes on it, and the dense regression they found reproduces exactly at 1.393. Both are gone because the hash is no longer touched.

## What actually changed

`version_store.rs` produced the bad key itself, and it is the one place that knows the value is a float:

```rust
- Some(f.to_bits() as i64)
+ Some(key_from_f64(*f))
...
- 1 => Value::Float(f64::from_bits(key as u64)),
+ 1 => Value::Float(f64_from_key(key)),
```

`key_from_f64` runs the bit pattern through the SplitMix64 finalizer; `f64_from_key` undoes it. Every step is invertible (the xor-shifts are undone by applying the shift at `s`, `2s`, ..., the multiplies by their modular inverses), so the encoding is a bijection: distinct doubles keep distinct keys and grouping stays exact. `-0.0` and NaN payloads stay exactly as distinguishable as the raw bit pattern left them, so no grouping semantics move.

The helpers live next to `hash` in `i64_map.rs`, because they exist to satisfy its contract, and that contract is now written down there.

## Results

Interleaved best of 12 runs, alternating binaries, since criterion disagreed with itself across runs on this machine (the same unmodified code measured `p = 0.12, no change` and `+9.2%, p = 0.00` for the same case):

| case | before | after | |
|---|---|---|---|
| whole doubles, 40k groups | 587 ms | **4.6 ms** | 128x |
| money 0.01 steps, 40k groups | 74.7 ms | **4.8 ms** | 15x |
| irregular doubles, 40k groups | 4.8 ms | 4.7 ms | parity |
| integer keys, 200k groups | 26.8 ms | 26.0 ms | parity |

Irregular doubles come out at parity rather than slower: the mix costs about a nanosecond per row and buys back the probes it removes.

## Tests

- `test_float_key_round_trips` asserts bit-exact recovery over edge values (zeros, infinities, subnormals, a NaN payload, MIN/MAX/EPSILON) plus 26k generated bit patterns.
- `test_float_key_is_injective` asserts 50k distinct doubles keep 50k distinct keys, so no two groups can merge.
- `test_float_keys_spread_across_buckets` sweeps three capacities at their 3/4 grow threshold across five float families. Against the old raw `to_bits()` key it fails with `768 whole float keys landed in 1 of 1024 buckets`.
- `test_hash_keeps_dense_keys_collision_free` sweeps the same capacities and asserts dense ranges stay collision free. It passes today; it is there because the rejected v1 of this PR broke exactly that, and the test I had then only checked one lucky capacity.

## Known and unchanged

An integer column whose values share low zero bits still clusters: 42 probes for the worst power-of-two stride measured, against 24576 for the float pattern fixed here. The benchmark keeps that case visible. Fixing it would mean mixing integer group keys too, which costs `GROUP BY <integer column>` its collision-free property, so it is deliberately left alone.

## Gates

- `cargo nextest run`: 4984 passed
- `cargo nextest run --features test-filedb`: 4986 passed
- `cargo +nightly miri nextest run --lib -E 'test(/^common::i64_map::/)'`: 33 passed

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -198,6 +198,10 @@ harness = false
 name = "delete_complex"
 harness = false
 
+[[bench]]
+name = "group_by_keys"
+harness = false
+
 [profile.bench]
 lto = "thin"
 codegen-units = 1
--- a/benches/group_by_keys.rs
+++ b/benches/group_by_keys.rs
@@ -1,0 +1,103 @@
+// Copyright 2025 Stoolap Contributors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! GROUP BY throughput across grouping-key distributions
+//!
+//! Run with: cargo bench --bench group_by_keys
+//!
+//! The grouping fast path keys its hash map by the raw integer, or by
+//! `f64::to_bits()` for a float column. Some perfectly ordinary column values
+//! are adversarial for a hash that only mixes part of the key: a DOUBLE column
+//! holding round values has all-zero low mantissa bits, and an integer column
+//! holding a large stride has all-zero low bits. This benchmark keeps those
+//! shapes next to the benign ones so a hash change that reintroduces the
+//! clustering shows up as a large, obvious regression rather than as a slow
+//! query somebody reports later.
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use std::hint::black_box;
+use stoolap::Database;
+
+/// One row per group, so the grouping map holds exactly this many keys.
+///
+/// The value is chosen so the map sits at its 3/4 grow threshold rather than
+/// wherever a round number happens to land: capacity is
+/// `next_power_of_two(n * 4 / 3)`, so 12288 keys fill 16384 slots exactly.
+/// Probe length is worst there, which is where a weak hash shows itself.
+const ROW_COUNT: i64 = 12_288;
+fn setup(name: &str, key: impl Fn(i64) -> f64) -> Database {
+    let db = Database::open(&format!("memory://bench_group_by_{}", name)).unwrap();
+    db.execute(
+        "CREATE TABLE t (id INTEGER PRIMARY KEY, k FLOAT, v INTEGER)",
+        (),
+    )
+    .unwrap();
+    let insert = db
+        .prepare("INSERT INTO t (id, k, v) VALUES ($1, $2, $3)")
+        .unwrap();
+    for i in 1..=ROW_COUNT {
+        insert.execute((i, key(i), i)).unwrap();
+    }
+    db
+}
+
+fn setup_int(name: &str, key: impl Fn(i64) -> i64) -> Database {
+    let db = Database::open(&format!("memory://bench_group_by_{}", name)).unwrap();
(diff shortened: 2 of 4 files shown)


semihalev:
Both fixed in 2f7e8865, and the counterexample reproduces exactly.

**The dense-key claim was too strong.** The pre-mix is `k ^ (k >> 16)`, so the guarantee only holds while `k >> 16` is constant. Inside one 2^16 block the mix is `k ^ const`, and both that and the odd multiply permute the low bits, which is provable. Across a boundary the two halves get different constants and can overlap. My three ranges all happened to sit inside a block, so the test asserted a universal property it never exercised.

Reproduced at `33_554_048..33_554_816`, capacity 1024: **384 buckets, 2.023 average probes**, exactly as reported.

I then swept a 3/4-full window across every alignment of five different boundaries at three capacities, to find how bad it actually gets rather than guessing a bound:

| capacity | keys | worst buckets | worst avg probe | worst max probe |
|---|---|---|---|---|
| 1024 | 768 | 384 (50%) | 2.023 | 6 |
| 4096 | 3072 | 3072 (100%) | 1.000 | 1 |
| 65536 | 49152 | 49152 (100%) | 1.000 | 1 |

Only the 10-bit mask degrades, and the worst case is exactly your example. So:

- `test_hash_keeps_dense_keys_collision_free` now uses block-aligned ranges and its doc comment carries the proof and its limit.
- `test_hash_bounds_dense_keys_across_a_block_boundary` pins the boundary case at your counterexample with a baseline bound rather than leaving it undescribed, so a later change to the hash cannot quietly make it worse.

The narrowed test still catches what it was written for: the rejected rotate mixer puts `0..768` into 710 of 1024 buckets, so it fails on the first shape.

**Benchmark header** no longer describes the float key as `f64::to_bits()`; it now says it used to be keyed that way and explains why the encoding exists, and notes the integer stride case is still keyed raw.

Gates re-run: fmt, clippy `-D warnings`, `cargo nextest run` 4985 passed, `--features test-filedb` 4987 passed, differential oracle 9 passed.

PR #62, opened 2026-09-02
perf(aggregation): one allocation per group in the storage GROUP BY

semihalev:
Step 3 of the audit, narrowed after review: the allocation chain in the single-column storage GROUP BY. No answer changes, only allocations.

## 1. One accumulator pool instead of one vector per group

```rust
let accums = int_groups.entry(key).or_insert_with(|| vec![Accum::default(); aggregates.len()]);
```

A 40k-group scan paid 40k allocations before doing any work, and consecutive groups' accumulators landed wherever the allocator put them. `Accum` is 64 bytes and the count is fixed for the whole query, so they now live in one flat pool with the maps holding a `u32` ordinal into it: a single growing allocation for the scan, and each group's accumulators contiguous.

The ordinal is counted rather than derived from the pool length, so the aggregate count never appears in a divisor. `compute_grouped_aggregates` is public and its signature allows an empty aggregate list, which the executor never sends but the contract permits, and dividing by it would panic.

## 2. The result row was built as two vectors and then concatenated

Each group produced a one-element `group_values` and a separate `aggregate_values`, and the only consumer does `values.extend(r.aggregate_values)`, reallocating the first every time. The key vector is now sized for both halves at the producer, so the append lands in place. Merging the two fields into one vector would remove the second allocation as well, but `GroupedAggregateResult` is public API, so that belongs in 0.5.0.

## 3. Finalization consumes `other_groups`

It iterated by reference, so every TEXT key stayed alive next to the clone made from it for the rest of the loop. It consumes the map now.

## Numbers

Interleaved best of 10, two binaries alternated, 40k groups with 4 rows each:

| query | before | after | |
|---|---|---|---|
| `SELECT k, SUM(v) FROM t GROUP BY k` | 3.44 ms | 2.18 ms | 1.58x |
| `SELECT k, SUM(v), COUNT(*), MIN(v) FROM t GROUP BY k` | 4.40 ms | 3.46 ms | 1.27x |

Medians move the same way (1.53x, 1.21x).

## Tests

- `test_grouped_aggregates_with_no_aggregates` calls the public API with an empty aggregate list through a read-committed visibility checker, so the arena path is actually reached. It fails with `attempt to divide by zero` against the first version of this branch.
- `tests/group_by_expression_aggregate_test.rs` pins the results of GROUP BY with an expression argument, a FILTER, a HAVING over an expression aggregate, a TEXT key and a NULL group, and records what each shape answers for `0.0` versus `-0.0`.

## Dropped from this PR after review

Widening the executor's raw-entry gate is a **semantic** change, not a reordering: the grouping paths disagree about float keys, one comparing bit patterns and the other `Value` equality, so moving a query between them moves its group membership.

```
                          origin/main   with the gate widened
SELECT v, SUM(w)     ...     3 groups      3 groups
SELECT v, COUNT(*)   ...     3 groups      3 groups
SELECT v, SUM(w * 1) ...     3 groups      2 groups
```

Merging `0.0` with `-0.0` is probably the semantic we want, but it has to be decided for every grouping path at once with the membership visible in the diff. The signed-zero test records today's answers so that change cannot arrive as a side effect. It is queued with the TEXT-key and volume-path work, which touch the same code.

## Also still queued

- The TEXT key path allocates a `CompactArc<Value>` per row to probe `other_groups`, because `AHashMap` has no raw-entry API. Moving that map to hashbrown is where the hasher-mismatch bug in #52 came from, so it gets its own PR.
- The volume (sealed segment) grouping builds `(vec![key], vec![Accum; n])` per group in four places and clones the key to probe.
- Merging `GroupedAggregateResult`'s two fields: 0.5.0, public API.

## Gates

- `cargo nextest run`: 4992 passed
- `cargo nextest run --features test-filedb`: 4994 passed
- `cargo test --test group_by_expression_aggregate_test` (CI mode): 6 passed

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -3747,7 +3747,14 @@ impl Executor {
         let has_limit = limit.is_some();
         let mut current_group_count: usize = 0; // Track actual group count for LIMIT optimization
 
-        if all_simple_columns && expr_vm.is_none() {
+        // `all_simple_columns` already means no GROUP BY item is an expression,
+        // so the VM is not needed to build the keys. It was also gated on
+        // `expr_vm.is_none()`, which is false whenever any aggregate carries an
+        // expression, an ORDER BY or a FILTER, and that sent queries like
+        // `SELECT k, SUM(x * 1.1) ... GROUP BY k` down the slow path even
+        // though their keys are plain columns. Aggregates are evaluated later
+        // from the groups, not here.
+        if all_simple_columns {
             // Fast path: extract key values directly from row columns
             let column_indices: Vec<usize> = precomputed_group_by
                 .iter()
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -3747,14 +3747,7 @@ impl Executor {
         let has_limit = limit.is_some();
         let mut current_group_count: usize = 0; // Track actual group count for LIMIT optimization
 
-        // `all_simple_columns` already means no GROUP BY item is an expression,
-        // so the VM is not needed to build the keys. It was also gated on
-        // `expr_vm.is_none()`, which is false whenever any aggregate carries an
-        // expression, an ORDER BY or a FILTER, and that sent queries like
-        // `SELECT k, SUM(x * 1.1) ... GROUP BY k` down the slow path even
-        // though their keys are plain columns. Aggregates are evaluated later
-        // from the groups, not here.
-        if all_simple_columns {
+        if all_simple_columns && expr_vm.is_none() {
             // Fast path: extract key values directly from row columns
             let column_indices: Vec<usize> = precomputed_group_by
                 .iter()
--- a/src/storage/mvcc/version_store.rs
+++ b/src/storage/mvcc/version_store.rs
@@ -6067,11 +6067,23 @@ impl VersionStore {
 
             // Track key type: 0=Integer, 1=Float, 2=Boolean, 3=Mixed/Other
             let mut key_type: u8 = 255; // uninitialized
-            let mut int_groups: I64Map<Vec<Accum>> = I64Map::new();
+                                        // Accumulators live in one flat pool, `n_aggs` per group, and the
+                                        // maps hold the group's ordinal into it. A Vec per group would be
+                                        // one allocation per distinct group, which for a high-cardinality
+                                        // GROUP BY is the dominant cost of the scan.
+            let n_aggs = aggregates.len();
+            let mut accums: Vec<Accum> = Vec::new();
+            let new_group = |accums: &mut Vec<Accum>| -> u32 {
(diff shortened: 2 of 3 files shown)


semihalev:
All three confirmed and fixed in d8485fc8.

## [P2] Group membership, not row order

You were right, and it is worse than the note I wrote in the PR body. Reproduced directly, three rows: `0.0`, `-0.0`, `1.0`.

```
                          origin/main   with the gate widened
SELECT v, SUM(w)     ...     3 groups      3 groups
SELECT v, COUNT(*)   ...     3 groups      3 groups
SELECT v, SUM(w * 1) ...     3 groups      2 groups
```

The path a query lands on decides whether float keys are compared by bit pattern or by `Value` equality, so moving a query between paths moves its group membership. **The gate change is out of this PR.**

It is still worth doing, and merging `0.0` with `-0.0` is probably the semantic we want, but as you say it has to be decided for every grouping path at once. `test_signed_zero_group_membership_is_stable_per_shape` now records what each shape answers today, so that change arrives with the numbers updated in the same commit rather than as a side effect. Queued behind the TEXT-key and volume-path work, which touch the same code.

What remains here is only the allocation work, which does not move any answer.

## [P2] Divide by zero on an empty aggregate list

Confirmed: `compute_grouped_aggregates` is public, its signature allows `aggregates: &[]`, and the previous code answered that with a group per distinct key and no aggregate values. The executor guards it (`if aggregates.is_empty() { return None }`), so this is API-only, but the contract does not.

Fixed by removing the division rather than special-casing it: the ordinal is counted, so the aggregate count never appears in a divisor, and `n_aggs == 0` falls out as empty slices.

`test_grouped_aggregates_with_no_aggregates` calls the API directly through a read-committed visibility checker so the arena path is actually reachable. It panics with `attempt to divide by zero` against the previous commit.

## Owned iteration for `other_groups`

Confirmed and applied. It now consumes the map, so each `GroupKey` is released as its value is moved out instead of being held next to its clone for the rest of the loop.

## Numbers after the change

Same method, interleaved best of 10 at 40k groups:

| query | before | after | |
|---|---|---|---|
| `SELECT k, SUM(v) FROM t GROUP BY k` | 3.44 ms | 2.18 ms | 1.58x |
| `SELECT k, SUM(v), COUNT(*), MIN(v) FROM t GROUP BY k` | 4.40 ms | 3.46 ms | 1.27x |
| `SELECT k, SUM(v * 1.1) FROM t GROUP BY k` | 12.44 ms | 12.12 ms | parity |

The third is parity now, as it should be with the gate reverted.

The behaviour note about group order is withdrawn from the PR body along with the gate.

## Gates

- `cargo nextest run`: 4992 passed
- `cargo nextest run --features test-filedb`: 4994 passed
- `cargo test --test group_by_expression_aggregate_test` (CI mode): 6 passed

PR #63, opened 2026-09-02
perf(index): drop the hash index's write-only hash_to_rows map

semihalev:
64 lines deleted, nothing else changed.

## What it was

`HashIndex` kept two maps of the same information:

- `hash_to_values: hash -> [(values, row_ids)]`, read by `find`, `get_row_ids_equal_into` and the unique-constraint check
- `hash_to_rows: hash -> row_ids`, **read by nothing**

Every `add_entry`, `remove_entry`, `add_batch_slice` and `remove_batch_slice` maintained the second one. The comment above the read path said "First check hash_to_rows for quick path (no collisions)", but the line under it has always read `hash_to_values`. `rg 'hash_to_rows.read'` finds nothing: every single access is a write lock.

## What it cost

Per indexed write:

- a second `RwLock` write acquisition
- a hash lookup
- a `binary_search` plus a sorted `insert`, or a `retain` over a `CompactVec<i64>`
- one 32-byte heap buffer per distinct indexed value

An update pays it twice, once to pull the row out from under its old hash and once to file it under the new one, which is why updates and deletes gain the most.

This is not an obscure path. `auto_select_index_type` returns `IndexType::Hash` for a single-column TEXT or JSON index, so it covers most string indexes.

## Numbers

Interleaved best of 8, two binaries alternated, 100k rows with an index on a TEXT column:

| operation | before | after | |
|---|---|---|---|
| INSERT 100k | 83.9 ms | 77.3 ms | 1.09x |
| UPDATE a quarter of the rows | 78.0 ms | 51.7 ms | **1.51x** |
| DELETE a fifth of the rows | 33.5 ms | 23.5 ms | **1.43x** |
| 10k point lookups | 2215 ms | 2189 ms | unchanged |

Medians agree (1.02x, 1.51x, 1.41x, 1.03x). Lookups are unchanged because they never touched the map, which is the point.

Memory goes too: a 24-byte table entry plus a 32-byte heap buffer per distinct indexed value, so roughly 56 bytes and one allocation per distinct value, on top of the allocator's own overhead.

## Tests

No new test. The change removes a structure nothing reads, and the existing tests are exactly the right guard: `test_hash_index_basic`, `test_hash_index_remove`, `test_hash_index_update`, `test_hash_index_duplicate_values`, `test_hash_index_unique_constraint`, `test_hash_index_null_values`, `test_hash_index_multi_column`, plus `tests/index_operations_test.rs` at the SQL level. If the deletion had taken a live line with it, they would fail.

## Gates

- `cargo nextest run`: 4992 passed
- `cargo nextest run --features test-filedb`: 4994 passed

### DIFF
--- a/src/storage/index/hash.rs
+++ b/src/storage/index/hash.rs
@@ -39,8 +39,8 @@
 //!
 //! ## Known Limitations:
 //!
-//! ### Triple Lock Pattern for Write Operations
-//! The `add` and `remove` methods acquire three RwLock write locks simultaneously:
+//! ### Paired Lock Pattern for Write Operations
+//! The `add` and `remove` methods acquire both RwLock write locks simultaneously:
 //! - `row_to_hash`: Maps row_id -> hash for efficient removal
 //! - `hash_to_values`: Maps hash -> values for collision handling
 //!
@@ -53,7 +53,7 @@
 //!
 //! **Alternative considered**: Using a single RwLock<struct> would have similar
 //! contention characteristics. A concurrent map could help for truly concurrent writes
-//! but would complicate atomicity across the three maps.
+//! but would complicate atomicity across the two maps.
 
 use parking_lot::RwLock;
 use std::hash::{BuildHasher, Hash, Hasher};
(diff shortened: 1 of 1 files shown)


semihalev:
Fixed. The section header, the lock count and the closing sentence all said three; with `hash_to_rows` gone the write paths take `row_to_hash` and `hash_to_values`, so it now reads "Paired Lock Pattern", "both RwLock write locks" and "across the two maps".

Comment-only change: `cargo fmt` and a build, no test run, as you said.

PR #67, opened 2026-09-02
perf(parser): slice tokens from the input, stop copying text through String

semihalev:
Second and last part of the string/buffer wave: text that was built one allocation at a time.

## 1. The lexer built tokens char by char

`read_identifier`, `read_number`, `read_line_comment`, `read_block_comment`, `read_parameter`, `read_named_parameter` and `read_quoted_identifier` all did:

```rust
let mut result = SmartString::new("");
result.push(self.ch); self.read_char();
while ... { result.push(self.ch); self.read_char(); }
```

`SmartString::push` past the 15-byte inline buffer is `as_str().to_string()`, push, `from_string` (a fresh Arc), drop the old one. So a token of n bytes cost about 2n allocations and n full copies of itself. A 400-character comment, the shape an ORM attaches to every statement, parsed slower than ten short queries.

Every reader advances only through `read_char`, so `position` is always the first byte of the character that ended the token. The readers now record the start and slice `input[start..position]` once, exactly as `read_string_literal` already did. A quoted identifier is sliced too, unless a doubled quote turns up inside it, in which case the prefix is copied once and the rest appended to that copy.

## 2. `Token::new` copied the token again

It took `impl AsRef<str>` and did `SmartString::from(literal.as_ref())`, so the SmartString the lexer had just built was copied into a second one: for a long token, another heap string and Arc. It takes `impl Into<SmartString>` now and the lexer's value moves in. `&str` callers are unaffected.

## 3. `Value::text` went through a temporary `String`

```rust
- pub fn text(value: impl Into<String>) -> Self { Value::Text(SmartString::from_string_shared(value.into())) }
+ pub fn text(value: impl Into<SmartString>) -> Self { Value::Text(value.into()) }
```

For a `&str` that was `String::from` (malloc), copy into the inline buffer, free, for every text value up to 15 bytes. The representation is unchanged: inline up to 15 bytes, the shared heap form above. Seven of the 598 callers passed `&String`, which also cost a `String` clone; they pass `.as_str()` now. The hot callers are the C ABI's parameter binding (`ffi/value.rs`), so every driver pays this per TEXT parameter.

One correction to the audit: it listed `volume/format.rs:439` as a per-row reader. `read_value` there is only called for zone map bounds when a volume is opened, not per row. The per-row TEXT decode on scans does not go through `Value::text`.

## 4. `to_lowercase` allocated for every identifier

`SmartString::to_lowercase` was `from_string(self.as_str().to_lowercase())`: a String allocated, copied back inline, freed. `Identifier::new` calls it for every identifier of every parsed statement. An inline ASCII string is now folded in place in the 15-byte buffer. Unicode lowercasing of pure ASCII is the ASCII fold byte for byte, so nothing changes; non-ASCII input still takes the general path, and a test covers `Ünal Çağ` and `İSTANBUL` to keep it that way.

## Numbers

Interleaved best of 8, two binaries alternated, each query unique so the plan cache cannot help:

| | before | after | |
|---|---|---|---|
| parse, 400-char comment | 28.58 us | **4.22 us** | 6.8x |
| parse, long identifiers | 6.99 us | **3.23 us** | 2.2x |
| parse, short query | 2.67 us | 2.62 us | parity |
| `Value::text`, 2 to 14 bytes | 10-14 ns | **2-3 ns** | 4-5x |
| `Value::text`, 34 bytes | 24 ns | 23 ns | parity |

Medians and bests agree throughout.

## Tests

Six new lexer tests pin exact literals for multi-byte identifiers (`çağ_ü`, `日本語`, so the byte bounds land on char boundaries), a 300-character comment and a long identifier, `#` comments, `--` at end of input, escaped quoted identifiers on both the sliced and the copied path, parameters, and every reader's error path. Two new SmartString tests cover the ASCII fast path staying inline and non-ASCII input still folding correctly.

## Gates

- `cargo nextest run`: 5000 passed
- `cargo nextest run --features test-filedb`: 5002 passed

### DIFF
--- a/src/common/smart_string.rs
+++ b/src/common/smart_string.rs
@@ -335,13 +335,33 @@ impl SmartString {
         *self = Self::from_string(s);
     }
 
+    /// Lowercase copy. An inline ASCII string, which is what nearly every
+    /// identifier is, is folded in the 15-byte buffer without touching the
+    /// heap; Unicode lowercasing of pure ASCII is byte for byte the ASCII
+    /// fold, so the result is the same as the general path.
     #[inline]
     pub fn to_lowercase(&self) -> SmartString {
+        if let Some(len) = self.tag.inline_len() {
+            if self.data[..len].is_ascii() {
+                let mut copy = self.clone();
+                copy.data[..len].make_ascii_lowercase();
+                return copy;
+            }
+        }
         SmartString::from_string(self.as_str().to_lowercase())
     }
 
+    /// Uppercase copy, with the same inline ASCII fast path as
+    /// [`SmartString::to_lowercase`].
     #[inline]
     pub fn to_uppercase(&self) -> SmartString {
+        if let Some(len) = self.tag.inline_len() {
+            if self.data[..len].is_ascii() {
+                let mut copy = self.clone();
+                copy.data[..len].make_ascii_uppercase();
+                return copy;
+            }
+        }
         SmartString::from_string(self.as_str().to_uppercase())
     }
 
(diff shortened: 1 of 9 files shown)


semihalev:
Right, that was a source-compatibility break, not a runtime one. Fixed in the follow-up commit.

`SmartString` gains two impls:

```rust
impl From<&String> for SmartString      { fn from(s: &String) -> Self { SmartString::new(s.as_str()) } }
impl From<&SmartString> for SmartString { fn from(s: &SmartString) -> Self { s.clone() } }
```

The first reads the borrowed `String` straight into the buffer, so a caller that passed `&String` to the old `Into<String>` parameter keeps compiling and no longer pays for the temporary `String` on the way. The second is a clone, which on the heap representation is the O(1) Arc bump, not a text copy.

The seven `.as_str()` edits are reverted: those call sites are back to the exact text that compiled against the old signature, and they compile against the new one, which makes them the compatibility regression check you described. A unit test covers both impls (inline stays inline, a long borrowed `String` round-trips, `&SmartString` clones).

Gates on the new head: fmt, clippy `-D warnings`, `cargo nextest run` 5001 passed, `--features test-filedb` 5003 passed, differential oracle 9 passed.

PR #75, opened 2026-09-02
perf(storage): snapshot cold segments once per batch fetch, not per row

semihalev:
## Why

Wave H of the executor audit, storage fetch part. On a table with sealed cold segments, `collect_rows_by_ids` and `fetch_rows_by_ids_into` called `find_segment_row` per probed id. Each call took four lock reads (hot row set, tombstones, pending tombstones, manifest), cloned two Arcs and collected the manifest's segment ids into a fresh `Vec`. A `sample` profile of an index nested loop join over sealed rows put `find_segment_row` at 44% of the fetch loop.

## What

- Both batch paths take one `statement_snapshot()` and search with `find_segment_row_in`, the helper the UPDATE and DELETE paths already use.
- `find_segment_row_in` returns a reference into the snapshot instead of cloning the `ColdSegment` (its `ColumnMapping` holds a `Vec`). The first attempt, which only hoisted the snapshot, measured at parity because of that per-row clone.
- The batch paths take the volume's mapping from the snapshot's own segment, as the UPDATE path does, instead of a second manager lookup.
- `find_segment_row` and `find_segment_row_retry` lost their last callers and are removed.

## Measured

100k-row table sealed by close and reopen, 20k-row child table, both binaries from one probe source, interleaved, 8 rounds.

| Shape | main | this PR | |
|---|---|---|---|
| INL join, 5000 probes into cold rows | 871 us | 779 us | 1.12x best, 1.13x median |
| `id IN (1000 literals)` on cold rows | 3774 us | 3787 us | unchanged |

The IN shape is unchanged because it does not go through the id fetch at all: the profile shows a full volume scan with `InListExpr::evaluate_fast` per row. That is a separate finding (a PK IN list on sealed data should be an id fetch) and not part of this PR.

## Tests

No behaviour change; volume unit tests (54) and the volume, snapshot, segment, concurrency, upsert and checkpoint integration targets pass.

### DIFF
--- a/src/storage/volume/table.rs
+++ b/src/storage/volume/table.rs
@@ -922,11 +922,11 @@ impl SegmentedTable {
     /// mid-statement on cold-volume access. Segments compacted away after
     /// the snapshot are simply found in the snapshot's older volumes,
     /// which is exactly the statement's view of the data.
-    fn find_segment_row_in(
+    fn find_segment_row_in<'a>(
         &self,
-        snap: &super::manifest::StatementSnapshot,
+        snap: &'a super::manifest::StatementSnapshot,
         row_id: i64,
-    ) -> Option<(u64, super::manifest::ColdSegment, usize)> {
+    ) -> Option<(u64, &'a super::manifest::ColdSegment, usize)> {
         if self.hot.has_row_id(row_id) {
             return None;
         }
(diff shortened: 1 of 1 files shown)


semihalev:
Closing: both findings are right, and fixing them properly removes the win.

With a seal read guard held across the batch, a metadata-only `cold_snapshot()`, and column data loaded per segment on demand (`ensure_volume` once per segment switch, never per row), the INL join over sealed rows measures 1.01x best / 0.99x median against main over 8 interleaved rounds. The 1.12x this PR reported came from `statement_snapshot()` calling `ensure_columns()`, which decompressed every segment up front, so later `get_row` calls skipped the LZ4 work the profile had shown. That is the bulk load in finding 2, not a cheaper lookup.

The per-row lock reads and the segment id `Vec` the finder used to take are not where the time goes; row decode is. Nothing here is worth shipping. Branch deleted.

PR #76, opened 2026-09-02
perf(executor): probe IN-list ids even when the list pushed down as a range

semihalev:
## Why

Found while measuring Wave H storage fetches. `EXPLAIN SELECT * FROM t WHERE id IN (1,2,3)` shows `Index Scan using __pk_t_id, Index Cond: id >= 1 AND <= 3`: pushdown turns an IN list into a min..max range. With ids spread across the table the range covers everything, so ten ids on a sealed 100k-row table cost 3 ms (a full volume scan with `InListExpr` per row, per `sample`), and the hot table walked its whole PK index for a thousand ids.

`try_in_list_index_optimization` already probes the listed ids directly, but the call site only tried it when `needs_memory_filter` was set, which a pushable IN list never leaves behind.

## What

- The fast path is tried whenever the WHERE has an IN list on the primary key or an indexed column; it returns `None` otherwise, so other shapes are untouched.
- The path applies LIMIT and OFFSET itself when there is no ORDER BY but reported `order_limit_applied = false`, so the caller applied them a second time. `test_in_list_with_offset` caught it the moment plain IN lists reached the path. It now returns whether it applied them.

## Measured

100k rows, both binaries built from one probe source, interleaved, best of 3.

| Shape | main | this PR |
|---|---|---|
| sealed table, `id IN (10)` | 3.0 ms | 3.8 us |
| sealed table, `id IN (100)` | 3.1 ms | 22 us |
| sealed table, `id IN (1000)` | 3.2 ms | 270 us |
| hot table, `id IN (10)` | 18.5 us | 2.9 us |
| hot table, `id IN (1000)` | 201-247 us | 200-209 us |
| hot table, `age IN (3)` (secondary index, 3226 rows) | 68-73 us | 69-71 us |

Row counts identical on every shape.

## Tests

`test_spread_pk_in_list_probes_ids`: spread ids with ORDER BY, an extra predicate, LIMIT, NOT IN, COUNT, and LIMIT/OFFSET without ORDER BY applied exactly once. The four IN-list and index optimizer targets pass (71 tests).

Rows without ORDER BY now come back in id order from the probe rather than scan order, which is not a guarantee either way.

### DIFF
--- a/src/executor/index_optimizer.rs
+++ b/src/executor/index_optimizer.rs
@@ -1216,7 +1216,7 @@ impl Executor {
         table_alias: Option<&str>,
         ctx: &ExecutionContext,
         classification: &Arc<QueryClassification>,
-    ) -> Result<Option<(Box<dyn QueryResult>, CompactArc<Vec<String>>)>> {
+    ) -> Result<Option<(Box<dyn QueryResult>, CompactArc<Vec<String>>, bool)>> {
         // Extract IN list info: (column_name, values, is_negated, remaining_predicate)
         let (column_name, values, is_negated, remaining_predicate) =
             match Self::extract_in_list_info(where_expr, ctx) {
@@ -1264,7 +1264,7 @@ impl Executor {
                     CompactArc::clone(&output_columns),
                     RowVec::new(),
                 );
-                return Ok(Some((Box::new(result), output_columns)));
+                return Ok(Some((Box::new(result), output_columns, false)));
             }
         }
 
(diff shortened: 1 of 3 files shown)


semihalev:
Review round: all four findings fixed with a red-first test (`test_in_list_fast_path_keeps_query_semantics`), see the follow-up commit.

- `extract_in_list_info` now ANDs whatever a nested AND hands back with that level's sibling, so `id IN (..) AND grp = 1 AND v = 10` keeps every conjunct.
- Window functions and an ORDER BY key that is not projected stay on the normal pipeline (gate: `!has_window_functions && !order_by_needs_extra_columns`).
- With DISTINCT or DISTINCT ON the path applies no LIMIT/OFFSET and reports `limit_applied = false`.

Not touched: `extract_in_subquery_info` has the same dropped-leftover shape in its AND arm. It feeds the IN-subquery path, not this one; flagging it rather than widening the PR.

PR #81, opened 2026-09-03
perf(join): build projected join rows from the two inputs, resolve the projection without a map

semihalev:
## Why

The two join rows where SQLite still led in BENCHMARKS.md: INNER JOIN (19.7 vs 12.1 us) and Self JOIN (12.6 vs 8.8 us). A `sample` profile of the INNER JOIN benchmark shape put 16% of the time in `combine_owned_into_buffer` (mostly `RowStorage::into_vec`) and 7% in `compute_join_projection`.

- `combine_owned_into_buffer` and `create_combined_row` converted both input rows to `Vec<Value>` before picking the one or two projected columns. For a shared row that clones every column; for an owned `CompactVec` it reallocates.
- `compute_join_projection` built a combined `Vec<String>` of every column name plus a lowercase `StringMap` on each execution of the statement.

## What

- `Row::take_or_clone(idx)`: moves the value out of owned storage, clones it out of shared storage. The two combine paths take only the projected values from the rows they own.
- `compute_join_projection` resolves each SELECT column by scanning the two column lists (usually under 20 names) instead of building a map, with the same rules as `build_column_index_map`: exact case-insensitive match first, then a qualified column's base name when exactly one column carries it. A qualified reference is matched piecewise, without formatting the name.

## Measured

Probe with the benchmark's tables (10k users, 50k orders, same indexes), 10 interleaved rounds against `main`, best / median:

| Shape | main | this PR | |
|---|---|---|---|
| INNER JOIN with LIMIT 100 | 19.3 / 20.7 us | 14.8 / 15.3 us | 1.30x / 1.35x |
| Self JOIN on age, LIMIT 100 | 12.0 / 12.4 us | 9.0 / 9.1 us | 1.33x / 1.36x |
| LEFT JOIN + GROUP BY | 52.7 / 54.0 us | 52.8 / 54.0 us | parity |
| Complex JOIN+GROUP+HAVING | 60.1 / 61.4 us | 60.4 / 62.0 us | parity |

The two aggregate shapes do not use projection pushdown, so they are unchanged by design.

## Tests

No behaviour change; the three join targets (25 tests) and the `utils` and `index_nested_loop` unit tests (87) pass, plus both full suites.

Still on the profile for this shape, for a follow-up: the hash index copies the whole id list for `status = 'completed'` (16.7k ids) on every execution although the outer side only needs a few hundred rows, and the outer side re-checks the status predicate on rows the index already matched exactly.

### DIFF
--- a/src/core/row.rs
+++ b/src/core/row.rs
@@ -491,6 +491,17 @@ impl Row {
         self.storage.into_vec()
     }
 
+    /// One value out of a row that is being consumed: moved out of owned
+    /// storage, cloned out of shared storage. Avoids converting the whole
+    /// row when only a few columns are wanted.
+    #[inline]
+    pub fn take_or_clone(&mut self, idx: usize) -> Value {
+        match &mut self.storage {
+            RowStorage::Owned(vec) => std::mem::take(&mut vec[idx]),
+            RowStorage::Shared(arc) => arc[idx].clone(),
+        }
+    }
+
     /// Extract the first value, consuming the row
     #[inline]
     pub fn take_first_value(self) -> Option<Value> {
--- a/src/executor/operators/index_nested_loop.rs
+++ b/src/executor/operators/index_nested_loop.rs
@@ -199,30 +199,23 @@ impl IndexNestedLoopJoinOperator {
     /// Uses `get_unchecked` for performance - projection indices are validated
     /// at query planning time against the table schemas.
     #[inline]
-    fn combine_owned_into_buffer(&mut self, outer: Row, inner: Row) {
+    fn combine_owned_into_buffer(&mut self, mut outer: Row, mut inner: Row) {
         match &self.projection {
             Some(proj) => {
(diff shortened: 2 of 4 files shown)


semihalev:
Review round: all three findings are in the follow-up commit.

- A repeated source (`SELECT o.v, o.v`) is cloned instead of moved when the projection repeats a source. This was already wrong on `main` through `into_values` for both storage kinds (second use came back NULL); the new `test_inl_join_projects_a_repeated_column_twice` fails there.
- Duplicate full column names resolve to the later column, as the index map they replaced did, for bare and qualified lookups; a unit test pins it together with the fallback rules.
- Case folds the Unicode way again: ASCII names stay allocation-free, others compare through `char::to_lowercase`, so `ÄNAME` keeps the fused projection.

Writing the sealed-rows variant of that test turned up a separate, older bug: after a checkpoint, index probes on a file database stop seeing sealed rows entirely. That gets its own PR; the sealed-rows test lives there.

PR #82, opened 2026-09-03
fix(storage): keep sealed rows visible to executor index probes

semihalev:
## Why

Found while writing a sealed-rows test for #81. On a file-backed database, every executor path that probes a secondary index directly stops seeing rows once they are sealed into volumes, which happens on `PRAGMA CHECKPOINT`, on the background checkpoint cycle, and on close:

```sql
PRAGMA CHECKPOINT;
SELECT u.name, o.amount FROM users u INNER JOIN orders o ON u.id = o.user_id;
-- 0 rows (60 before the checkpoint; COUNT(*) over the same join still 60)
INSERT INTO orders VALUES (61, 5, 61.0);
-- the same join now returns 1 row: only the hot one
```

`SegmentedTable::get_index_on_column` and the engine's `get_all_indexes` hand out the hot index, whose ids cover unsealed rows only. Affected: index nested loop join with a secondary index (`IndexLookupStrategy::SecondaryIndex`), IN-list / IN-subquery / IN-hashset probes, correlated EXISTS and COUNT through an index, recursive CTE probes, the GROUP BY btree path. Not affected: primary-key probes, plain `WHERE col = v` scans (the volume filter path reads sealed rows through zone maps and bloom filters), memory databases.

This is the "hot index after seal" item from the July review; it was marked not reproducible then.

## What

- `Table::lookup_index_on_column` and `StorageEngine::get_lookup_indexes`: an index the executor may probe on its own, meaning its ids cover every visible row. The default implementations delegate to the existing methods, so memory tables are unchanged.
- The rule for a persistent database is static, not a check on the current segment state. Such a database can seal at any moment, so the engine hands out every one of its tables through `SegmentedTable`, creating the segment manager on first use (the same helper the seal path uses). `SegmentedTable::lookup_index_on_column` returns `None` and `get_lookup_indexes` returns empty for a persistent database, so no segment-state check sits on the probe path. Before this, a persistent table had no manager until its first seal and came back as a bare `MVCCTable`; a handle taken before the first checkpoint kept reading the hot store after sealed rows were removed from it.
- Inside the wrapper, the hot-only short circuits (`scan`, `collect_*`, `fetch_rows_by_ids_into`, `select*`, partition and index min/max, grouped aggregates) run under the seal read fence through one helper, so the `has_segments` decision and the hot snapshot cannot be split by a seal. The readers that merge a hot value with cold values (`row_count`, `fast_row_count`, `sum_column`, `min_column`, `max_column`, the filtered and grouped aggregates, the partition readers) hold the fence for their whole body, so a seal cannot complete between the hot and the cold half; `seal_overlap()` alone did not cover a seal that finished in between. The remaining hot-first paths (active ids, distinct values, unique conflict, index min/max) merge into sets or min/max, where a row seen on both sides changes nothing.
- The probing consumers switch to the lookup variants and therefore use their scan paths on persistent databases. DDL, SHOW and EXPLAIN keep `get_index_on_column` / `get_all_indexes` for metadata. The vector-search probe is left as is.

Persistent databases lose the direct-probe fast paths for these shapes until a cold-aware index exists; a wrong answer is not a fast path.

## Tests

`tests/sealed_index_visibility_test.rs`: INL join with and without LIMIT, LEFT JOIN, IN list, correlated COUNT, correlated EXISTS, IN subquery; asserted before a checkpoint, after it, after a hot insert on top, and after a reopen. Fails on `main` at the first post-checkpoint assertion (0 rows instead of 60). The join, index optimizer, correlated subquery and subquery coverage targets pass (134 tests), plus both full suites.

### DIFF
--- a/src/executor/cte.rs
+++ b/src/executor/cte.rs
@@ -2191,12 +2191,12 @@ impl Executor {
             if schema.columns[pk_idx].name_lower == table_col_lower {
                 // It's a primary key lookup - most efficient!
                 IndexLookupStrategy::PrimaryKey
-            } else if let Some(index) = table.get_index_on_column(&table_col_unqualified) {
+            } else if let Some(index) = table.lookup_index_on_column(&table_col_unqualified) {
                 IndexLookupStrategy::SecondaryIndex(index)
             } else {
                 return Ok(None);
             }
-        } else if let Some(index) = table.get_index_on_column(&table_col_unqualified) {
+        } else if let Some(index) = table.lookup_index_on_column(&table_col_unqualified) {
             IndexLookupStrategy::SecondaryIndex(index)
         } else {
             return Ok(None);
--- a/src/executor/index_optimizer.rs
+++ b/src/executor/index_optimizer.rs
@@ -895,7 +895,7 @@ impl Executor {
 
         // If not PK, check for index
         let index = if !is_pk_column {
-            match table.get_index_on_column(&column_name) {
+            match table.lookup_index_on_column(&column_name) {
                 Some(idx) => Some(idx),
                 None => return Ok(None), // No PK, no index, can't optimize
             }
@@ -1249,7 +1249,7 @@ impl Executor {
         // and stops at LIMIT; so it is only taken when a memory filter is
         // needed anyway. Primary key values are the row ids themselves.
         let index = if !is_pk_column {
-            match table.get_index_on_column(&column_name) {
+            match table.lookup_index_on_column(&column_name) {
                 Some(idx) => Some(idx),
                 None => return Ok(None), // No PK, no index, can't optimize
             }
@@ -1607,7 +1607,7 @@ impl Executor {
 
         // If not PK, check for index (only for non-negated IN)
         let index = if !is_pk_column {
-            match table.get_index_on_column(&column_name) {
+            match table.lookup_index_on_column(&column_name) {
                 Some(idx) => Some(idx),
                 None => return Ok(None), // No PK, no index, can't optimize
             }
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -9179,7 +9179,7 @@ impl Executor {
         }
 
         // Check if there's a secondary index on the inner column
-        if let Some(index) = table.get_index_on_column(&inner_col_unqualified) {
+        if let Some(index) = table.lookup_index_on_column(&inner_col_unqualified) {
             return Some((
                 table_name,
                 IndexLookupStrategy::SecondaryIndex(index),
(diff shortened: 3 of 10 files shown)


semihalev:
Addressed the TOCTOU finding by making the rule static rather than guarding the probe.

- The engine marks every table it builds as seal-capable when persistence is configured (`MVCCTable::mark_seal_capable` in `get_table`). `MVCCTable::lookup_index_on_column` returns `None` for such a table, `SegmentedTable::lookup_index_on_column` always returns `None`, and `get_lookup_indexes` returns empty when the engine has a persistence manager.
- No segment-state check remains on the probe path, so there is no window between a check and the probe, and the pre-first-seal case (table not yet wrapped in `SegmentedTable`) is covered by the same rule.
- Memory databases are untouched and keep the direct probes.

`sealed_index_visibility_test` plus the join, index optimizer, correlated subquery and subquery coverage targets pass (134 tests), and both full suites are green.

semihalev:
Second round, on the bare `MVCCTable` handoff.

- `get_table_for_transaction` now wraps every table of a persistent database in `SegmentedTable`, creating the segment manager on first use through the helper the seal path already uses. A handle taken before the first checkpoint reads hot plus cold through the wrapper after it; no caller of a persistent database ever holds a bare `MVCCTable`.
- Inside the wrapper, the 20 hot-only short circuits (`scan`, `collect_*`, `fetch_rows_by_ids_into`, `select*`, partition and index min/max, grouped aggregates) run under the seal read fence via one helper, so the `has_segments` decision and the hot snapshot are one step against a concurrent seal. The hot-first paths (sum/min/max, active ids, distinct, unique conflict) keep their overlap and skip-set protocol, which already handles a seal landing on either side of the hot read.
- `MVCCTable` lost the `seal_capable` flag; the wrapper's `lookup_index_on_column` is the single rule.
- Side effect checked: a never-sealed table now gets an empty `manifest.bin` at the next full checkpoint. Every manager is stamped with the same checkpoint LSN, drop deletes the directory and rename moves it, and the loader skips managers without a schema, so replay start and cleanup are unchanged.

Memory databases are untouched. `sealed_index_visibility_test` plus the join, index optimizer, correlated subquery, subquery coverage and transaction residual targets pass (137 tests); both full suites green.

semihalev:
Third round, on the aggregate readers.

- Every `SegmentedTable` reader that merges a hot value with cold values now holds the seal read fence for its whole body: `row_count`, `fast_row_count`, `sum_column`, `min_column`, `max_column`, `compute_filtered_aggregates`, `compute_grouped_aggregates`, `get_partition_count`, `collect_rows_grouped_by_partition`, `get_rows_for_partition_value`. A seal cannot complete between the hot and the cold half any more, so the first-seal `COUNT(*) = 0`, the reversed-order double count and the SUM/AVG double count are closed at the source. `seal_overlap()` checks stay as they were.
- The four of those that went through the `unsealed` helper take the plain `has_segments` check under the outer fence instead, so the fence is never re-entered on one thread (parking_lot's fair `RwLock` would deadlock a recursive read behind a waiting seal). Checked that no fenced body calls another fenced method.
- Left alone on purpose: `get_active_row_ids`, `compute_distinct_values`, `get_partition_values`, `find_unique_conflict_row_id`, `has_row_id`, `get_index_min_value`/`max_value`. They merge into sets, booleans or min/max, and a row seen on both sides of a seal changes nothing there.

Aggregate, group by, persistence and volume targets pass, plus both full suites.

PR #83, opened 2026-09-03
fix(join): a LIMIT join keeps fetching until the limit is met

semihalev:
## Why

Started as the J2 join fixed-cost pass and turned into a correctness fix. Both LIMIT pushdowns on joins fetch one capped chunk of the driving side and never come back for more:

- The index nested loop join reads `4 x LIMIT` outer rows (min 100). When most outer rows have no match, the result stops short. With 1000 orders of which 50 point at an existing user, `SELECT o.amount, u.name FROM orders o INNER JOIN users u ON o.user_id = u.id LIMIT 100` returns 20 rows instead of 50.
- A cross-table WHERE was applied after the rows had been counted against the limit, so `... WHERE (u.id + o.id) % 2 = 0 LIMIT 10` returns 5 rows when 25 qualify.
- The GROUP BY semi-join reduction caps the left side at `LIMIT` (`2 x LIMIT` for INNER JOIN). When the first left rows have no match, or HAVING or a right-side filter drops groups, it returns fewer groups than exist, down to none: `SELECT u.name, COUNT(o.id) FROM users u INNER JOIN orders o ON u.id = o.user_id GROUP BY u.id, u.name LIMIT 10` returns 0 rows when the matched users sit at the end of the table.

## What

- Index nested loop with LIMIT: the outer side is fetched in chunks. The first chunk is the limit itself (min 100, as before) instead of four times it. When the joined rows stop short of the limit and the chunk was full, the join runs again with a chunk four times larger. The operator reports how many outer rows it pulled so a full chunk is detected exactly. The cross-table filter is applied per joined row by a shared collector, so only surviving rows count toward the limit.
- GROUP BY semi-join reduction: `execute_join_source` wraps the join in the same kind of loop. The reduction records the left chunk it used and whether it was full; when the final rows (after HAVING, OFFSET and LIMIT) stop short and the chunk was full, the join reruns with a left chunk four times larger. The reduction is only exact when each group comes from one left row, so GROUP BY now has to cover the left table's primary key; the initial chunk includes OFFSET. It stays off under a window function, `DISTINCT` or `DISTINCT ON`, which run on the grouped rows afterwards and need every group.
- `QueryResult::exact_len` gives the rerun check the length of a fresh materialized or limited result without iterating it. This is what keeps LEFT JOIN + GROUP BY at parity.

## Measurements

Probe at benchmark scale (10k users, 50k orders, memory), best of 6 interleaved runs against a main binary built from the same probe source:

| Query | main | this PR |
|---|---|---|
| INNER JOIN, LIMIT 100 | 14 us | 12 us |
| Self JOIN, LIMIT 100 | 8 us | 6 us |
| LEFT JOIN + GROUP BY, LIMIT 100 | 52 us | 52 us |
| INNER JOIN + GROUP BY + HAVING, LIMIT 50 | 60 us | 59 us |

The two wins come from the smaller first outer chunk: 100 outer rows instead of 400 for a join that matches every row.

## Tests

`tests/join_limit_shortfall_test.rs`: outer rows without a match (both join orders, with a WHERE, LIMIT 30), cross-table filter counted after filtering, GROUP BY reduction with late matches, with HAVING, with a right-side filter on a LEFT JOIN, a check that the grouped values are complete rather than partial, a window function over the grouped rows, and DISTINCT over them. All fail on `main`. Join, limit, hash join, aggregation targets pass, plus both full suites.

## Left open

GROUP BY with `LIMIT n OFFSET m` applies LIMIT before OFFSET, on a single table as well (50 groups: `LIMIT 10 OFFSET 45` returns 0 rows, `LIMIT 10 OFFSET 5` returns 5). Pre-existing on `main`, aggregation path, separate fix.

### DIFF
--- a/src/executor/operators/index_nested_loop.rs
+++ b/src/executor/operators/index_nested_loop.rs
@@ -121,6 +121,8 @@ pub struct IndexNestedLoopJoinOperator {
     // State tracking
     opened: bool,
     outer_exhausted: bool,
+    // Outer rows pulled so far; the executor uses it to see a full chunk
+    outer_rows_seen: usize,
 }
 
 impl IndexNestedLoopJoinOperator {
@@ -175,6 +177,7 @@ impl IndexNestedLoopJoinOperator {
             true_expr: ConstBoolExpr::true_expr(),
             opened: false,
             outer_exhausted: false,
+            outer_rows_seen: 0,
         }
     }
 
@@ -293,6 +296,10 @@ impl IndexNestedLoopJoinOperator {
         }
     }
 
+    pub fn outer_rows_seen(&self) -> usize {
+        self.outer_rows_seen
+    }
+
     /// Look up matching inner rows for the current outer row.
     /// Uses internal buffers to avoid allocations.
     fn lookup_inner_rows(&mut self, key_value: &Value) -> Result<()> {
@@ -340,6 +347,7 @@ impl IndexNestedLoopJoinOperator {
         match self.outer.next()? {
             Some(row_ref) => {
                 let outer_row = row_ref.into_owned();
+                self.outer_rows_seen += 1;
 
                 // Get the join key value from the outer row
                 let key_value = match outer_row.get(self.outer_key_idx) {
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -3899,6 +3899,7 @@ impl Executor {
             left_alias.as_deref(),
             left_table,
             &join_source.condition,
+            classification,
         );
 
         let (left_rows, left_columns, right_rows, right_columns) = if let Some(plan) =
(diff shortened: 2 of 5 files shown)


semihalev:
Both findings addressed; the reduction now stays off for them.

- `get_semijoin_reduction_limit` returns `None` when the query has a window function, `DISTINCT` or `DISTINCT ON`. Both run on the grouped rows after the join produced them, so a chunk that meets the limit is not enough: `COUNT(*) OVER ()` counted the chunk's groups, and DISTINCT could collapse the rows below the limit after the rerun check had passed.
- Two tests in `tests/join_limit_shortfall_test.rs`: the window one saw `COUNT(*) OVER () = 31` instead of 50 before the gate, the DISTINCT one got 4 distinct values instead of 5 without its half of the gate. Both pass now; the join, limit, group by, window and aggregation targets pass, plus both full suites.

PR #86, opened 2026-09-03
fix(executor): apply OFFSET before LIMIT, and only once, on every early-exit path

semihalev:
## Why

Follow-up to the note left in #83: `GROUP BY ... LIMIT n OFFSET m` applied LIMIT before OFFSET. Looking for the cause turned up a family of early-exit paths with the same shape of mistake, in both directions.

Rows are built, then the caller applies OFFSET and LIMIT unless the path reports them as applied. With 50 groups and `LIMIT 10 OFFSET 45`:

- Grouped aggregation with a WHERE, window fetches by partition or index order, and the streaming GROUP BY over a BTREE index stop at `LIMIT` rows, so after the caller skips 45 nothing is left: 0 rows instead of 5.
- The IN and NOT IN subquery index path applies OFFSET and LIMIT itself but reported them as not applied, so the caller skipped 45 again: 0 rows, also for `OFFSET 45` without a LIMIT. NOT IN on the primary key trimmed its ids before the shared tail trimmed them once more.
- The hash-set index path, which an EXISTS is rewritten into when no direct index probe is available (a file-backed database after #82), had the same flag problem: `EXISTS (...) LIMIT 10 OFFSET 45` returned 0 rows there.
- The early termination scan for a subquery in WHERE (`amount > (SELECT AVG(...))`) collected `LIMIT + OFFSET` rows and reported both as applied without skipping any: 50 rows instead of 5.

## What

- `aggregation_limit`, the window fetch limits (`literal_fetch_limit`) and the streaming GROUP BY early exit (`extract_limit_hint`) build `LIMIT + OFFSET` rows.
- The early termination scan drops the OFFSET rows and cuts to LIMIT before reporting the pair as applied.
- `try_in_subquery_index_optimization` and `try_in_hashset_index_optimization` return whether they applied OFFSET and LIMIT (they do unless ORDER BY or DISTINCT has to run on the rows first), and the callers pass that through. The early termination scan stays off under DISTINCT for the same reason.
- The index-order window fetch keeps the full fetch for LEAD, for a frame that reaches forward, for a named window, and under any frame other than an explicit ROWS one for every function except ROW_NUMBER, RANK, DENSE_RANK, LAG and FIRST_VALUE: the implicit RANGE frame includes the current row's later peers, which a fetch cut at LIMIT + OFFSET rows may not hold. The cut fetch is also used only when every window in the SELECT orders by exactly the column and direction the fetch follows, with no PARTITION BY; a second window ordered another way would rank the cut subset. And the whole index-order path, cut or not, is taken only when that column is the primary key or declared NOT NULL: the index places NULLs where Value's ordering does, the window's ORDER BY places them the other way round by default, and the pre-sorted aggregate path takes the rows as they come. The NOT IN primary key walk stops early only when nothing later can drop or reorder rows, and leaves the trimming to the shared tail.

Without an OFFSET nothing changes on these paths; the early termination scan moves at most `LIMIT` rows once more.

## Tests

`tests/offset_before_limit_test.rs`: GROUP BY on a scan, a filtered scan, with HAVING, with ORDER BY, over joins (INNER, LEFT, HAVING), the streaming BTREE GROUP BY, window fetches by index order and by partition, keyset pagination, a CTE, DISTINCT, and IN / NOT IN / EXISTS / scalar subqueries in WHERE, plus an ORDER BY over the IN subquery path to show the set is sorted before it is limited, DISTINCT over the early termination, IN subquery and hash-set paths, LEAD over the window fetch, SUM / LAST_VALUE over an ORDER BY window whose last peer group the cut fetch would split, two ROW_NUMBER windows ordered by different columns, and a nullable BTREE column checked against the same data in a table without the index, for ROW_NUMBER, COUNT(*) and SUM in ASC, DESC and explicit NULLS FIRST / LAST, plus the limited result against the same slice of the unlimited one. The GROUP BY, window, partition, IN subquery, NOT IN and scalar cases fail on `main`; the EXISTS case fails on `main` with the file-backed test feature. Subquery, limit, group by, window, aggregation and join targets pass (420 tests), plus both full suites.

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -418,16 +418,23 @@ impl Executor {
             && stmt.having.is_none()
             && !stmt.distinct;
 
+        // OFFSET is applied after aggregation, so the groups it skips must be
+        // built too: the pushed limit is LIMIT + OFFSET
         let aggregation_limit = if can_push_limit {
-            stmt.limit.as_ref().and_then(|limit_expr| {
-                let mut eval = ExpressionEval::compile(limit_expr, &[])
-                    .ok()?
-                    .with_context(ctx);
+            let eval_usize = |expr: &Expression| {
+                let mut eval = ExpressionEval::compile(expr, &[]).ok()?.with_context(ctx);
                 eval.eval_slice(&Row::new()).ok().and_then(|v| match v {
                     crate::core::Value::Integer(n) if n >= 0 => Some(n as usize),
                     _ => None,
                 })
-            })
+            };
+            let limit = stmt.limit.as_deref().and_then(eval_usize);
+            match stmt.offset.as_deref() {
+                None => limit,
+                Some(offset_expr) => {
+                    limit.and_then(|l| eval_usize(offset_expr).map(|o| l.saturating_add(o)))
+                }
+            }
         } else {
             None
         };
--- a/src/executor/index_optimizer.rs
+++ b/src/executor/index_optimizer.rs
@@ -529,20 +529,16 @@ impl Executor {
         None
     }
 
-    /// Check if all window functions in the statement are safe for LIMIT pushdown.
-    ///
-    /// Safe functions (don't depend on total row count):
-    /// - ROW_NUMBER, RANK, DENSE_RANK
-    /// - LAG, LEAD
-    /// - FIRST_VALUE, LAST_VALUE, NTH_VALUE
-    ///
-    /// Unsafe functions (depend on total row count):
-    /// - NTILE (bucket size depends on total count)
-    /// - PERCENT_RANK (formula: (rank-1)/(total-1))
-    /// - CUME_DIST (formula: rows_up_to_current/total)
-    pub(crate) fn is_window_safe_for_limit_pushdown(stmt: &SelectStatement) -> bool {
+    /// Whether every window function in the statement can run on a fetch cut at
+    /// LIMIT + OFFSET rows in the given index order: each window must follow
+    /// that order exactly and must not look past the current row.
+    pub(crate) fn is_window_safe_for_limit_pushdown(
+        stmt: &SelectStatement,
+        order_col: &str,
+        ascending: bool,
+    ) -> bool {
         for col_expr in &stmt.columns {
-            if !Self::is_expr_window_safe(col_expr) {
+            if !Self::is_expr_window_safe(col_expr, order_col, ascending) {
                 return false;
             }
         }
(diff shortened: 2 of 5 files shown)


semihalev:
All three addressed.

- Early termination scan: `can_early_terminate` now requires no `DISTINCT` and no `DISTINCT ON`, so the rows reach DISTINCT untrimmed and the caller applies LIMIT and OFFSET after it.
- IN subquery and hash-set paths: one `limit_here = order_by.is_empty() && !distinct && distinct_on.is_empty()` governs the early id trimming, the NOT IN early stop, the late trimming and the returned flag, the same rule the IN-list path uses.
- Window fetch by index order: `is_expr_window_safe` now rejects `LEAD`, any frame whose start or end reaches forward (`FOLLOWING`, `UNBOUNDED FOLLOWING`) and a named window reference, so those keep the full fetch.

Tests added to `tests/offset_before_limit_test.rs`: `DISTINCT` over the early termination scan, over the IN subquery path, over the hash-set path (EXISTS), a LIMIT-only DISTINCT check, and `LEAD(id) OVER (ORDER BY id) ... LIMIT 10 OFFSET 45` asserting the last row's LEAD is 56. The DISTINCT IN-subquery and LEAD cases failed before the change on both database modes. Related targets, clippy and both full suites pass.

semihalev:
Addressed. `is_expr_window_safe` now keeps the cut fetch only for an explicit `ROWS` frame (forward bounds already excluded) or, under any other frame including the implicit `RANGE` one, for functions that ignore the frame end: `ROW_NUMBER`, `RANK`, `DENSE_RANK`, `LAG`, `FIRST_VALUE`. Aggregates, `LAST_VALUE` and `NTH_VALUE` over an ORDER BY window fall back to the full fetch.

Test: `SUM(amount) OVER (ORDER BY category) ... LIMIT 1 OFFSET 44` over 100 rows with two rows per category returned 1629 instead of 1702 before the change (the 45th row's peer was past the cut); the explicit `RANGE BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW` form and `LAST_VALUE` are covered too. Both database modes, the window and limit targets, clippy and both full suites pass.

semihalev:
Addressed. `is_window_safe_for_limit_pushdown` takes the column and direction the fetch follows and every window in the SELECT, nested ones included, must order by exactly that key in that direction with no PARTITION BY; otherwise the full fetch is used. The doc on the function describes the current rule.

Test: `ROW_NUMBER() OVER (ORDER BY id), ROW_NUMBER() OVER (ORDER BY amount DESC) ... LIMIT 10 OFFSET 45` over 100 orders; before the change the second window ranked amount 46 as 10 within the 55-row cut, now 55. Both database modes, the window and limit targets, clippy and both full suites pass.

semihalev:
Addressed by restricting rather than teaching the fetch about NULL placement: the cut fetch is used only when the window's ordering column is the primary key or declared `NOT NULL`. Any other column keeps the full fetch, whose rows the window sorts itself with its own NULL placement.

Test: a BTREE-indexed nullable column with 20 NULL rows inserted first; for `ASC`, `DESC`, `ASC NULLS FIRST` and `DESC NULLS LAST` at two offsets, `LIMIT 10 OFFSET n` must equal rows `n..n+10` of the unlimited result. The `ASC OFFSET 0` case returned the NULL rows before the change. Both database modes, the window and limit targets, clippy and both full suites pass.

semihalev:
Confirmed and fixed. A probe against a table without the index showed the pre-sorted aggregate path using the index's NULL order: `COUNT(*) OVER (ORDER BY k)` gave 20 for a NULL row where the sorted path gives 100, and `SUM(k)` gave NULL where it gives the total. `ROW_NUMBER` and `LAST_VALUE` matched because the ranking path sorts for itself. The whole index-order path is now taken only when the ordering column is the primary key or declared `NOT NULL`; a nullable column falls to `collect_all_rows` with no pre-sorted state, so the window sorts.

Test: a BTREE-indexed nullable column and the same data in a table without the index as the oracle; `ROW_NUMBER()`, `COUNT(*)` and `SUM(k)` over `ORDER BY k` in `ASC`, `DESC`, `ASC NULLS FIRST` and `DESC NULLS LAST` must match the oracle row for row, and `LIMIT 10 OFFSET n` must equal the same slice of the unlimited result; the first NULL row's `ROW_NUMBER` is asserted as 81. The aggregate cases fail on the previous commit. The peer-group test's `category` column is now `NOT NULL` so it still exercises the cut fetch. One aside: the window sorter ignores an explicit `NULLS FIRST` / `NULLS LAST` on every path (only the cache key sees it); pre-existing and separate. `tests/index_optimizer_test.rs` declares `products.price` as `NOT NULL` so its `ROW_NUMBER ... LIMIT 5` test keeps exercising the pushdown.

PR #87, opened 2026-09-03
perf(join): size the first GROUP BY chunk for HAVING and grow it twofold

semihalev:
## Why

Measuring the benchmark on `main` against the BENCHMARKS.md baseline (435bf023) showed one real regression: `Complex JOIN+GROUP+HAVING` at 213 to 240 us against 52 us. Since #83 the GROUP BY reduction reruns the join with a larger left chunk when the final rows stop short of the limit. On the benchmark's data the first chunk of 100 active users yields 44 groups after `HAVING COUNT(o.id) > 1`, so every execution reran with 400 rows. The baseline's 52 us measured an incomplete answer: 44 rows for `LIMIT 50`.

## What

- `semijoin_caps` doubles the first left chunk under HAVING, on top of the doubling an INNER JOIN already had, so the benchmark query is answered by one 200-row pass.
- A rerun doubles the chunk instead of quadrupling it, so a rerun that still happens costs at most twice the pass before it.

## Measurements

Random data shaped like the benchmark (10k users, 50k orders, random user ids and statuses), best of 5 interleaved runs against a 435bf023 binary built from the same probe source:

| Query | baseline | this PR |
|---|---|---|
| INNER JOIN + GROUP BY + HAVING, LIMIT 50 | 49 us (44 rows) | 90 us (50 rows) |
| LEFT JOIN + GROUP BY, LIMIT 100 | 52 us | 53 us |
| INNER JOIN, LIMIT 100 | 19 us | 11 us |
| Self JOIN, LIMIT 100 | 12 us | 6 us |

One full benchmark run on this branch puts the row at 91 us; `main` sits at 213 to 240 us over six runs. The remaining gap to the baseline is the price of the complete answer: about 115 left rows are needed for 50 groups here, and the pass cost grows with the chunk.

## Tests

`tests/join_limit_shortfall_test.rs` still covers the rerun and gains a huge-LIMIT case: the doubled chunk is capped at `i64::MAX` before it becomes the synthetic LIMIT literal, so `LIMIT 5000000000000000000` with HAVING no longer overflows into a negative literal. Join, limit and group by targets pass, plus both full suites.

### DIFF
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -3785,7 +3785,7 @@ impl Executor {
             if !short {
                 return Ok((result, columns, flag, deferred));
             }
-            left_cap = Some(pass.cap.saturating_mul(2));
+            left_cap = Some(pass.cap.saturating_mul(2).min(i64::MAX as usize));
         }
     }
 
@@ -9025,11 +9025,11 @@ impl Executor {
         if stmt.having.is_some() {
             cap = cap.saturating_mul(2);
         }
-        Some((cap, limit))
+        // The chunk becomes a LIMIT literal, which is an i64
+        Some((cap.min(i64::MAX as usize), limit))
     }
 
-    /// Check if semi-join reduction optimization can be applied and return parameters
-    /// Returns Some((limit_value, left_key_col, right_key_col)) if applicable, None otherwise
+    /// Check if semi-join reduction optimization can be applied and return its plan
     ///
     /// Conditions for semi-join reduction:
     /// 1. INNER JOIN or LEFT JOIN (not RIGHT or FULL)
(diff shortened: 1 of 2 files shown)


semihalev:
Addressed. The first chunk and every rerun cap the doubled value at `i64::MAX` before it becomes the synthetic LIMIT literal. Test added to `tests/join_limit_shortfall_test.rs`: a LEFT JOIN with HAVING and an INNER JOIN with HAVING, both with `LIMIT 5000000000000000000`, failed on this branch with `LIMIT must be non-negative, got -8446744073709551616` and now return every group. The doc on `get_semijoin_reduction_limit` describes the current sizing (INNER doubles, HAVING doubles again, reruns double the chunk) and no longer claims a tuple return.

PR #88, opened 2026-09-03
fix(window): keep same-named windows apart and honour NULLS FIRST and LAST

semihalev:
## Why

Two wrong results found while reviewing #86, both on `main`:

- `SELECT id, ROW_NUMBER() OVER (ORDER BY id), ROW_NUMBER() OVER (ORDER BY amount DESC) FROM t` returned `[id, <rank by amount>, NULL]`: both windows were filed under the name `ROW_NUMBER()`, so the second overwrote the first and the third column found nothing. Aliases avoided it.
- `ROW_NUMBER() OVER (ORDER BY k ASC NULLS FIRST)` ranked the NULL rows last: the window sorter placed NULLs by direction alone and only the cache key ever looked at `NULLS FIRST` / `LAST`.

## What

- `parse_window_functions` gives every top-level window, plain or aliased, a unique internal name from its position (`__wf_<column>_0`, the scheme nested windows already use); the select item keeps `FUNCTION()` or the alias as the output name and uses the internal name as the value key. No alias is ever a key, so an alias cannot collide with another window's name and two windows may share an alias.
- `ColumnarOrderByValues` carries a NULL placement per ORDER BY column: the explicit `NULLS FIRST` / `LAST` when given, otherwise last ascending and first descending, which is what the sorter did before. The typed integer and float fast paths write the NULL rows first or last by that flag, and the single-column and multi-column comparators order a NULL against a value by it instead of reversing a fixed "NULLs greatest" rule.

Without an explicit NULLS clause nothing changes.

## Tests

`tests/window_columns_nulls_test.rs`: two and three same-named windows keep their own columns (`ROW_NUMBER`, `COUNT(*)` with and without ORDER BY); an alias spelling another window's internal name and two windows with the same alias keep theirs; NULL rows and the first non-NULL row are ranked as expected for `ASC`, `DESC`, `ASC NULLS LAST`, `ASC NULLS FIRST`, `DESC NULLS FIRST`, `DESC NULLS LAST`. Both fail on `main`. Window, order, aggregation and group by targets pass in both database modes, plus both full suites.

### DIFF
--- a/src/executor/window.rs
+++ b/src/executor/window.rs
@@ -246,6 +246,9 @@ pub struct ColumnarOrderByValues {
     columns: Vec<Vec<Value>>,
     /// Ascending flags: one per ORDER BY column
     ascending: Vec<bool>,
+    /// Where NULLs sort, one per ORDER BY column: NULLS FIRST / LAST when
+    /// given, otherwise last ascending and first descending
+    nulls_first: Vec<bool>,
     /// Number of rows
     num_rows: usize,
 }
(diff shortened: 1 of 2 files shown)


semihalev:
Addressed. `parse_window_functions` now gives an aliased window the same position-based internal name a plain one gets (`__wf_<column>_0`), and the aliased select item uses that name as the value key with the alias only as `output_name`. No alias is ever a key any more.

Tests added to `tests/window_columns_nulls_test.rs`: the alias `__wf_1_0` next to a plain `ROW_NUMBER()`, and two windows both aliased `rn`; both returned `[1, 100, NULL]`-style overwritten columns before and keep their own values now. Window, order, aggregation and alias targets pass in both database modes, plus both full suites.

PR #89, opened 2026-09-03
perf(join): continue a short LIMIT join from where its outer chunk ended

semihalev:
## Why

Measuring `main` against the BENCHMARKS.md baseline showed `INNER JOIN` at 23 us against 19.7 us. #83 made a LIMIT join fetch an outer chunk the size of the limit and rerun from the start with a chunk four times larger when the joined rows stop short. On the benchmark the delete phase leaves one order in a hundred without a user, so every execution of `... INNER JOIN ... WHERE o.status = 'completed' LIMIT 100` joined 100 rows, came up one short, and joined 400 more. The same query with no missing user runs in 11.6 us.

## What

- The continuation keeps the joined rows and fetches only the outer rows after the chunk, through a new row offset on `execute_table_expression_with_filter_limit` (a `SELECT * ... LIMIT n OFFSET m` on the table). The next chunk is sized from the shortfall and the match rate so far, at least 32 rows and at most three times what was fetched, so a run with no matches at all still grows geometrically.
- The first chunk carries a sixteenth of slack over the limit (min 100), so the odd missing match is covered without a second fetch.
- The outer fetch always hands back exactly the chunk for a plain table: when `execute_simple_table_scan` says it did not apply the synthetic LIMIT and OFFSET (an AS OF path, or a path that stopped early at their sum without applying them), the fetch applies the pair through the normal limited result. A table function applies them too when it has no filter, generating `offset + limit` rows; a CTE, a view, a subquery or a filtered table function comes back whole. The join drops its chunk limit when the first fetch says so.
- Every outer fetch and the inner probes of a join run on one transaction: `ExecutionContext` carries a `StatementSnapshot`, created by the join before its first fetch, and `execute_simple_table_scan` reads through it instead of opening a standalone transaction when the statement has no explicit one. The snapshot opens read committed, the default; a continuation fetches one row of overlap and, if that row is no longer the one the chunk ended on, the join restarts on a snapshot isolation transaction, where every chunk is consistent by construction. Snapshot isolation on every join would cost about two microseconds, a fifth of the benchmark's join rows.
- Inside an explicit transaction the scan reads through that transaction, whose isolation may let a commit move rows between fetches, so the outer side is fetched whole there, and the inner table comes from that transaction too, so the join sees the rows the transaction inserted, changed or deleted.

## Measurements

Full benchmark binary, one machine:

| Row | baseline 435bf023 | main | this PR |
|---|---|---|---|
| INNER JOIN | 19.7 us | 23.1 us | 12.0 to 13.0 us |
| Self JOIN | 13.4 us | 7.5 us | 7.7 to 8.1 us |
| Complex JOIN+GROUP+HAVING | 49.5 us (44 rows for LIMIT 50) | 80.4 us | 80 to 96 us |

Random benchmark-shaped data, best of 5 against a baseline binary: INNER JOIN 19 to 12 us, Self JOIN 11 to 6 us, LEFT JOIN + GROUP BY 52 to 51 us. Against the pre-PR main over eight rounds the medians are equal: INNER JOIN 12.0 vs 12.1 us, Self JOIN 7.0 vs 7.0 us.

## Tests

`tests/join_limit_shortfall_test.rs` keeps covering a join that needs several continuations (5 matches per 100 outer rows), asserts that the rows across chunk boundaries are exactly the matching ones, none repeated or missing, and checks a CTE, a view and a temporal (AS OF) source of exactly 100 rows on the outer side, which must return their five matches once each, plus a table function outer whose only match sits past the first chunk, an outer with a correlated filter, and the shortfall queries inside an explicit transaction with an uncommitted matching pair and an uncommitted delete. Join, limit and offset targets pass, plus both full suites.

### DIFF
--- a/src/executor/context.rs
+++ b/src/executor/context.rs
@@ -677,6 +677,32 @@ pub fn cache_exists_correlation(
 ///
 /// Note: This struct uses Arc for immutable shared data to make cloning cheap
 /// during correlated subquery processing where context is cloned per row.
+/// A transaction shared by every read of one statement
+#[derive(Clone)]
+pub struct StatementSnapshot(Arc<Mutex<Box<dyn crate::storage::traits::Transaction>>>);
+
+impl StatementSnapshot {
+    pub fn new(transaction: Box<dyn crate::storage::traits::Transaction>) -> Self {
+        Self(Arc::new(Mutex::new(transaction)))
+    }
+
+    pub fn get_table(
+        &self,
+        name: &str,
+    ) -> crate::core::Result<Box<dyn crate::storage::traits::Table>> {
+        self.0
+            .lock()
+            .unwrap_or_else(|poisoned| poisoned.into_inner())
+            .get_table(name)
+    }
+}
+
+impl std::fmt::Debug for StatementSnapshot {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str("StatementSnapshot")
+    }
+}
+
 #[derive(Debug, Clone)]
 pub struct ExecutionContext {
     /// Query parameters ($1, $2, etc.) - wrapped in Arc for cheap cloning
@@ -710,6 +736,9 @@ pub struct ExecutionContext {
     cte_data: Option<Arc<CteDataMap>>,
     /// Current transaction ID for CURRENT_TRANSACTION_ID() function
     transaction_id: Option<u64>,
+    /// One transaction for every fetch of a statement that has no explicit
+    /// transaction, so its reads share a snapshot
+    statement_snapshot: Option<StatementSnapshot>,
 }
 
 /// Type alias for CTE data: (columns, rows) with Arc for zero-copy sharing
@@ -745,6 +774,7 @@ impl ExecutionContext {
             outer_columns: None,
             cte_data: None,
             transaction_id: None,
+            statement_snapshot: None,
         }
     }
 
@@ -905,6 +935,7 @@ impl ExecutionContext {
             outer_columns: self.outer_columns.clone(),
             cte_data: self.cte_data.clone(),
             transaction_id: self.transaction_id,
+            statement_snapshot: self.statement_snapshot.clone(),
         }
     }
 
@@ -925,6 +956,7 @@ impl ExecutionContext {
             outer_columns: self.outer_columns.clone(),
             cte_data: self.cte_data.clone(),
             transaction_id: self.transaction_id,
+            statement_snapshot: self.statement_snapshot.clone(),
         }
     }
 
@@ -960,6 +992,7 @@ impl ExecutionContext {
             outer_columns: Some(outer_columns), // Arc clone = cheap
             cte_data: self.cte_data.clone(),    // Arc clone = cheap
             transaction_id: self.transaction_id,
+            statement_snapshot: self.statement_snapshot.clone(),
         }
     }
 
@@ -1012,6 +1045,7 @@ impl ExecutionContext {
             outer_columns: self.outer_columns.clone(),
             cte_data: Some(cte_data),
             transaction_id: self.transaction_id,
+            statement_snapshot: self.statement_snapshot.clone(),
         }
     }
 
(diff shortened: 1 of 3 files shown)


semihalev:
Both addressed, and the test the description promised is in this time.

- P1: `execute_table_expression_with_filter_limit` returns whether it applied the limit and offset; only the plain-table path does, a CTE, view, subquery or table function comes back whole with `false`. The join drops its chunk limit when the first fetch says so, and the GROUP BY reduction's `chunk_full` uses the flag too. Test: a CTE and a view of exactly 100 orders with five matching users, `LIMIT 10`, must return the five ids once each.
- P2: `ExecutionContext` gains a `StatementSnapshot` (one transaction behind an `Arc<Mutex>`). The join creates one before the first outer fetch and uses it for every outer chunk and for the inner table, and `execute_simple_table_scan` reads through it instead of opening a standalone transaction when the statement carries one (an explicit transaction still wins, an `AS OF` query still takes its own). The inner probes and the outer side of a join were on different snapshots before as well; they now share one.
- The chunk-boundary test is in the commit now: `LIMIT 100` over 50 matches spread across several continuations returns exactly the matching ids, and `LIMIT 30` returns 30 distinct ones.

Join, CTE, transaction, limit and offset targets pass, clippy is clean, both full suites pass.

semihalev:
All three addressed.

- P1, table function outer: `execute_tvf_for_join` takes the offset too and reports whether it applied the pair; without a filter it generates `offset + limit` rows and drops the first `offset`, with a filter it comes back whole and unbounded, so the join never chunks a source it cannot refetch. Test: `generate_series(1, 200)` against a table holding only 150, `LIMIT 1`, returns 150.
- P1, `bounded`: the plain-table fetch now passes through the third value of `execute_simple_table_scan` instead of `true`. Test: a temporal outer (`first100 AS OF TIMESTAMP ... f`) of exactly 100 rows with five matches, `LIMIT 10`, returns the five ids once each.
- P2: the statement snapshot is opened under `SnapshotIsolation` (`new_statement_snapshot`), so a commit between two fetches of one statement cannot move rows across a chunk boundary.

Join, table function, AS OF, CTE, limit and offset targets pass, clippy is clean, both full suites pass.

semihalev:
Both addressed.

- P1: when `execute_simple_table_scan` says it did not apply the synthetic LIMIT and OFFSET, the fetch wraps its result in the normal `LimitedResult` (offset then limit) and reports the chunk as bounded, so a path that stopped early at their sum still yields exactly the chunk and the join keeps continuing. Test: an outer with a correlated filter (`o.amount >= (SELECT MIN(...) ... WHERE o2.user_id = o.user_id)`), `LIMIT 100` over 50 spread matches, must return exactly the matching ids.
- P2: inside an explicit transaction the outer side is fetched whole (`outer_limit = None`), since the scan reads through that transaction and its isolation may let a commit move rows between fetches. Test: the shortfall queries inside `BEGIN ... COMMIT` return the complete answer. Keeping the chunks under an explicit snapshot-isolation transaction would need an isolation getter on the `Transaction` trait; left out here.

Join, table function, AS OF, CTE, limit and offset targets pass, clippy is clean, both full suites pass.

semihalev:
Addressed. `join_table` hands the join its inner table from the active transaction when one exists and from the statement snapshot otherwise, at both the first probe and every continuation, so both sides of the join read through the same transaction inside `BEGIN`.

The explicit-transaction test now inserts an uncommitted user with a matching order and expects the join to count 51 and to find the new order by id, then deletes a user and expects 50, then rolls back and expects 50 again. Join, table function, AS OF, CTE, transaction, limit and offset targets pass, clippy is clean, both full suites pass.

semihalev:
Follow-up on the snapshot cost seen in bench-compare (INNER JOIN 11.60 to 14.28 us, Self JOIN 6.99 to 8.27 us): isolating it showed the snapshot isolation transaction itself costs about 2 us per join, and even `set_isolation_level(ReadCommitted)` is not free.

- The statement snapshot now opens read committed, the default, with no isolation call.
- A continuation fetches one row of overlap (offset `fetched - 1`) and compares it with the row the chunk ended on (`IndexNestedLoopJoinOperator::last_outer_row`). If it is unchanged the rows after it are the continuation; if a commit moved it, or it is gone, the join restarts on a snapshot isolation transaction and every later chunk is consistent by construction. The overlap check runs only when a continuation happens.
- First-chunk slack is a sixteenth of the limit instead of an eighth; the eighth was not the cost, but a sixteenth covers the benchmark's miss rate just as well.

Random benchmark-shaped data, eight interleaved rounds against the pre-PR main, medians: INNER JOIN 12.0 vs 12.1 us, Self JOIN 7.0 vs 7.0 us. All shortfall tests pass, plus the related targets, clippy and both full suites.

PR #90, opened 2026-09-04
perf(join): stream the groups of a GROUP BY + LIMIT join through the right side's index

semihalev:
## Why

`Complex JOIN+GROUP+HAVING` went from 45 us on the baseline (an incomplete answer: 44 rows for `LIMIT 50`) to about 85 us for the complete answer after #83 and #87. The profile of the correct path showed why: the GROUP BY reduction fetches a chunk of left rows, fetches every matching right row through an IN filter, hash joins them, hashes them again into groups and applies HAVING, computing all 200 groups of the chunk to keep 50. More than half the time was the hash aggregation and the joined rows it materialises.

## What

A streaming path for a GROUP BY + LIMIT join whose groups are the left rows (the reduction's own eligibility: INNER or LEFT JOIN, GROUP BY covering the left primary key and only left columns, LIMIT, no ORDER BY, window function or DISTINCT):

- The right side is probed per left row through its index with the streaming index nested loop operator; the right-side WHERE is its residual filter, a cross-table WHERE is applied per joined row.
- Each group is folded as the joined rows stream past, with the aggregate functions from the registry (`accumulate`, `result`, `reset`), COUNT(*) included, DISTINCT honoured. The group columns cover the left primary key, so a change in them is the next left row.
- HAVING runs on the finished group through the same `RowFilter::with_aliases` binding the hash path uses, the SELECT projection through `apply_post_aggregation_expressions`, and the scan stops after LIMIT + OFFSET groups have passed.
- The outer side is fetched in chunks with the continuation and snapshot rules of #89; inside an explicit transaction it is read whole.
- Not covered, and left to the reduction: an aggregate over an expression or with FILTER or ORDER BY, a GROUP BY expression, ROLLUP, CUBE or GROUPING SETS, a WHERE with a subquery, a LEFT JOIN with a right-side filter, a right key without an index.

## The statement snapshot opens read committed

#89 opens a statement snapshot for every limited index join so that the outer chunks and the inner probes read one state. It is opened under snapshot isolation, and setting that level costs about two microseconds per join, a fifth of the benchmark's INNER JOIN and Self JOIN rows. The first commit here changes that for the limited join and the grouped join alike:

- The snapshot opens read committed, the engine default, so no isolation call is made. Every outer fetch and every inner probe of one join still go through that one transaction.
- A continuation refetches one row of overlap before the next chunk. When it is still the row the previous chunk ended on, nothing moved under the join and the rows after it are the continuation. When a commit moved it, the join restarts from the beginning on a snapshot isolation transaction, where the chunks line up by construction. The index nested loop operator exposes the outer row it ended on for that check.
- The first chunk keeps a sixteenth of slack over the limit instead of an eighth.
- Inside an explicit transaction nothing changes: the outer side is read whole through the transaction.

## Measurements

Random benchmark-shaped data (`join_rand` probe, 10k users, 30k orders, unseeded), best of 6 interleaved runs against current main, in us:

| Query | main | this PR |
|---|---|---|
| INNER JOIN + GROUP BY + HAVING, LIMIT 50 | 88 | 48 |
| LEFT JOIN + GROUP BY, LIMIT 100 | 51 | 44 |
| INNER JOIN, LIMIT 100 | 13 | 12 |
| Self JOIN, LIMIT 100 | 8 | 6 |

`examples/benchmark` rows, three interleaved runs against current main on the same machine, median in us:

| Row | main | this PR |
|---|---|---|
| INNER JOIN | 18.5 | 15.4 |
| LEFT JOIN + GROUP BY | 55.3 | 48.2 |
| CTE + JOIN | 43.9 | 44.3 |
| Complex JOIN+GROUP+HAVING | 106.8 | 74.8 |
| Self JOIN (same age) | 10.5 | 9.6 |

The Complex row averages 20 executions with no warmup, about 1.2 ms of work, so a single slow first execution moves it a lot: one of the three runs here reported 306 us with the other 19 executions at their usual 55 to 65 us, the same effect main shows at a smaller size. Read that row from several runs.

## Tests

`tests/grouped_join_stream_test.rs` checks every shape against the same query with an ORDER BY, which takes the general hash path: INNER and LEFT, COUNT(*), COUNT(col), COUNT(DISTINCT), SUM, AVG, MIN, MAX, aliases, HAVING with and without early stop, a right-side and a left-side WHERE, a cross-table WHERE, an expression over aggregates in the SELECT, GROUP BY on the key alone, and OFFSET. A ROLLUP keeps its total row, and a WHERE with a scalar, IN or expression subquery matches the general path, for the grouped and the plain limited join. `join_limit_shortfall_test` still covers the reduction and the continuation. Join, group by, aggregation, subquery and CTE targets pass, plus both full suites.

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -742,7 +742,7 @@ impl Executor {
 
     /// Apply post-aggregation expressions to the result
     /// This handles expressions like `CASE WHEN SUM(x) > 100 THEN 'big' ELSE 'small' END`
-    fn apply_post_aggregation_expressions(
+    pub(crate) fn apply_post_aggregation_expressions(
         &self,
         stmt: &SelectStatement,
         ctx: &ExecutionContext,
@@ -1358,7 +1358,7 @@ impl Executor {
 
     /// Parse aggregate functions from SELECT list
     /// Returns: (aggregations, non_agg_columns, post_agg_expressions)
-    fn parse_aggregations(
+    pub(crate) fn parse_aggregations(
         &self,
         stmt: &SelectStatement,
     ) -> Result<(Vec<SqlAggregateFunction>, Vec<String>)> {
@@ -1643,7 +1643,7 @@ impl Executor {
     }
 
     /// Parse GROUP BY clause
-    fn parse_group_by(
+    pub(crate) fn parse_group_by(
         &self,
         stmt: &SelectStatement,
         _base_columns: &[String],
--- a/src/executor/operators/index_nested_loop.rs
+++ b/src/executor/operators/index_nested_loop.rs
@@ -300,6 +300,11 @@ impl IndexNestedLoopJoinOperator {
         self.outer_rows_seen
     }
 
+    /// The outer row the join is on, or ended on
+    pub fn last_outer_row(&self) -> Option<&Row> {
+        self.current_outer_row.as_ref()
+    }
+
     /// Look up matching inner rows for the current outer row.
     /// Uses internal buffers to avoid allocations.
     fn lookup_inner_rows(&mut self, key_value: &Value) -> Result<()> {
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -6567,6 +6567,16 @@ impl Executor {
         }) {
             return Ok(None);
         }
+        if stmt.group_by.modifier != crate::parser::ast::GroupByModifier::None {
+            return Ok(None);
+        }
+        if [left_filter, right_filter, cross_filter]
+            .into_iter()
+            .flatten()
+            .any(Self::has_subqueries)
+        {
+            return Ok(None);
+        }
         let group_by = self.parse_group_by(stmt, &[])?;
         let mut group_names = Vec::with_capacity(group_by.len());
         for item in &group_by {
@@ -6580,7 +6590,7 @@ impl Executor {
 
         // One transaction for the outer fetches and the inner probes; inside an
         // explicit transaction the outer side is read whole
-        let snapshot = match ctx.statement_snapshot() {
+        let mut snapshot = match ctx.statement_snapshot() {
             Some(snapshot) => snapshot.clone(),
             None => self.new_statement_snapshot(crate::core::IsolationLevel::ReadCommitted)?,
         };
@@ -6839,10 +6849,10 @@ impl Executor {
                 outer_result = more;
                 continue;
             }
-            let restart =
+            snapshot =
                 self.new_statement_snapshot(crate::core::IsolationLevel::SnapshotIsolation)?;
-            ctx_join = ctx.with_statement_snapshot(restart.clone());
-            inner_table = Some(self.join_table(&restart, &table_name)?);
+            ctx_join = ctx.with_statement_snapshot(snapshot.clone());
+            inner_table = Some(self.join_table(&snapshot, &table_name)?);
             consistent = true;
             groups.clear();
             outer_limit = Some(fetched.saturating_add(next));
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -6520,6 +6520,81 @@ impl Executor {
         Ok(StatementSnapshot::new(transaction))
     }
 
+    /// True when a filter carries a subquery anywhere inside it. The grouped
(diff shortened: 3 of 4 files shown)


semihalev:
Benchmark rows, `examples/benchmark` release build, three interleaved runs against main at #88 on the same machine, median in us:

| Query | main (#88) | this branch |
|---|---|---|
| INNER JOIN | 19.9 | 11.8 |
| LEFT JOIN + GROUP BY | 46.1 | 38.3 |
| CTE + JOIN | 36.0 | 37.4 |
| Complex JOIN+GROUP+HAVING | 59.8 | 59.0 |
| Self JOIN (same age) | 13.0 | 7.1 |
| HAVING complex | 87.9 | 85.8 |

INNER JOIN and Self JOIN carry #89. Complex is the complete answer on both sides now. On the benchmark's data about half the users pass HAVING, so the first chunk already holds the 50 groups and the two paths do similar work; the streaming path wins on data with a lower pass rate (85 to 47 us on random data in the description).

One thing to know when reading the Complex row: it averages 20 executions with no warmup, about 1.2 ms of work in total. In roughly one run out of ten the first of those executions takes 2 to 3 ms (a per-execution trace showed the other 19 at 52 to 64 us, and no single index probe or group finish over 300 us inside the slow one), and that alone reports the row as 190 to 300 us. Main at #88 shows the same effect at a smaller size (99 to 115 us). Thirty back to back runs of this branch: 29 between 56 and 103 us, one at 312.

semihalev:
Three findings confirmed and fixed in the last commit:

1. ROLLUP, CUBE and GROUPING SETS: the streaming path now takes a plain GROUP BY only (`GroupByModifier::None`). A test with `GROUP BY ROLLUP(u.id) LIMIT 1000` lost the total row before the fix and matches the general path now.
2. WHERE subqueries: the raw filters were compiled into the join filters, where a subquery has no executor. A query whose left, right or cross filter contains a subquery now takes the reduction, which resolves subqueries first. Tests with a scalar, an IN and an expression subquery returned 0 rows before the fix and match the general path now. The plain limited join was checked with the same WHERE and was already correct.
3. Snapshot after a restart: the restart now replaces `snapshot` itself, so every later chunk probes the inner table through the snapshot isolation transaction, the same as the non-grouped path.

semihalev:
Confirmed and fixed. The gate now uses its own detector, `filter_has_subquery`, an exhaustive match over every `Expression` variant with no wildcard: CAST, CASE, LIKE, BETWEEN, lists, function arguments, FILTER and ORDER BY of a call, windows, and the table-source variants (those count as a subquery; they never belong in a WHERE). An IN list still counts only when a subquery sits in it, so the benchmark's `status IN (...)` keeps streaming (46 to 49 us on the random-data probe, main 90 to 94). The nested test covers BETWEEN between two subqueries and an IN subquery, both matching the general path.

Writing those tests turned up three wrong answers on main itself, outside this PR, probed on `fa4ded20` with a plain 100 users x 300 orders table:

- A subquery inside CAST or inside a function call, `WHERE o.amount > CAST((SELECT AVG(amount) FROM orders) AS FLOAT)` or `WHERE COALESCE((SELECT MAX(amount) FROM orders WHERE user_id = u.id), 0) > 900`, returns 0 rows on every path: `has_subqueries` and `try_process_expression_subqueries` do not look inside those containers, so the raw subquery reaches the evaluator and gives NULL.
- A correlated EXISTS in a join's WHERE, `WHERE EXISTS (SELECT 1 FROM orders x WHERE x.user_id = u.id AND x.amount > 900)`, returns 0 rows on every join path, while the same predicate written as an IN subquery or as `(SELECT MAX(amount) ... WHERE x.user_id = u.id) > 900` returns 28 groups and 84 rows, and the single-table EXISTS returns 28.
- A subquery inside CASE, `WHERE CASE WHEN o.amount > (SELECT AVG(amount) FROM orders) THEN 1 ELSE 0 END = 1`, is right with ORDER BY (100 groups) and returns 0 with LIMIT: the GROUP BY reduction compiles that filter raw.

The streaming path now bails on all of those shapes, which lands it on main's answer. The fixes belong in the WHERE subquery processing and the reduction, with their own tests; I kept them out of this PR to keep the scope to the join and will open them separately.

PR #91, opened 2026-09-04
perf(join): filter the inner fetch and project the joined row in the grouped stream

semihalev:
## Why

After #90 the benchmark's `Complex JOIN+GROUP+HAVING` gives the complete answer at about 53 us against 45 us for the incomplete one it used to give. A sample of the streaming path on that query put a fifth of the time in the right-side `status IN (...)` predicate, evaluated on every joined row through the join filter's expression machine, and another tenth in building the full twelve-column joined row for groups that read four of its columns.

## What

- The right-side WHERE goes to the storage layer as the predicate of the inner fetch, through a new `with_inner_filter` on the index nested loop operator. It is converted by the same schema-aware pushdown a scan uses, against the inner table's schema, and only a predicate the pushdown takes whole replaces the residual filter; a partial or refused conversion (a function call, for one) keeps the residual on the joined row as before. For the benchmark's `status IN (...)` a non-matching order is never materialised.
- Without a cross-table filter the operator's projection pushdown builds only the columns the group keys and the aggregates read, in that order, so the group boundary check and the accumulators index a four-column row instead of a twelve-column one.

- Found on the way and fixed in the pushdown registry, which scans use as well: the OR, NOT and XOR rules converted their children with a helper that drops the partial flag, so `(amount > 10 AND status LIKE 'c!_%' ESCAPE '!') OR amount < 0` came back as a whole conversion of `(amount > 10) OR amount < 0` and the caller dropped its memory filter. The three rules now take a child only when it converted whole.

The join changes are inside the streaming grouped path; the plain limited join is untouched.

## Measurements

Random benchmark-shaped data (`join_rand` probe, 10k users, 30k orders, unseeded), best of 6 interleaved runs against main at #90, in us:

| Query | main | this PR |
|---|---|---|
| INNER JOIN + GROUP BY + HAVING, LIMIT 50 | 48 | 35 |
| LEFT JOIN + GROUP BY, LIMIT 100 | 44 | 37 |
| INNER JOIN, LIMIT 100 | 11 | 11 |
| Self JOIN, LIMIT 100 | 6 | 6 |

`examples/benchmark` rows, three interleaved runs against main at #90 on the same machine, median in us:

| Row | main | this PR |
|---|---|---|
| INNER JOIN | 15.9 | 15.2 |
| LEFT JOIN + GROUP BY | 51.7 | 41.2 |
| CTE + JOIN | 45.7 | 41.7 |
| Complex JOIN+GROUP+HAVING | 78.9 | 57.2 |
| Self JOIN (same age) | 9.1 | 8.7 |

The Complex row averages 20 executions with no warmup; read it from several runs.

## Tests

`grouped_join_stream_test` gained a right-filter test: an IN list with a range, a function call the storage layer cannot take, a right filter next to a cross-table filter (projection off), LIKE with a left filter, and an OR of two right-side predicates, each with and without an early stop, all against the general path. A second test with its own tables checks a `TIMESTAMP '...'` comparison and a `LIKE ... ESCAPE` on the right side, the two shapes the plain converter gets wrong. A third test computes the rows and groups of the OR with a partial child from the generated data and checks a plain scan and the grouped stream against them. Join and shortfall targets pass, plus both full suites.

### DIFF
--- a/src/executor/operators/index_nested_loop.rs
+++ b/src/executor/operators/index_nested_loop.rs
@@ -29,7 +29,7 @@ use crate::core::value::NULL_VALUE;
 use crate::core::{Result, Row, RowVec, Value};
 use crate::executor::expression::JoinFilter;
 use crate::executor::operator::{ColumnInfo, Operator, RowRef};
-use crate::storage::expression::ConstBoolExpr;
+use crate::storage::expression::{ConstBoolExpr, Expression};
 use crate::storage::traits::{Index, Table};
 use smallvec::SmallVec;
 
@@ -117,6 +117,8 @@ pub struct IndexNestedLoopJoinOperator {
 
     // Expression for fetching rows (always true - we apply residual separately)
     true_expr: ConstBoolExpr,
+    // A predicate on inner columns the fetch applies instead of the residual
+    inner_filter: Option<Box<dyn Expression>>,
 
     // State tracking
     opened: bool,
@@ -175,6 +177,7 @@ impl IndexNestedLoopJoinOperator {
             // Pre-allocate row buffer to avoid per-row allocation
             row_buffer: Row::with_capacity(total_cols),
             true_expr: ConstBoolExpr::true_expr(),
+            inner_filter: None,
             opened: false,
             outer_exhausted: false,
             outer_rows_seen: 0,
@@ -204,6 +207,13 @@ impl IndexNestedLoopJoinOperator {
             repeats_source,
         });
         self.schema = projected_schema;
+        self
+    }
+
+    /// A filter the inner fetch applies to each candidate row, for a
+    /// predicate on inner columns only
+    pub fn with_inner_filter(mut self, filter: Box<dyn Expression>) -> Self {
+        self.inner_filter = Some(filter);
         self
     }
 
@@ -340,9 +350,13 @@ impl IndexNestedLoopJoinOperator {
         }
 
         // Fetch matching rows directly into inner_rows buffer
+        let filter: &dyn Expression = match &self.inner_filter {
+            Some(filter) => filter.as_ref(),
+            None => &self.true_expr,
+        };
         self.inner_table.fetch_rows_by_ids_into(
             &self.row_id_buffer,
-            &self.true_expr,
+            filter,
             &mut self.current_inner_rows,
         )
     }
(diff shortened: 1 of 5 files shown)


semihalev:
Confirmed and fixed. The inner fetch's predicate now comes from `pushdown::try_pushdown` against the inner table's schema with the execution context, the same call a scan makes, and the residual join filter is dropped only when the registry converted the whole predicate (`needs_memory_filter == false`); a partial or refused conversion keeps the residual and no storage predicate. The plain converter is no longer used here.

Regression test added with its own tables: a `TIMESTAMP '...'` comparison and a `LIKE ... ESCAPE` on the right side, each checked for the full result and for an early stop against the general path. Before the fix the streaming path returned no rows for either shape while the general path returned its groups; the benchmark's `status IN (...)` still goes to storage, so the numbers in the description hold.

semihalev:
Both confirmed and fixed.

1. The registry's OR, NOT and XOR rules converted their children through `convert_expr`, which drops the partial flag, so `(o.amount > 10 AND o.status LIKE 'c!_%' ESCAPE '!') OR o.amount < 0` came back as a whole conversion of `(amount > 10) OR amount < 0`. That was wrong for a plain scan on main as well, not only for the inner fetch here: the new test counts the rows of that predicate on the table and got every order over 10 instead of the escaped-LIKE ones before the fix. The three rules now take a child only through `convert_expr_exact`, which returns a child that converted whole; anything else makes the rule refuse, and the caller keeps its memory filter. The AND rule already carried the flag. The test checks the scan count and the grouped stream's groups against values computed from the generated data, so it does not depend on another path being right.
2. The inner predicate is now converted once, against the first inner table's schema, and each chunk's operator gets a clone, so an expression fixed at conversion is the same on every chunk and the conversion is not repeated.

semihalev:
Confirmed and fixed: the XOR rule converts each operand once and clones the storage expression for its second place in `(a AND NOT b) OR (NOT a AND b)`, so both copies carry the same value. The partial-child test gained the XOR form of the same predicate, checked against the count computed from the generated data.

semihalev:
Correction to the last reply: the XOR form of the partial-child test did not land in the XOR commit, only the rule change did. It is in the commit after it, checked against the count computed from the generated data.

PR #92, opened 2026-09-04
fix(executor): resolve a WHERE subquery wherever it sits, on every join path

semihalev:
## Why

Writing the tests for #90 and #91 turned up four WHERE subquery shapes that return no rows on main, each verified against values computed from the generated data:

| Shape | Where it failed |
|---|---|
| `o.amount > CAST((SELECT AVG(amount) FROM orders) AS FLOAT)` | every path, scan and join |
| `CASE WHEN o.amount > (SELECT AVG(amount) FROM orders) THEN 1 ELSE 0 END = 1` | GROUP BY + LIMIT join |
| `COALESCE((SELECT MAX(amount) FROM orders x WHERE x.user_id = u.id), 0) > 200` | every path |
| `EXISTS (SELECT 1 FROM orders x WHERE x.user_id = u.id AND x.amount > 200)` | every join path |

## What

- **Detectors see every container.** The classification walker that sets `where_has_subqueries`, its correlated counterpart, and the executor's `has_subqueries` / `has_correlated_subqueries` now descend into CAST, LIKE, lists, DISTINCT and CASE, so a nested subquery is processed like a bare one.
- **The join's constant-false short-circuit skips a WHERE with a subquery.** It compiles the raw WHERE against no columns and evaluates it on an empty row; a subquery has no value there, so `EXISTS (...)` and `COALESCE((SELECT ...), 0) > 200` came out constant false and the join returned empty before running. That was the cause of the EXISTS and COALESCE rows above.
- **Uncorrelated subqueries fold before the filters are split.** The join path processes the WHERE once, then partitions it, so the reduction, the index nested loop and the grouped stream all see plain predicates instead of raw subqueries. The filter partition and the qualifier helpers learned the hashed set an IN subquery folds to, which they used to take for a constant and push to the wrong side.
- **Correlated predicates stay with the joined row.** The filter partition cannot see the qualifiers inside a subquery (a bare EXISTS was pushed to the left scan "arbitrarily", wrong as soon as it read the right side), so a correlated predicate now goes post-join, the fast paths are skipped, and the general path resolves its subqueries per joined row with that row as the outer row. The per-row processor also descends into function calls, CASE, CAST and LIKE now.

## Tests

`tests/where_subquery_containers_test.rs`: the four shapes above on the scan, the plain join, the join with LIMIT and the grouped join with ORDER BY and with LIMIT, plus a correlated EXISTS that reads both sides of the join (`x.amount > o.amount`), which can only be answered on the joined row. Every expected value is computed from the generated data. All failed on main before the fix. Join, subquery, CTE and aggregation targets pass, plus both full suites.

## Measurements

`examples/benchmark` join and subquery rows, three interleaved runs against main at #91 on the same machine, median in us. None of these queries has a subquery in a join's WHERE, so the change is not on their path; this is a parity check.

| Row | main | this PR |
|---|---|---|
| INNER JOIN | 14.0 | 14.7 |
| LEFT JOIN + GROUP BY | 40.2 | 39.2 |
| Scalar subquery | 9.3 | 9.7 |
| IN subquery | 111.8 | 110.2 |
| EXISTS subquery | 2.7 | 2.6 |
| CTE + JOIN | 39.5 | 40.3 |
| Complex JOIN+GROUP+HAVING | 54.5 | 56.2 |
| NOT IN subquery | 74.2 | 78.9 |
| NOT EXISTS subquery | 26.8 | 27.0 |
| Self JOIN (same age) | 8.3 | 8.2 |
| Nested subquery (3 lvl) | 367.5 | 370.4 |
| Correlated in SELECT | 265.5 | 266.8 |
| Compare with subquery | 5.8 | 5.8 |

### DIFF
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -405,20 +405,23 @@ impl Executor {
         // back as a column, since an aggregate cannot resolve one itself
         let lifted_stmt;
         let lifted_columns;
-        let (stmt, base_rows, base_columns) = match stmt
-            .columns
-            .iter()
-            .any(Self::has_subqueries)
-            .then(|| Self::lift_subqueries_in_aggregates(&stmt.columns))
+        let (stmt, base_rows, base_columns) = match Self::aggregates_hold_subquery(stmt)
+            .then(|| Self::lift_subqueries_in_aggregates(stmt))
             .flatten()
         {
-            Some((columns, lifted)) => {
-                let (rows, all_columns) =
-                    self.materialize_lifted_subqueries(&lifted, ctx, base_rows, base_columns)?;
-                lifted_stmt = SelectStatement {
-                    columns,
-                    ..stmt.clone()
-                };
+            Some((rewritten, lifted)) => {
+                let qualifier = stmt
+                    .table_expr
+                    .as_deref()
+                    .and_then(super::utils::get_table_alias_from_expr);
+                let (rows, all_columns) = self.materialize_lifted_subqueries(
+                    &lifted,
+                    ctx,
+                    base_rows,
+                    base_columns,
+                    qualifier.as_deref(),
+                )?;
+                lifted_stmt = rewritten;
                 lifted_columns = all_columns;
                 (&lifted_stmt, rows, lifted_columns.as_slice())
             }
@@ -5221,6 +5224,9 @@ impl Executor {
         classification: &std::sync::Arc<QueryClassification>,
     ) -> Result<Option<Box<dyn crate::storage::traits::QueryResult>>> {
         // classification is passed from caller to avoid redundant cache lookups
+        if Self::aggregates_hold_subquery(stmt) {
+            return Ok(None);
+        }
 
         // Quick eligibility checks using cached classification
         if classification.has_where {
@@ -5419,6 +5425,10 @@ impl Executor {
     ) -> Result<Option<Box<dyn crate::storage::traits::QueryResult>>> {
         use crate::storage::mvcc::version_store::AggregateOp;
 
+        if Self::aggregates_hold_subquery(stmt) {
+            return Ok(None);
+        }
+
         // --- Eligibility checks ---
 
         if !classification.has_where {
@@ -5598,7 +5608,7 @@ impl Executor {
         classification: &std::sync::Arc<QueryClassification>,
     ) -> Result<Option<Box<dyn crate::storage::traits::QueryResult>>> {
         // Quick eligibility checks using cached classification
-        if classification.has_where {
+        if classification.has_where || Self::aggregates_hold_subquery(stmt) {
             return Ok(None);
         }
         if classification.has_group_by {
@@ -5963,6 +5973,10 @@ impl Executor {
     ) -> Result<Option<Box<dyn QueryResult>>> {
         use crate::common::SmartString;
         use smallvec::SmallVec;
+
+        if Self::aggregates_hold_subquery(stmt) {
+            return Ok(None);
+        }
 
         // Quick eligibility checks
         if !classification.has_group_by {
@@ -6352,6 +6366,10 @@ impl Executor {
     ) -> Option<Box<dyn QueryResult>> {
         use crate::parser::ast::GroupByModifier;
         use crate::storage::mvcc::version_store::AggregateOp;
+
+        if Self::aggregates_hold_subquery(stmt) {
+            return None;
+        }
 
         // Only for GROUP BY without WHERE or HAVING
         if classification.has_where || !classification.has_group_by || classification.has_having {
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -401,6 +401,30 @@ impl Executor {
         base_rows: RowVec,
         base_columns: &[String],
     ) -> Result<Box<dyn QueryResult>> {
+        // A subquery inside an aggregate is resolved per input row and read
+        // back as a column, since an aggregate cannot resolve one itself
+        let lifted_stmt;
+        let lifted_columns;
+        let (stmt, base_rows, base_columns) = match stmt
+            .columns
+            .iter()
+            .any(Self::has_subqueries)
+            .then(|| Self::lift_subqueries_in_aggregates(&stmt.columns))
+            .flatten()
+        {
+            Some((columns, lifted)) => {
+                let (rows, all_columns) =
+                    self.materialize_lifted_subqueries(&lifted, ctx, base_rows, base_columns)?;
+                lifted_stmt = SelectStatement {
+                    columns,
+                    ..stmt.clone()
+                };
+                lifted_columns = all_columns;
+                (&lifted_stmt, rows, lifted_columns.as_slice())
+            }
+            None => (stmt, base_rows, base_columns),
+        };
+
         // Parse aggregations and group by columns
         let (aggregations, _non_agg_columns) = self.parse_aggregations(stmt)?;
         let group_by_columns = self.parse_group_by(stmt, base_columns)?;
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -1298,17 +1298,16 @@ impl Executor {
         None
     }
 
-    /// Execute GROUP BY aggregation and return raw columns/rows for window function processing
-    /// This is used when both GROUP BY and window functions are present in the query.
-    /// Window functions operate on the aggregated result.
-    pub(crate) fn execute_aggregation_for_window(
+    /// Aggregate, then run the window functions over the aggregate's output.
+    /// A subquery inside an aggregate is resolved per input row first, and
+    /// both stages read the same rewritten statement
+    pub(crate) fn execute_aggregation_then_windows(
         &self,
         stmt: &SelectStatement,
         ctx: &ExecutionContext,
         base_rows: &[(i64, Row)],
         base_columns: &[String],
-    ) -> Result<(Vec<String>, RowVec)> {
-        // A subquery inside an aggregate is resolved per input row first
+    ) -> Result<Box<dyn QueryResult>> {
         let lifted_stmt;
         let lifted_rows;
         let lifted_columns;
@@ -1340,7 +1339,21 @@ impl Executor {
             }
             None => (stmt, base_rows, base_columns),
         };
+        let (agg_columns, agg_rows) =
+            self.execute_aggregation_for_window(stmt, ctx, base_rows, base_columns)?;
+        self.execute_select_with_window_functions(stmt, ctx, &agg_rows, &agg_columns)
+    }
 
+    /// Execute GROUP BY aggregation and return raw columns/rows for window function processing
+    /// This is used when both GROUP BY and window functions are present in the query.
+    /// Window functions operate on the aggregated result.
+    fn execute_aggregation_for_window(
+        &self,
+        stmt: &SelectStatement,
+        ctx: &ExecutionContext,
+        base_rows: &[(i64, Row)],
+        base_columns: &[String],
+    ) -> Result<(Vec<String>, RowVec)> {
         // Parse aggregations and group by columns
         let (aggregations, _non_agg_columns) = self.parse_aggregations(stmt)?;
         let group_by_columns = self.parse_group_by(stmt, base_columns)?;
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -406,18 +406,19 @@ impl Executor {
         let lifted_stmt;
         let lifted_columns;
         let (stmt, base_rows, base_columns) = match Self::aggregates_hold_subquery(stmt)
-            .then(|| Self::lift_subqueries_in_aggregates(stmt))
+            .then(|| Self::lift_subqueries_in_aggregates(stmt, base_columns))
             .flatten()
         {
-            Some((rewritten, lifted)) => {
+            Some((rewritten, lifted, names)) => {
                 let qualifier = stmt
                     .table_expr
                     .as_deref()
                     .and_then(super::utils::get_table_alias_from_expr);
                 let (rows, all_columns) = self.materialize_lifted_subqueries(
                     &lifted,
+                    &names,
                     ctx,
-                    base_rows,
+                    &base_rows,
                     base_columns,
                     qualifier.as_deref(),
                 )?;
@@ -1307,6 +1308,39 @@ impl Executor {
         base_rows: &[(i64, Row)],
         base_columns: &[String],
     ) -> Result<(Vec<String>, RowVec)> {
+        // A subquery inside an aggregate is resolved per input row first
+        let lifted_stmt;
+        let lifted_rows;
+        let lifted_columns;
+        let (stmt, base_rows, base_columns) = match Self::aggregates_hold_subquery(stmt)
+            .then(|| Self::lift_subqueries_in_aggregates(stmt, base_columns))
+            .flatten()
+        {
+            Some((rewritten, lifted, names)) => {
+                let qualifier = stmt
+                    .table_expr
+                    .as_deref()
+                    .and_then(super::utils::get_table_alias_from_expr);
+                let (rows, all_columns) = self.materialize_lifted_subqueries(
+                    &lifted,
+                    &names,
+                    ctx,
+                    base_rows,
+                    base_columns,
+                    qualifier.as_deref(),
+                )?;
+                lifted_stmt = rewritten;
+                lifted_rows = rows;
+                lifted_columns = all_columns;
+                (
+                    &lifted_stmt,
+                    lifted_rows.as_slice(),
+                    lifted_columns.as_slice(),
+                )
+            }
+            None => (stmt, base_rows, base_columns),
+        };
+
         // Parse aggregations and group by columns
         let (aggregations, _non_agg_columns) = self.parse_aggregations(stmt)?;
         let group_by_columns = self.parse_group_by(stmt, base_columns)?;
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -1298,6 +1298,20 @@ impl Executor {
         None
     }
 
+    /// The ORDER BY of an aggregate that orders its own input, with the
+    /// NULLS placement the query asked for
+    pub(crate) fn aggregate_order_keys(
+        order_by: &[crate::parser::ast::OrderByExpression],
+    ) -> Vec<crate::functions::AggregateOrder> {
+        order_by
+            .iter()
+            .map(|order| crate::functions::AggregateOrder {
+                ascending: order.ascending,
+                nulls_first: order.nulls_first,
+            })
+            .collect()
+    }
+
     /// Aggregate, then run the window functions over the aggregate's output.
     /// A subquery inside an aggregate is resolved per input row first, and
     /// both stages read the same rewritten statement
@@ -2243,12 +2257,7 @@ impl Executor {
                 for (i, agg) in aggregations.iter().enumerate() {
                     if !agg.order_by.is_empty() {
                         if let Some(ref mut func) = agg_funcs[i] {
-                            let directions: Vec<bool> = agg
-                                .order_by
-                                .iter()
-                                .map(|o| o.ascending) // true = ASC, false = DESC
-                                .collect();
-                            func.set_order_by(directions);
+                            func.set_order_by(Self::aggregate_order_keys(&agg.order_by));
                         }
                     }
                 }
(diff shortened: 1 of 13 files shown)


semihalev:
All four confirmed and fixed, each with a test whose expected values come from the generated data.

1. Parent outer row: the per-row map now starts from `ctx.outer_row()`, and every outer reference in the processed predicate is bound to its value before the row filter compiles. That second part mattered: the row filter resolved `u2.id` to the joined row's `u.id` by unqualified name, so a join inside a scalar subquery whose WHERE compared a correlated subquery with the outer `u2.id` returned 90 for every outer row; main answered it through the left pushdown, and this branch now answers it on the post-join path (90, 87, 84). One shape stays open and is not this PR's: an EXISTS that reads the outer `u2` from inside the join returns 0 on main for the single-table form as well, so two levels of correlation through EXISTS were never supported; noted for a separate fix.
2. LIKE, list, DISTINCT and IN lists: `process_where_subqueries` and `process_correlated_where` now walk the same containers the detectors do, and an IN list may hold a subquery. Tests: `name LIKE (SELECT name ...) || '%'` and `id IN ((SELECT MIN(id) ...), 2)`, on the scan and on the join.
3. Simple CASE operand: `case.value` is walked in the classification walker and in both correlated detectors. Tests: an uncorrelated and a correlated operand.
4. ALL and ANY: `has_correlated_subqueries` recognises them, so the partition keeps a correlated one post-join, and `process_correlated_where` folds it under the row's context through the general processor. Test: `o.amount > ALL (SELECT amount FROM orders x WHERE x.user_id = u.id AND x.id <> o.id)` returns each user's largest order.

The Code Coverage job's FK timing threshold (105 ms against a 100 ms limit under instrumentation) is unrelated to this change.

semihalev:
All four confirmed and fixed, each with a test whose expected values come from the generated data.

1. IN list: `has_subqueries`, `has_correlated_subqueries` and the classification's correlated detector recurse into both sides of an IN, so a list holding a subquery is detected wherever it sits. Test: `DELETE FROM orders WHERE id IN ((SELECT MIN(id) FROM orders), 2)` deletes the two rows.
2. Parent-row binding: `substitute_outer_references` walks CAST, CASE (operand, branches, else), expression lists and DISTINCT. Test: the join inside a scalar subquery compares its correlated subquery with `CAST(u2.id AS FLOAT) * 10` and with a searched CASE over `u2.id` (90, 87, 84). A simple CASE with `u2.id` as its operand cannot be tested yet: once the operand is bound to a literal it hits a separate, pre-existing evaluator bug, `WHERE id > CASE 1 WHEN 1 THEN 10 ELSE 30 END` returns 0 rows on main for a plain table too (a column operand and a searched CASE are fine). Noted for its own fix.
3. Wrapped outer reference: `has_outer_column_reference` and `references_outer_columns` walk CAST, LIKE, DISTINCT, lists and the simple CASE operand. Test: `EXISTS (SELECT 1 FROM orders x WHERE x.user_id = CAST(u.id AS INTEGER) AND x.amount > 200)` on the join, plain and grouped.
4. Mixed predicates: the WHERE folds conjunct by conjunct, so an uncorrelated subquery folds even when another conjunct is correlated, and a derived table or CTE side receives a plain predicate. Test: the correlated EXISTS next to `o.amount > (SELECT AVG(amount) FROM orders)` against a derived table, a CTE and a plain table.

semihalev:
All four confirmed; three fixed with tests, the fourth fixed on the detection side with the execution gap left as a separate item.

1. Hashed IN: `substitute_outer_references` walks `InHashSet`, so a column of the surrounding query on the left of a correlated IN is bound before the row filter compiles. Test: `u2.id IN (SELECT x.user_id FROM orders x WHERE x.user_id = u.id)` inside a join inside a scalar subquery.
2. NULL in ALL and ANY: the converter keeps the NULLs of the set in view: with a NULL present, an ALL folds to `(left op agg) AND NULL` and an ANY to `(left op agg) OR NULL`, the same for `= ANY` and `<> ALL`, and a set of NULLs folds to NULL. Nine uncorrelated cases (`200 > ALL (10, 11, 12, NULL)` is UNKNOWN, `NOT (5 > ANY (...))` with a NULL is UNKNOWN, an empty ALL is TRUE, and so on) plus the correlated ALL over a user whose set holds a NULL.
3. Left operand of ALL and ANY: `has_subqueries` already answered true; `has_correlated_subqueries`, the classification's correlated detector and the WHERE walker now look at the left operand, and `process_where_subqueries` resolves it before the fold, so a correlated scalar subquery on the left counts and is bound per row. Test: `(SELECT MAX(x.amount) ... WHERE x.user_id = u.id) > ALL (SELECT amount FROM orders WHERE user_id = 20)`.
4. Aggregate FILTER: every detector walks a call's FILTER, so a subquery correlated only through one is evaluated per row instead of once. Its value is still wrong, and that is a pre-existing gap outside this PR: the executor substitutes outer references in a correlated subquery's WHERE only, not in its SELECT list, so `u.id` inside `COUNT(*) FILTER (WHERE x.user_id = u.id)` or inside `SUM(x.user_id * u.id)` binds to the inner row's `id` by unqualified name. Main returns 1, 1, 1 for the FILTER form and 6, 30, 72 for the SUM form (3, 12, 27 expected); the branch returns the same. It needs the outer row bound into the subquery's SELECT list before execution, which I would do in its own change with tests for both forms.

semihalev:
Both confirmed and fixed.

1. `has_outer_column_reference` and `references_outer_columns` walk the left operand of ALL and ANY, so `u.id * 10 > ALL (SELECT ...)` inside an EXISTS makes it correlated. Test: a subquery correlated only through that operand, on a plain table (10 users), on the join (30 rows) and grouped (10 groups); before the fix `u.id` bound to the inner row's `id` and every user passed.
2. `try_process_expression_subqueries` resolves the left operand of ALL and ANY before the fold and walks an aggregate's FILTER, the same shapes its detector reports. Test: `(SELECT 2) > ALL (SELECT 1)` and `(SELECT MAX(amount) FROM orders) > ALL (...)` in the SELECT list. A subquery inside an aggregate's FILTER or argument is resolved by the processor now, but the aggregation path never consults the processed SELECT list: every `parse_aggregations` call reads the statement's own columns, so `COUNT(*) FILTER (WHERE amount > (SELECT AVG(amount) FROM orders))` returns 0 and `SUM(CASE WHEN amount > (SELECT AVG(...)) THEN 1 ELSE 0 END)` returns NULL on main as on this branch. That is the same pre-existing gap as the outer references in a correlated subquery's SELECT list noted in the last round; both belong to one follow-up that binds the processed columns into the aggregation path, with tests for the FILTER, argument, and outer-reference forms.

semihalev:
All three confirmed and fixed.

1. FILTER: the statement's SELECT list has its subqueries resolved right after classification, before any of the six `parse_aggregations` sites read it, whenever it holds no correlated subquery. `SELECT COUNT(*) FILTER (WHERE (SELECT 1) = 1) FROM users` gives 30; `COUNT(*) FILTER (WHERE amount > (SELECT AVG(amount) FROM orders))` gives 45, plain and grouped; `SUM(CASE WHEN amount > (SELECT AVG(...)) THEN 1 ELSE 0 END)` gives 45 too, so the argument form from the last round is covered by the same change.
2. `process_correlated_expression` folds ALL and ANY under the row after resolving the left operand, and walks LIKE, lists and DISTINCT. Test: `(SELECT MAX(x.amount) ... WHERE x.user_id = u.id) > ALL (SELECT 0)` per user in the SELECT list; main returns 0 for every row. The FROM-less form in the example, `(SELECT u.id)`, fails before any of this on main as well: `SELECT u.id, (SELECT u.id) FROM users u` is a compile error there (a subquery without FROM does not receive the outer row), so the test uses a FROM-ful correlated subquery.
3. `try_process_expression_subqueries` walks IN (a subquery on the right folds as in a WHERE, a list on the right is walked), BETWEEN, LIKE, lists and DISTINCT, the containers its detector reports. Tests: `(SELECT 1) BETWEEN 0 AND 2`, `(SELECT 1) IN (1, 2)`, `1 IN ((SELECT 1), 2)`, `(SELECT 'ab') LIKE 'a%'`, and a correlated BETWEEN per row.

semihalev:
Aggregation, CTE and subquery rows of `examples/benchmark`, three interleaved runs against main at #91 on the same machine, median in us:

| Row | main (#91) | this PR |
|---|---|---|
| Aggregation (GROUP BY) | 45.3 | 45.4 |
| INNER JOIN | 12.3 | 12.0 |
| Scalar subquery | 8.3 | 8.2 |
| CTE + JOIN | 34.0 | 34.7 |
| Complex JOIN+GROUP+HAVING | 47.3 | 46.1 |
| COUNT DISTINCT | 0.3 | 0.3 |
| Multi aggregates (6) | 112.5 | 113.9 |
| Multiple CTEs (2) | 20.6 | 20.9 |
| Correlated in SELECT | 235.1 | 241.0 |
| GROUP BY (2 columns) | 146.7 | 147.1 |
| HAVING complex | 89.2 | 86.9 |

semihalev:
Aggregation, CTE and subquery rows of `examples/benchmark`, three interleaved runs against main at #91 on the same machine, median in us:

| Row | main (#91) | this PR |
|---|---|---|
| Aggregation (GROUP BY) | 46.1 | 45.0 |
| INNER JOIN | 12.1 | 12.1 |
| Scalar subquery | 8.2 | 8.3 |
| CTE + JOIN | 35.9 | 36.0 |
| Complex JOIN+GROUP+HAVING | 47.8 | 49.6 |
| COUNT DISTINCT | 0.3 | 0.3 |
| Multi aggregates (6) | 113.1 | 116.0 |
| Multiple CTEs (2) | 21.0 | 22.1 |
| Correlated in SELECT | 249.6 | 235.6 |
| GROUP BY (2 columns) | 149.0 | 148.0 |
| Derived table (FROM sub) | 392.4 | 420.9 |
| HAVING complex | 90.9 | 91.2 |

The Derived table row rechecked with five more interleaved runs, best: main 385.0, this PR 381.0; median: main 407.0, this PR 399.0.

semihalev:
Aggregation, window and subquery rows of `examples/benchmark`, eight interleaved runs against main at #91 on the same machine with a browser loading it, in us. Best-of is the column to read under that load.

| Row | main (#91) best | this PR best | main median | this PR median |
|---|---|---|---|---|
| Aggregation (GROUP BY) | 50.9 | 51.6 | 82.6 | 82.6 |
| Scalar subquery | 8.2 | 8.6 | 15.1 | 15.6 |
| CTE + JOIN | 38.8 | 33.2 | 62.9 | 64.9 |
| Window ROW_NUMBER | 268.8 | 263.2 | 437.9 | 443.3 |
| Window ROW_NUMBER (PK) | 7.1 | 8.3 | 10.2 | 10.9 |
| Window PARTITION BY | 10.8 | 9.3 | 15.0 | 15.3 |
| Multi window funcs (3) | 452.4 | 449.5 | 785.6 | 747.8 |
| Multi aggregates (6) | 122.5 | 126.0 | 213.2 | 208.0 |
| Correlated in SELECT | 278.1 | 278.7 | 441.9 | 406.9 |
| GROUP BY (2 columns) | 155.1 | 167.1 | 306.3 | 265.5 |
| Derived table (FROM sub) | 415.9 | 440.4 | 801.3 | 723.7 |
| Window ROWS frame | 394.1 | 423.3 | 723.3 | 708.8 |
| HAVING complex | 99.4 | 103.9 | 168.4 | 178.3 |

The PK window row was the only one over the noise band on best-of, so it got twelve more interleaved passes on its own: main median 7.39 us, this PR 7.64 us. Main's lower best is a single pass before the browser load began; under the same load the two sit within 0.3 us of each other. The window entry only walks the statement for a subquery before parsing the window functions, nothing else runs when there is none.

semihalev:
All four confirmed and fixed.

1. The two stages share one entry now, `execute_aggregation_then_windows`, so the window stage reads the statement the aggregate stage rewrote, and the aggregate lift leaves window functions alone. Test: `SELECT u.id % 2, SUM((SELECT MAX(x.amount) FROM orders x WHERE x.user_id = u.id)), ROW_NUMBER() OVER (ORDER BY u.id % 2) ... GROUP BY u.id % 2` gives 2430 and 2280, and a subquery reading a parent column the group rows do not carry gives 4710.
2. `WindowFunctionInfo` carries the FILTER and the call's own ORDER BY, and both are applied on every frame. A FILTER hands the aggregate a NULL for the rows it rejects, which every aggregate in `src/functions/aggregate` skips, so the frame walk stays exactly as it was. Tests: `COUNT(*) FILTER (WHERE (SELECT 0) = 1) OVER ()` gives 0, the running form over ids 1 to 6 gives 0,1,1,2,2,3, the partitioned form gives 1,2,1,2,1,2, `COUNT(DISTINCT city) FILTER (WHERE id > 3) OVER ()` gives 2, and `GROUP_CONCAT(name ORDER BY id DESC) OVER ()` gives user3,user2,user1.
3. The window stage lifts per input row: every SELECT expression beside a window function, the OVER clauses and `stmt.window_defs`. Tests: `ROW_NUMBER() OVER (ORDER BY (SELECT -u.id))` reverses the ranks, the same through a named window, `COUNT(*) OVER (PARTITION BY (SELECT u.id % 2))` gives 3 per parity. The indexed partition fetch is skipped while such a subquery is present, and the lifted columns stay out of SELECT *.
4. A lifted subquery's column takes the subquery's own text, so the result column keeps the name it had, two identical subqueries share one column, and a name already taken by an input column, bare or qualified through a join, gets a suffix. Test: a table whose column is named after the subquery text still sums to 303 on both sides of a join.

A FROM-less subquery now reads the parent row in its WHERE and its SELECT list, which is what `SUM((SELECT u.id)) OVER (PARTITION BY ...)` needs. `SELECT (SELECT u.id) FROM users u` stays a compile error, the same as on main.

Six new tests in `tests/where_subquery_containers_test.rs` cover these. All six fail on the previous head `aa61ba6c` and pass here.

semihalev:
Aggregation, window and subquery rows of `examples/benchmark`, sixteen interleaved runs against main at #91 on the same machine, median in us:

| Row | main (#91) | this PR | delta |
|---|---|---|---|
| Aggregation (GROUP BY) | 49.9 | 45.9 | -7.9% |
| Scalar subquery | 8.9 | 9.0 | +1.3% |
| CTE + JOIN | 36.8 | 37.0 | +0.5% |
| Window ROW_NUMBER | 250.1 | 250.2 | +0.1% |
| Window ROW_NUMBER (PK) | 6.5 | 6.2 | -5.3% |
| Window PARTITION BY | 9.2 | 9.3 | +0.7% |
| Multi window funcs (3) | 431.1 | 431.0 | -0.0% |
| Multi aggregates (6) | 120.8 | 122.0 | +1.0% |
| Correlated in SELECT | 250.2 | 250.8 | +0.2% |
| GROUP BY (2 columns) | 154.1 | 156.8 | +1.8% |
| Derived table (FROM sub) | 419.4 | 411.1 | -2.0% |
| Window ROWS frame | 389.5 | 399.1 | +2.5% |
| HAVING complex | 94.4 | 95.4 | +1.0% |

The `Window ROWS frame` row is the one that moved. Measured against the previous head `aa61ba6c` rather than main, that head is level with main and this one sits about 2% above it. The cost is not the FILTER, which the frame walk never sees: it comes from adding any code at all to `compute_aggregate_window_function`, a function of about 700 lines where the compiler's decisions swing on size. Splicing the old copy of that one function back in restores parity, and moving the setup out into two helpers took the gap from 3% to 2%. The rest of the rows are within the noise of this machine.

semihalev:
All three confirmed and fixed. The first needed one fix beyond the line you marked.

1. Parameters. The lift shared a column between two subqueries whose text matched, and an anonymous marker prints without the binding behind it, so the second read the first one's value. The lift now shares a column only when the text carries no marker, and two lifted columns never take one name.

   The same text is also the key of the scalar and IN subquery caches, so fixing the lift alone still gave the wrong answer. That one is not new here: on main today, `SELECT (SELECT MAX(v) FROM t WHERE t.id = ?) FROM t LIMIT 1` returns 10 for a binding of 1 and 10 again for a binding of 2, and the two-marker form in one statement returns the first value twice. The caches now key on the bindings as well as the text, and step aside when the text holds an anonymous marker. Test: two markers in one statement give 360 and 660, and two executions of one statement give 12 and 22.

2. Navigation control arguments. `Executor::window_control_argument` names the argument each of these reads off no row, the offset of LEAD and LAG, the group count of NTILE, the n of NTH_VALUE, and the window lift leaves it alone. Tests: `LAG(id, (SELECT 1))` gives the previous id, `NTILE((SELECT 2))` gives 1,1,2,2, `NTH_VALUE(id, (SELECT 2))` gives NULL,2,2,2, and a correlated control argument compiles and runs as it did before.

3. NULLS placement. `AggregateOrder` carries the direction and the placement, and `compare_sort_keys` reads both, so an explicit NULLS FIRST or NULLS LAST holds. This covers the plain aggregate path too, which dropped the placement the same way. `GROUP_CONCAT(v ORDER BY k NULLS FIRST)` over k of 2, NULL, 1 gives y,z,x and NULLS LAST gives z,x,y, both in a window and without one. `ORDER BY k DESC` with no NULLS clause still puts them first. The two aggregates that sorted their entries by hand now share the one comparator, and their private copies of it are gone.

Three new tests in `tests/where_subquery_containers_test.rs`. All three fail on `77bb39e1` and pass here.

One thing I left alone: `LAG(u.id, 1) OVER (ORDER BY u.id)` returns NULL for every row when the argument is qualified, on main as much as here. It has nothing to do with subqueries, so the test uses the unqualified form.

semihalev:
Aggregation, window and subquery rows of `examples/benchmark`, fourteen interleaved runs against main at #91 on the same machine, median in us:

| Row | main (#91) | this PR | delta |
|---|---|---|---|
| Aggregation (GROUP BY) | 47.1 | 46.4 | -1.3% |
| Scalar subquery | 8.6 | 8.8 | +2.7% |
| CTE + JOIN | 35.3 | 36.2 | +2.5% |
| Window ROW_NUMBER | 245.8 | 250.4 | +1.9% |
| Window ROW_NUMBER (PK) | 6.2 | 6.2 | +1.3% |
| Window PARTITION BY | 9.0 | 8.8 | -2.5% |
| Multi window funcs (3) | 431.9 | 421.7 | -2.4% |
| Multi aggregates (6) | 119.9 | 120.3 | +0.3% |
| Correlated in SELECT | 242.1 | 254.3 | +5.0% |
| GROUP BY (2 columns) | 153.0 | 150.4 | -1.7% |
| Derived table (FROM sub) | 416.5 | 420.9 | +1.1% |
| Window ROWS frame | 380.0 | 393.7 | +3.6% |
| HAVING complex | 92.7 | 92.1 | -0.7% |

The `Window ROWS frame` row still carries the couple of percent explained in the previous comment. Nothing in this round touches a hot loop: the subquery cache key grows only for a statement that binds parameters, and the aggregate sort keys are read once per window function.

semihalev:
One follow-up on the head above. Building the cache key turned the lookup into a nested match, and `execute_scalar_subquery` runs once per row of a correlated subquery, so the benchmark's `Correlated in SELECT` row lost 4% to the shape alone. The key and the lookup are two statements again and the row is level: measured against the previous head over eighteen interleaved runs, median +0.6%, pairwise -0.6%. Both suites stay green, 5091 and 5093.

semihalev:
Both confirmed and fixed.

1. The bindings go into the cache key in debug form, which carries the type a value prints without and escapes a separator a text could otherwise hold. Test: `SELECT (SELECT $1) IS NULL` gives true for a NULL binding and false for the text that spells it, on the same statement in the same thread.

2. `Executor::fold_window_control_arguments` runs right after `parse_window_functions`, on both paths that parse them, and resolves the control argument of each navigation function when the subquery there reads no row itself. Leaving the argument out of the lift was only half of it: the fold before parsing reached it only while nothing else in the SELECT list held that fold back. Test, with a correlated column beside it: `NTILE((SELECT 2))` gives 1,1,2,2, `LAG(id, (SELECT 1))` gives NULL,1,2,3, `NTH_VALUE(id, (SELECT 2))` gives NULL,2,2,2.

Two new tests. Both fail on `74a42088` and pass here.

semihalev:
Aggregation, window and subquery rows of `examples/benchmark`, sixteen interleaved runs against main at #91 on the same machine, median in us:

| Row | main (#91) | this PR | delta |
|---|---|---|---|
| Aggregation (GROUP BY) | 45.7 | 45.4 | -0.7% |
| Scalar subquery | 8.3 | 8.4 | +0.7% |
| CTE + JOIN | 34.0 | 33.6 | -1.3% |
| Window ROW_NUMBER | 238.2 | 237.8 | -0.2% |
| Window ROW_NUMBER (PK) | 5.8 | 5.9 | +0.7% |
| Window PARTITION BY | 8.6 | 8.3 | -3.3% |
| Multi window funcs (3) | 413.4 | 400.6 | -3.1% |
| Multi aggregates (6) | 113.7 | 117.2 | +3.0% |
| Correlated in SELECT | 237.5 | 239.6 | +0.9% |
| GROUP BY (2 columns) | 152.2 | 148.0 | -2.7% |
| Derived table (FROM sub) | 411.3 | 411.8 | +0.1% |
| Window ROWS frame | 368.1 | 375.5 | +2.0% |
| HAVING complex | 90.4 | 90.3 | -0.1% |

The `Multi aggregates` row read +3.0% here, so I gave it eighteen more interleaved runs on its own: median -0.1%, pairwise +1.0%. Nothing this round runs per row. The control arguments are folded once, right after the window functions are parsed, and the cache key is built only where it was already being built.

PR #94, opened 2026-09-05
fix: correlated subqueries, CTEs, views, joins, DML and expressions answer the way SQLite and DuckDB do

semihalev:
A correctness campaign driven by static review and by differential runs against SQLite 3.51 and DuckDB 1.2. Only a case where both engines agree and stoolap differs counted as a bug; 34 waves of about 1630 queries were run over correlated subqueries, joins, window functions, set operations, DML, expression evaluation, three-valued logic and layered views and CTEs. Every wrong answer found is fixed here, each with a test that fails on the head before its fix.

**Correlated subqueries**
- A subquery reads the parent row wherever it names it: the select list, an aggregate argument or FILTER, HAVING, CASE, and a HAVING that names the parent is classified as correlated. Names the subquery's own FROM defines win, and bare names still mean the inner table.
- The parent's columns are bound in scope before a derived table, view, CTE or join splits its WHERE, and the scope lists every leaf of a nested join.
- A correlated subquery over a view, a derived table or a materialized CTE is evaluated per row in the WHERE and in the select list, also when a sort column the select list leaves out rides along, and the column it is bound to never shadows a source column.
- Correlation is detected through a nested subquery, a derived table in the FROM, and an ORDER BY that reads the parent row; a subquery in ORDER BY is sorted by whether or not the select list carries its key.
- EXISTS answers what its subquery returns, and NOT EXISTS keeps its rows when the outer value is NULL; the counting shortcuts and the batch aggregate paths stay honest about what they count.
- The parent row is bound into the subqueries a WHERE holds, so a subquery two levels down that reads only the outermost row folds once instead of running for every row of the query holding it, and the plain conjuncts beside a correlated subquery are pushed down to storage. A three-level probe went from 102 seconds to 37 milliseconds.

**CTEs and views**
- Every reference is counted before a CTE is inlined, including set-operation branches, the select list, HAVING and ORDER BY, so a CTE named twice is materialized once and read from both places.
- The query-on-CTE-result path hands ORDER BY expressions, DISTINCT and set operations to execute_select, and a later CTE's subqueries read the earlier ones.
- A predicate pushed to a view on a join side is applied; a column of a view, derived table or CTE that the select list leaves out can still sort the result.

**Joins, set operations, windows**
- A join decides a match where it has the row in hand and asks the whole ON clause where an index answers its key; a join filter runs in its transaction.
- USING names its column by the bare name for every select list and keeps the qualified copy beside it.
- ORDER BY, LIMIT and OFFSET apply to the whole set operation.
- Windows: OVER (w ...) builds on a named window and prints what it adds; a window the select list does not carry can sort; the aggregates a window orders or partitions by are computed; NULL sort keys frame together; a window argument reads its column through its table.

**DML and DDL**
- DELETE removes what it matched on a table without a key; every key that can refuse a delete or an update is asked before anything changes, through cascades and with a correlated WHERE.
- INSERT ... ON CONFLICT DO UPDATE takes a WHERE, counts a skipped row as not affected, reports an error the condition raises, and runs in its transaction; WITH before UPDATE and DELETE; SET col = DEFAULT writes the column's default expression.
- A table-level PRIMARY KEY is a key, composite key columns are NOT NULL, a transaction can take again a key it deleted, and a committed unique entry is read against the row's latest local version.

**Expressions and functions**
- NOT over AND and OR follows three-valued logic in storage filters; a comparison against NULL stays unknown; x * 0 is not folded (NULL for a NULL x); an empty set matches nothing.
- IN over items that are not constants compiles to one InList op that judges the value once; an expression's name keeps the parentheses precedence and float grouping need, so SUM(a * (b % 3)) and SUM(a * b % 3) stay two aggregates; a timestamp, JSON or vector bound into an expression keeps its type.
- A volatile function such as RANDOM() is judged once, in memory, wherever it sits in the WHERE (CASE, BETWEEN, IN lists included), instead of once in storage and again in memory.
- Booleans count as one or nought in SUM and AVG on every path; SUBSTRING holds its window at the type's limits; searching an empty string finds nothing; TRIM, LTRIM and RTRIM take the characters to strip; a number cast from text ignores surrounding whitespace.

**Left open, as dialect or design**
NULLs first on DESC and last on ASC (PostgreSQL rule), backslash as an escape in literals, CAST('3.9' AS INTEGER) and float equality following SQLite, and the parser gaps UPDATE ... FROM, GROUPS BETWEEN, EXCLUDE, WITH inside a subquery, WITH RECURSIVE with UNION, several named windows in one WINDOW clause, and a correlated subquery inside JOIN ON. XOR is stoolap-only. A primary key UPDATE stays an explicit error by design.

**Test runs**
`.config/nextest.toml` adds two profiles: the default run leaves out the tests that judge wall-clock time and the two concurrency stress binaries, which turn red under load; `cargo nextest run -P full` runs everything.

**Verification**
Every new test fails on the commit before its fix. Both suites pass with the full profile (5248, and 5250 with test-filedb), clippy is clean, and all 34 waves rerun clean after the last commit in about two minutes; the remaining differences are the dialect items above. Subquery benchmarks against main sit within a point of parity.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,18 @@
+# `cargo nextest run` uses the default profile, which leaves out the tests
+# that judge wall-clock time and the concurrency stress binaries: they turn
+# red under load and cost minutes. `cargo nextest run -P full` runs every
+# test, for CI and before a merge.
+
+[profile.default]
+default-filter = """
+not (
+  test(=storage::mvcc::version_store::tests::test_unique_constraint_performance_bulk_update)
+  | test(=test_fk_cascade_delete_performance_with_auto_index)
+  | binary(=concurrency_history_test)
+  | binary(=volume_concurrency_test)
+)
+"""
+slow-timeout = { period = "60s" }
+
+[profile.full]
+default-filter = "all()"
--- a/src/core/value.rs
+++ b/src/core/value.rs
@@ -600,6 +600,7 @@ impl Value {
                     Value::Integer(v) => Value::Integer(*v),
                     Value::Float(v) => Value::Integer(*v as i64),
                     Value::Text(s) => s
+                        .trim()
                         .parse::<i64>()
                         .map(Value::Integer)
                         .unwrap_or(Value::Null(target_type)),
@@ -613,6 +614,7 @@ impl Value {
                     Value::Float(v) => Value::Float(*v),
                     Value::Integer(v) => Value::Float(*v as f64),
                     Value::Text(s) => s
+                        .trim()
                         .parse::<f64>()
                         .map(Value::Float)
                         .unwrap_or(Value::Null(target_type)),
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -6181,12 +6181,21 @@ impl Executor {
         // Parse aggregations
         let (aggregations, non_agg_columns) = self.parse_aggregations(stmt)?;
 
-        // Must have only aggregations plus the GROUP BY column (no other regular columns)
-        // Non-agg columns must be exactly the GROUP BY column
-        if non_agg_columns.len() > 1 {
+        // The rows come out as the GROUP BY column and then the aggregates,
+        // so the select list has to read exactly that, in that order
+        if non_agg_columns.len() != 1 || !non_agg_columns[0].eq_ignore_ascii_case(&group_col_name) {
             return Ok(None);
         }
-        if non_agg_columns.len() == 1 && !non_agg_columns[0].eq_ignore_ascii_case(&group_col_name) {
+        let leads_with_group_column = match stmt.columns.first() {
+            Some(crate::parser::ast::Expression::Identifier(id)) => {
+                id.value_lower.eq_ignore_ascii_case(&group_col_name)
+            }
+            Some(crate::parser::ast::Expression::QualifiedIdentifier(qid)) => {
+                qid.name.value_lower.eq_ignore_ascii_case(&group_col_name)
+            }
+            _ => false,
+        };
+        if !leads_with_group_column {
             return Ok(None);
         }
 
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -3071,18 +3071,17 @@ impl Executor {
                         }
                         SimpleAgg::Sum(sum_col_idx) | SimpleAgg::Avg(sum_col_idx) => {
                             if let Some(value) = row.get(*sum_col_idx) {
-                                match value {
-                                    Value::Integer(v) => {
-                                        state.agg_values[i] += *v as f64;
-                                        state.agg_has_value[i] = true;
-                                        state.counts[i] += 1;
-                                    }
-                                    Value::Float(v) => {
-                                        state.agg_values[i] += v;
-                                        state.agg_has_value[i] = true;
-                                        state.counts[i] += 1;
-                                    }
-                                    _ => {}
+                                // A boolean counts as one or nought
+                                let numeric = match value {
+                                    Value::Integer(v) => Some(*v as f64),
+                                    Value::Float(v) => Some(*v),
+                                    Value::Boolean(b) => Some(*b as i64 as f64),
+                                    _ => None,
+                                };
+                                if let Some(v) = numeric {
+                                    state.agg_values[i] += v;
+                                    state.agg_has_value[i] = true;
+                                    state.counts[i] += 1;
                                 }
                             }
                         }
@@ -3319,18 +3318,17 @@ impl Executor {
                         }
                         SimpleAgg::Sum(sum_col_idx) | SimpleAgg::Avg(sum_col_idx) => {
                             if let Some(value) = row.get(*sum_col_idx) {
-                                match value {
-                                    Value::Integer(v) => {
-                                        state.agg_values[i] += *v as f64;
-                                        state.agg_has_value[i] = true;
-                                        state.counts[i] += 1;
-                                    }
-                                    Value::Float(v) => {
-                                        state.agg_values[i] += v;
-                                        state.agg_has_value[i] = true;
-                                        state.counts[i] += 1;
-                                    }
-                                    _ => {}
+                                // A boolean counts as one or nought
+                                let numeric = match value {
+                                    Value::Integer(v) => Some(*v as f64),
+                                    Value::Float(v) => Some(*v),
+                                    Value::Boolean(b) => Some(*b as i64 as f64),
+                                    _ => None,
+                                };
+                                if let Some(v) = numeric {
+                                    state.agg_values[i] += v;
+                                    state.agg_has_value[i] = true;
+                                    state.counts[i] += 1;
                                 }
                             }
                         }
@@ -3546,18 +3544,17 @@ impl Executor {
                     }
                     SimpleAgg::Sum(sum_col_idx) | SimpleAgg::Avg(sum_col_idx) => {
                         if let Some(value) = row.get(*sum_col_idx) {
-                            match value {
-                                Value::Integer(v) => {
-                                    state.agg_values[i] += *v as f64;
-                                    state.agg_has_value[i] = true;
-                                    state.counts[i] += 1;
-                                }
-                                Value::Float(v) => {
-                                    state.agg_values[i] += v;
-                                    state.agg_has_value[i] = true;
-                                    state.counts[i] += 1;
-                                }
-                                _ => {}
+                            // A boolean counts as one or nought
+                            let numeric = match value {
+                                Value::Integer(v) => Some(*v as f64),
+                                Value::Float(v) => Some(*v),
+                                Value::Boolean(b) => Some(*b as i64 as f64),
+                                _ => None,
+                            };
+                            if let Some(v) = numeric {
+                                state.agg_values[i] += v;
+                                state.agg_has_value[i] = true;
+                                state.counts[i] += 1;
                             }
                         }
                     }
@@ -5007,6 +5004,15 @@ impl Executor {
                                 sum_int += i;
                             }
                         }
+                        // A boolean counts as one or nought
+                        Value::Boolean(b) => {
+                            has_value = true;
+                            if has_float {
+                                sum_float += *b as i64 as f64;
+                            } else {
+                                sum_int += *b as i64;
+                            }
+                        }
                         Value::Float(f) => {
                             has_value = true;
                             if !has_float {
@@ -5031,6 +5037,14 @@ impl Executor {
                             sum_float += *i as f64;
                         } else {
                             sum_int += i;
+                        }
+                    }
+                    Value::Boolean(b) => {
+                        has_value = true;
+                        if has_float {
+                            sum_float += *b as i64 as f64;
+                        } else {
+                            sum_int += *b as i64;
                         }
                     }
                     Value::Float(f) => {
@@ -5079,6 +5093,14 @@ impl Executor {
                                     sum_float += *i as f64;
                                 } else {
                                     sum_int += i;
+                                }
+                            }
+                            Value::Boolean(b) => {
+                                has_value = true;
+                                if has_float {
+                                    sum_float += *b as i64 as f64;
+                                } else {
+                                    sum_int += *b as i64;
                                 }
                             }
                             Value::Float(f) => {
@@ -5161,6 +5183,11 @@ impl Executor {
                         sum += f;
                         count += 1;
                     }
+                    // A boolean counts as one or nought
+                    Value::Boolean(b) => {
+                        sum += *b as i64 as f64;
+                        count += 1;
+                    }
                     _ => {}
                 }
             }
@@ -5195,6 +5222,10 @@ impl Executor {
                             }
                             Value::Float(f) => {
                                 sum += f;
+                                count += 1;
+                            }
+                            Value::Boolean(b) => {
+                                sum += *b as i64 as f64;
                                 count += 1;
                             }
                             _ => {}
@@ -5909,6 +5940,8 @@ impl Executor {
                         let num = match value {
                             Value::Integer(i) => *i as f64,
                             Value::Float(f) => *f,
+                            // A boolean counts as one or nought
+                            Value::Boolean(b) => *b as i64 as f64,
                             _ => continue,
                         };
                         state.sum += num;
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -6310,6 +6310,11 @@ impl Executor {
                                             state.agg_has_value[i] = true;
                                             state.counts[i] += 1;
                                         }
+                                        Value::Boolean(v) => {
+                                            state.agg_values[i] += *v as i64 as f64;
+                                            state.agg_has_value[i] = true;
+                                            state.counts[i] += 1;
+                                        }
                                         _ => {}
                                     }
                                 }
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -514,8 +514,27 @@ impl Executor {
                 (result_columns, result_rows)
             } else {
                 // Pre-process scalar subqueries in HAVING clause
-                // This executes subqueries like (SELECT AVG(a) FROM t) and replaces them with values
-                let processed_having = self.process_where_subqueries(having, ctx)?;
+                // This executes subqueries like (SELECT AVG(a) FROM t) and replaces them with values.
+                // One that reads the group is left as written; apply_having
+                // runs it once per group with that group as the outer row
+                let has_correlated_having = Self::has_correlated_subqueries(having);
+                let processed_having = if has_correlated_having {
+                    (**having).clone()
+                } else {
+                    self.process_where_subqueries(having, ctx)?
+                };
+                let outer_table: Option<String> =
+                    stmt.table_expr.as_ref().and_then(|te| match te.as_ref() {
+                        Expression::TableSource(source) => Some(
+                            source
+                                .alias
+                                .as_ref()
+                                .map(|a| a.value_lower.to_string())
+                                .unwrap_or_else(|| source.name.value_lower.to_string()),
+                        ),
+                        Expression::Aliased(aliased) => Some(aliased.alias.value_lower.to_string()),
+                        _ => None,
+                    });
 
                 // Build aggregate expression aliases for HAVING clause
                 // IMPORTANT: Include ALL aggregates, not just aliased ones,
@@ -548,6 +567,7 @@ impl Executor {
                     &result_columns,
                     &agg_aliases,
                     &expr_aliases,
+                    outer_table.as_deref(),
                     ctx,
                 )?;
 
@@ -4925,6 +4945,7 @@ impl Executor {
     }
 
     /// Apply HAVING clause to aggregated results
+    #[allow(clippy::too_many_arguments)]
     fn apply_having(
         &self,
         result: Box<dyn QueryResult>,
@@ -4932,6 +4953,7 @@ impl Executor {
         columns: &[String],
         agg_aliases: &[(String, usize)],
         expr_aliases: &[(String, usize)],
+        outer_table: Option<&str>,
         ctx: &ExecutionContext,
     ) -> Result<Box<dyn QueryResult>> {
         // Materialize the result
@@ -4946,6 +4968,50 @@ impl Executor {
         // Combine all aliases for HAVING clause evaluation
         let mut all_aliases: Vec<(String, usize)> = agg_aliases.to_vec();
         all_aliases.extend_from_slice(expr_aliases);
+
+        // A subquery that reads the group it is asked about is run once per
+        // group, with the group's row standing as the outer row, the way a
+        // correlated subquery in the WHERE clause is run once per row. The
+        // group's columns are offered under their bare names and under the
+        // table's, since the subquery may name either
+        if Self::has_correlated_subqueries(having) {
+            let columns_arc = CompactArc::new(columns.to_vec());
+            let keys: Vec<(CompactArc<str>, Option<CompactArc<str>>)> = columns
+                .iter()
+                .map(|c| {
+                    let bare = c.to_lowercase();
+                    let qualified = outer_table.filter(|_| !bare.contains('.')).map(|t| {
+                        CompactArc::from(format!("{}.{}", t.to_lowercase(), bare).as_str())
+                    });
+                    (CompactArc::from(bare.as_str()), qualified)
+                })
+                .collect();
+            let mut filtered_rows = RowVec::new();
+            let mut new_id = 0i64;
+            for (_, row) in rows {
+                let mut outer_row_map: FxHashMap<CompactArc<str>, Value> = FxHashMap::default();
+                for (i, (bare, qualified)) in keys.iter().enumerate() {
+                    let value = row.get(i).cloned().unwrap_or_else(Value::null_unknown);
+                    if let Some(q) = qualified {
+                        outer_row_map.insert(q.clone(), value.clone());
+                    }
+                    outer_row_map.insert(bare.clone(), value);
+                }
+                let correlated_ctx =
+                    ctx.with_outer_row(outer_row_map, CompactArc::clone(&columns_arc));
+                let processed = self.process_correlated_where(having, &correlated_ctx)?;
+                let filter = RowFilter::with_aliases(&processed, columns, &all_aliases)?
+                    .with_context(&correlated_ctx);
+                if filter.matches_checked(&row)? {
+                    filtered_rows.push((new_id, row));
+                    new_id += 1;
+                }
+            }
+            return Ok(Box::new(ExecutorResult::new(
+                columns.to_vec(),
+                filtered_rows,
+            )));
+        }
 
         // Create RowFilter with all aliases and context
         let having_filter =
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -916,7 +916,8 @@ impl Executor {
                     Expression::FunctionCall(func) if is_aggregate_function(&func.function) => {
                         self.get_aggregate_column_name(func).to_lowercase()
                     }
-                    _ => continue, // Can't easily compare, assume mismatch
+                    // Anything else matches only as a GROUP BY expression column
+                    _ => self.expression_to_string(col_expr).to_lowercase(),
                 };
                 if i >= agg_columns.len() || agg_columns[i].to_lowercase() != expected_name {
                     columns_match = false;
@@ -993,9 +994,15 @@ impl Executor {
                             )));
                         }
                     } else {
-                        // Non-aggregate function - evaluate it
+                        // Non-aggregate function: a GROUP BY expression column is
+                        // read as it is, anything else is evaluated
+                        let expr_lower = self.expression_to_string(col_expr).to_lowercase();
                         final_columns.push(format!("{}(...)", func.function));
-                        column_sources.push(make_source(col_expr, is_correlated));
+                        if agg_col_index_map.contains_key(&expr_lower) && !is_correlated {
+                            column_sources.push(ColumnSource::AggColumn(expr_lower));
+                        } else {
+                            column_sources.push(make_source(col_expr, is_correlated));
+                        }
                     }
                 }
                 Expression::Aliased(aliased) => {
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -1501,6 +1501,17 @@ impl Executor {
                 &mut non_agg_columns,
             )?;
         }
+        // A named window a function refers to with OVER w may partition or
+        // order by an aggregate as well; it is computed like ORDER BY's
+        for window in &stmt.window_defs {
+            for expr in window
+                .partition_by
+                .iter()
+                .chain(window.order_by.iter().map(|o| &o.expression))
+            {
+                self.extract_aggregates_from_expr(expr, &mut aggregations, &mut non_agg_columns)?;
+            }
+        }
         // Mark any new aggregates (from ORDER BY) as hidden, but only if they're truly new
         // Helper to create a signature string for an aggregate including its filter
         let make_sig = |agg: &SqlAggregateFunction| -> (String, String, bool, String) {
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -1615,6 +1615,19 @@ impl Executor {
             Expression::Cast(cast) => {
                 self.extract_aggregates_from_expr(&cast.expr, aggregations, non_agg_columns)?;
             }
+            // An aggregate a window function takes as its argument, or
+            // partitions or orders by, is computed like any other
+            Expression::Window(window) => {
+                for expr in window
+                    .function
+                    .arguments
+                    .iter()
+                    .chain(window.partition_by.iter())
+                    .chain(window.order_by.iter().map(|o| &o.expression))
+                {
+                    self.extract_aggregates_from_expr(expr, aggregations, non_agg_columns)?;
+                }
+            }
             _ => {}
         }
         Ok(())
--- a/src/executor/aggregation.rs
+++ b/src/executor/aggregation.rs
@@ -1927,55 +1927,10 @@ impl Executor {
         self.expression_to_string(target_expr)
     }
 
-    /// Convert an expression to a display string
-    #[allow(clippy::only_used_in_recursion)]
+    /// Convert an expression to a display string: the one stringifier, so
+    /// two aggregates over different expressions never share a name
     fn expression_to_string(&self, expr: &Expression) -> String {
-        match expr {
-            Expression::FunctionCall(func) => {
-                let args: Vec<String> = func
-                    .arguments
-                    .iter()
-                    .map(|a| self.expression_to_string(a))
-                    .collect();
-                format!("{}({})", func.function, args.join(", "))
-            }
-            Expression::Identifier(id) => id.value.to_string(),
-            Expression::QualifiedIdentifier(qid) => {
-                format!("{}.{}", qid.qualifier.value, qid.name.value)
-            }
-            Expression::StringLiteral(lit) => format!("'{}'", lit.value),
-            Expression::IntegerLiteral(lit) => lit.value.to_string(),
-            Expression::FloatLiteral(lit) => lit.value.to_string(),
-            Expression::BooleanLiteral(lit) => lit.value.to_string(),
-            Expression::Case(case) => {
-                // Use the Display implementation for CaseExpression
-                format!("{}", case)
-            }
-            Expression::Infix(infix) => {
-                format!(
-                    "{} {} {}",
-                    self.expression_to_string(&infix.left),
-                    infix.operator,
-                    self.expression_to_string(&infix.right)
-                )
-            }
-            Expression::Prefix(prefix) => {
-                format!(
-                    "{}{}",
-                    prefix.operator,
-                    self.expression_to_string(&prefix.right)
-                )
-            }
-            Expression::Cast(cast) => {
-                format!(
-                    "CAST({} AS {})",
-                    self.expression_to_string(&cast.expr),
-                    cast.type_name
-                )
-            }
-            // For any other expression type, use the Display trait if implemented
-            _ => format!("{}", expr),
-        }
+        super::utils::expression_to_string(expr)
     }
 
     /// Execute global aggregation (no GROUP BY)
--- a/src/executor/context.rs
+++ b/src/executor/context.rs
@@ -978,6 +978,16 @@ impl ExecutionContext {
         outer_row: FxHashMap<CompactArc<str>, Value>,
         outer_columns: CompactArc<Vec<String>>,
     ) -> Self {
+        // A subquery inside a subquery reads the row its grandparent sits
+        // on as well, with the nearer scope winning where both name a column
+        let outer_row = match &self.outer_row {
+            Some(grandparent) if !grandparent.is_empty() => {
+                let mut merged = grandparent.clone();
+                merged.extend(outer_row);
+                merged
+            }
+            _ => outer_row,
+        };
         Self {
             params: self.params.clone(),             // Arc clone = cheap
             named_params: self.named_params.clone(), // Arc clone = cheap
--- a/src/executor/cte.rs
+++ b/src/executor/cte.rs
@@ -114,7 +114,7 @@ fn expr_to_normalized_string(expr: &Expression) -> String {
 
 /// Hint for CTE optimization including LIMIT and optional ORDER BY pushdown
 #[derive(Clone)]
-struct CtePushdownHint {
+pub(crate) struct CtePushdownHint {
     limit: u64,
     order_by: Vec<OrderByExpression>,
 }
@@ -232,14 +232,27 @@ impl Executor {
         // This enables streaming GROUP BY early termination
         let cte_limit_hints = self.compute_cte_limit_hints(stmt, with_clause);
 
-        // Create CTE registry
+        let mut cte_registry = self.materialize_ctes(with_clause, ctx, Some(&cte_limit_hints))?;
+
+        // Execute the main query with CTE registry
+        self.execute_main_query_with_ctes(stmt, ctx, &mut cte_registry)
+    }
+
+    /// Run the CTEs of a WITH clause in order and store each result, so
+    /// the statement that follows, a SELECT or a DML, can read them
+    pub(crate) fn materialize_ctes(
+        &self,
+        with_clause: &WithClause,
+        ctx: &ExecutionContext,
+        cte_limit_hints: Option<&StringMap<CtePushdownHint>>,
+    ) -> Result<CteRegistry> {
         let mut cte_registry = CteRegistry::new();
 
         // Execute each CTE in order
         for cte in &with_clause.ctes {
             // Check if we have a pushdown hint for this CTE
             let cte_name_lower: String = cte.name.value_lower.to_string();
-            let pushdown_hint = cte_limit_hints.get(&cte_name_lower);
+            let pushdown_hint = cte_limit_hints.and_then(|hints| hints.get(&cte_name_lower));
 
             // Execute the CTE query (handles recursive CTEs)
             let (columns, rows) = if cte.is_recursive {
@@ -287,8 +300,7 @@ impl Executor {
             cte_registry.store(&cte.name.value, columns, rows);
         }
 
-        // Execute the main query with CTE registry
-        self.execute_main_query_with_ctes(stmt, ctx, &mut cte_registry)
+        Ok(cte_registry)
     }
 
     /// Execute a single CTE query
--- a/src/executor/cte.rs
+++ b/src/executor/cte.rs
@@ -1140,11 +1140,20 @@ impl Executor {
         // OPTIMIZATION: Get cached query classification to avoid repeated AST traversals
         let classification = get_classification(stmt);
 
+        let cte_alias = stmt
+            .table_expr
+            .as_deref()
+            .and_then(super::utils::get_table_alias_from_expr)
+            .map(|name| name.to_lowercase());
+
         // Apply WHERE clause filter
         let filtered_rows = if let Some(ref where_clause) = stmt.where_clause {
             // Process subqueries in WHERE clause (e.g., IN subqueries on CTEs)
-            // Use cached classification to avoid AST traversal
-            let processed_where = if classification.where_has_subqueries {
+            // Use cached classification to avoid AST traversal; a correlated
+            // one is evaluated per row below
+            let processed_where = if classification.where_has_subqueries
+                && !Self::has_correlated_subqueries(where_clause)
+            {
                 self.process_where_subqueries(where_clause, ctx)?
             } else {
                 (**where_clause).clone()
@@ -1154,14 +1163,7 @@ impl Executor {
             // CTE's own
             let processed_where = match ctx.outer_row() {
                 Some(outer) => {
-                    let inner: Vec<String> = stmt
-                        .table_expr
-                        .as_deref()
-                        .and_then(super::utils::get_table_alias_from_expr)
-                        .map(|name| name.to_lowercase())
-                        .into_iter()
-                        .collect();
-                    let inner: Vec<&str> = inner.iter().map(String::as_str).collect();
+                    let inner: Vec<&str> = cte_alias.iter().map(String::as_str).collect();
                     let scope = super::utils::InnerScope {
                         tables: &inner,
                         schema: None,
@@ -1175,19 +1177,29 @@ impl Executor {
                 None => processed_where,
             };
 
-            // Compile filter once and reuse for all rows
-            let mut eval =
-                ExpressionEval::compile(&processed_where, &cte_columns)?.with_context(ctx);
+            if Self::has_correlated_subqueries(&processed_where) {
+                self.filter_rows_by_correlated_where(
+                    &processed_where,
+                    cte_rows,
+                    &cte_columns,
+                    cte_alias.as_deref(),
+                    ctx,
+                )?
+            } else {
+                // Compile filter once and reuse for all rows
+                let mut eval =
+                    ExpressionEval::compile(&processed_where, &cte_columns)?.with_context(ctx);
 
-            let mut result = RowVec::new();
-            let mut row_id = 0i64;
-            for (_, row) in cte_rows {
-                if eval.eval_bool_checked(&row)? {
-                    result.push((row_id, row));
-                    row_id += 1;
+                let mut result = RowVec::new();
+                let mut row_id = 0i64;
+                for (_, row) in cte_rows {
+                    if eval.eval_bool_checked(&row)? {
+                        result.push((row_id, row));
+                        row_id += 1;
+                    }
                 }
+                result
             }
-            result
         } else {
             cte_rows
         };
@@ -1218,8 +1230,13 @@ impl Executor {
             return Ok((columns, rows, !skip_order_limit));
         }
 
-        // Process scalar subqueries in SELECT columns before projection
-        let processed_columns = self.try_process_select_subqueries(&stmt.columns, ctx)?;
+        // Process scalar subqueries in SELECT columns before projection; a
+        // correlated one is evaluated per row by the projection
+        let processed_columns = if Self::has_correlated_select_subqueries(&stmt.columns) {
+            None
+        } else {
+            self.try_process_select_subqueries(&stmt.columns, ctx)?
+        };
         let columns_to_use = processed_columns.as_ref().unwrap_or(&stmt.columns);
 
         // Determine output columns
@@ -1259,6 +1276,7 @@ impl Executor {
                         columns_to_use,
                         &filtered_rows,
                         &cte_columns,
+                        cte_alias.as_deref(),
                         ctx,
                     )?;
                     // Append source columns that aren't in output
@@ -1295,6 +1313,7 @@ impl Executor {
                     columns_to_use,
                     &filtered_rows,
                     &cte_columns,
+                    cte_alias.as_deref(),
                     ctx,
                 )?
             } else {
@@ -1314,7 +1333,13 @@ impl Executor {
                 // Sort on source columns first, then project
                 let sorted =
                     self.apply_order_by_to_rows(filtered_rows, &stmt.order_by, &cte_columns)?;
-                self.project_cte_rows_from_columns(columns_to_use, &sorted, &cte_columns, ctx)?
+                self.project_cte_rows_from_columns(
+                    columns_to_use,
+                    &sorted,
+                    &cte_columns,
+                    cte_alias.as_deref(),
+                    ctx,
+                )?
             } else {
                 self.apply_order_by_to_rows(filtered_rows, &stmt.order_by, &cte_columns)?
             }
@@ -1325,6 +1350,7 @@ impl Executor {
                     columns_to_use,
                     &filtered_rows,
                     &cte_columns,
+                    cte_alias.as_deref(),
                     ctx,
                 )?
             } else {
@@ -1461,9 +1487,22 @@ impl Executor {
         columns: &[Expression],
         rows: &RowVec,
         cte_columns: &[String],
+        table_alias: Option<&str>,
         ctx: &ExecutionContext,
     ) -> Result<RowVec> {
         use super::expression::{compile_expression, ExecuteContext, ExprVM, SharedProgram};
+
+        // A correlated subquery reads each row as its outer row
+        if Self::has_correlated_select_subqueries(columns) {
+            return self.project_rows_with_alias(
+                columns,
+                RowVec::from_vec((**rows).clone()),
+                cte_columns,
+                None,
+                ctx,
+                table_alias,
+            );
+        }
 
         let col_index_map = build_column_index_map(cte_columns);
 
--- a/src/executor/cte.rs
+++ b/src/executor/cte.rs
@@ -1950,19 +1950,8 @@ impl Executor {
             self.count_cte_references_in_expr(where_clause, &mut where_ref_counts);
         }
 
-        // A reference from the select list, HAVING, ORDER BY or another
-        // branch of a set operation reads the CTE by name as well
-        for expr in stmt
-            .columns
-            .iter()
-            .chain(stmt.having.as_deref())
-            .chain(stmt.order_by.iter().map(|o| &o.expression))
-        {
-            self.count_cte_references_in_expr(expr, &mut where_ref_counts);
-        }
-        for set_op in &stmt.set_operations {
-            self.count_cte_references_in_stmt(&set_op.right, &mut where_ref_counts);
-        }
+        // A reference from any other clause reads the CTE by name as well
+        self.count_cte_references_outside_from(stmt, &mut where_ref_counts);
 
         // Only inline CTEs that:
         // 1. Are used exactly once in table expressions (FROM/JOIN)
@@ -2077,12 +2066,27 @@ impl Executor {
             self.count_cte_references_in_expr(where_clause, ref_counts);
         }
 
-        // Check SELECT columns, HAVING and ORDER BY for subqueries
+        self.count_cte_references_outside_from(stmt, ref_counts);
+    }
+
+    /// Count CTE references in the clauses other than FROM and WHERE: the
+    /// select list, HAVING, ORDER BY, named windows and set-operation branches
+    fn count_cte_references_outside_from(
+        &self,
+        stmt: &SelectStatement,
+        ref_counts: &mut StringMap<usize>,
+    ) {
+        let windows = stmt.window_defs.iter().flat_map(|w| {
+            w.partition_by
+                .iter()
+                .chain(w.order_by.iter().map(|o| &o.expression))
+        });
         for expr in stmt
             .columns
             .iter()
             .chain(stmt.having.as_deref())
             .chain(stmt.order_by.iter().map(|o| &o.expression))
+            .chain(windows)
         {
             self.count_cte_references_in_expr(expr, ref_counts);
         }
@@ -2175,6 +2179,18 @@ impl Executor {
             }
             Expression::Aliased(aliased) => {
                 self.count_cte_references_in_expr(&aliased.expression, ref_counts);
+            }
+            Expression::Window(window) => {
+                for expr in window
+                    .function
+                    .arguments
+                    .iter()
+                    .chain(window.function.filter.as_deref())
+                    .chain(window.partition_by.iter())
+                    .chain(window.order_by.iter().map(|o| &o.expression))
+                {
+                    self.count_cte_references_in_expr(expr, ref_counts);
+                }
             }
             _ => {}
         }
--- a/src/executor/cte.rs
+++ b/src/executor/cte.rs
@@ -1149,6 +1149,31 @@ impl Executor {
             } else {
                 (**where_clause).clone()
             };
+            // A parent row's column named through its own table is bound
+            // first, or the bare name it would fall back to could be the
+            // CTE's own
+            let processed_where = match ctx.outer_row() {
+                Some(outer) => {
+                    let inner: Vec<String> = stmt
+                        .table_expr
+                        .as_deref()
+                        .and_then(super::utils::get_table_alias_from_expr)
+                        .map(|name| name.to_lowercase())
+                        .into_iter()
+                        .collect();
+                    let inner: Vec<&str> = inner.iter().map(String::as_str).collect();
+                    let scope = super::utils::InnerScope {
+                        tables: &inner,
+                        schema: None,
+                    };
+                    super::utils::substitute_outer_references_in_scope(
+                        &processed_where,
+                        outer,
+                        &scope,
+                    )
+                }
+                None => processed_where,
+            };
 
             // Compile filter once and reuse for all rows
             let mut eval =
--- a/src/executor/cte.rs
+++ b/src/executor/cte.rs
@@ -312,17 +312,19 @@ impl Executor {
     ) -> Result<(Vec<String>, RowVec)> {
         // Check if the CTE references another CTE
         if let Some(ref table_expr) = stmt.table_expr {
-            // First check for simple CTE reference
-            // BUT: if there are set operations (UNION, etc.), we need to fall through
-            // to execute_select which handles them properly
-            if stmt.set_operations.is_empty() {
+            // First check for simple CTE reference, when the CTE path can
+            // finish the statement on its own; otherwise fall through to
+            // execute_select, which handles set operations and the rest
+            if Self::cte_result_path_finishes(stmt) {
                 if let Some(cte_name) = self.extract_cte_name_for_lookup(table_expr) {
                     if let Some((columns, rows)) = cte_registry.get(&cte_name) {
                         // Execute query against CTE result
                         // Clone Arc data into RowVec for processing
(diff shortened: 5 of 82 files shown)


semihalev:
All three confirmed, and a fourth turned up beside them. Every shape below now matches SQLite 3.51 on the same data.

1. The binding sat at the top of `execute_scalar_subquery`, above paths that read the statement as written, one of which keys a cache on the statement's address. A borrowed stack slot is reused by the next call, so `(SELECT u.id FROM orders x LIMIT 1)` beside a `SUM` was answered with the SUM. The binding moved down to where the general path runs. Test: the two subqueries side by side give 3,1 then 12,2 then 27,3 then 48,4.

2. Fixing the address alone was not enough, because the unbound statement went through the batch aggregate too and was wrong there on main already. One pass over the table answers every parent row at once, which holds only while the correlation predicate is the one place the parent row is read. `SUM(x.user_id * u.id) ... WHERE x.user_id = u.id` reads it inside the aggregate, so the pass carried the first parent's value into every row: main answers 6,30,72,132 and this branch answered 3,6,9,12, against SQLite's 3,12,27,48. The batch path now steps aside when the aggregate, the HAVING or the predicate beside the correlation reads the parent row, and the answers are 3,12,27,48 and 36,69,102,135.

3. `substitute_outer_references_inner` has an `Aliased` arm, so `(SELECT u.id AS parent_id FROM orders x LIMIT 1)` gives 1,2,3,4 and keeps its alias.

4. Found while reading finding 2: the batch aggregate's key named the table, the correlation column and the function name, never the aggregate's own argument, so two sums over one table and one correlation column shared an entry.

   ```sql
   SELECT u.id, (SELECT SUM(x.amount) FROM orders x WHERE x.user_id = u.id),
                (SELECT SUM(x.price)  FROM orders x WHERE x.user_id = u.id) FROM users u;
   -- main and this branch before the fix: 33|33, sqlite: 33|303
   ```

   This one is not new here, it is on main today. The key names the aggregate now.

Four more tests, ten in the file. Nine of the ten fail on `8e76d426`.

Both suites stay green, 5106 and 5108. Benchmark rows for subqueries were taken while the machine was busy, so the numbers are soft: correlated in SELECT reads about two percent above main and the rest sit at parity. Nothing per row changed for that query, its batch lookup is untouched, so I read that as noise. I can re-measure on a quiet machine if you want it pinned down.

semihalev:
Confirmed, and it went two steps further than the report.

1. The check only stopped the cache being read. `try_execute_and_cache_batch_aggregate` decides for itself whether it can run, so it still ran, and the lookup that came back empty afterwards answered a COUNT with zero. Both paths ask `Executor::aggregate_reads_parent_row` now.

2. The index COUNT shortcut is a second short way, and it was wrong on more than the parent row. It counts the rows an index points at, which answers a COUNT only when the count is of rows. On `8e76d426`, with an index on the correlation column:

   ```sql
   SELECT u.id, (SELECT COUNT(*) FILTER (WHERE x.amount > u.id * 10) FROM orders x WHERE x.user_id = u.id) FROM users u;   -- 3,3,3,3, sqlite 2,2,2,2
   SELECT u.id, (SELECT COUNT(DISTINCT x.user_id) FROM orders x WHERE x.user_id = u.id) FROM users u;                      -- 3,3,3,3, sqlite 1,1,1,1
   SELECT u.id, (SELECT COUNT(x.amount) FROM orders x WHERE x.user_id = u.id) FROM users u;                                -- counts a NULL amount
   ```

   The probe takes `COUNT(*)` alone: no FILTER, no DISTINCT, no column argument.

3. That still left the filtered count wrong with an index, and the trail led somewhere else entirely. `extract_aggregations_simple` reads neither the FILTER nor the DISTINCT of what it extracts, so the streaming path behind a grouped aggregate counts whole columns. It has nothing to do with subqueries:

   ```sql
   SELECT user_id, COUNT(*) FILTER (WHERE amount % 10 > 0) FROM orders GROUP BY user_id;  -- 3 per group, sqlite 2
   SELECT user_id, COUNT(DISTINCT user_id) FROM orders GROUP BY user_id;                  -- 3 per group, sqlite 1
   ```

   Writing the columns through the table hides it, since a qualified argument leaves that path with nothing to extract, which is why the correlated form only showed it once an index was there. This one is on main today. The path steps aside when an aggregate carries a FILTER or a DISTINCT.

Three more tests, thirteen in the file, each shape run with and without an index. Eleven of the thirteen fail on `8e76d426`.

Clippy clean, both suites green at 5109 and 5111. Benchmark, eighteen interleaved runs against current main: aggregation, GROUP BY, HAVING, scalar subquery, EXISTS, NOT EXISTS, correlated in SELECT and derived table all read at or below main.

semihalev:
Four more wrong answers, and the way I went looking for them.

I put the same queries through SQLite 3.51 and DuckDB 1.2 on identical data and kept only what those two agree on, so no single engine decides what is right. Two sets, 103 queries, covering projection and ORDER BY, aggregates, NULL semantics, types and casts, joins, set operations, subqueries, window functions and CTEs. Before this commit stoolap parted company with both engines on five of them, now on none. Four differences remain where SQLite and DuckDB part company with each other, and stoolap sits with SQLite on all four: `7 / 2`, `CAST(2.7 AS INTEGER)`, and the ANY and ALL syntax SQLite does not take.

1. **An outer join with a predicate beside its key.** Every pair the ON condition rejected became a row of the preserved side, so a left row that matched once and turned one candidate away came out twice.

   ```sql
   SELECT COUNT(*) FROM t LEFT JOIN u ON u.t_id = t.id AND u.w > 150;  -- 7, both engines 6
   SELECT COUNT(*) FROM t LEFT JOIN u ON u.t_id = t.id AND 1 = 0;      -- 7, both engines 6
   ```

   `apply_residual_post_join` keeps a row of the preserved side only where nothing matched, and one however many candidates it turned away. LEFT, RIGHT and FULL read the same way.

2. **A CASE that compares an operand.** The operand stayed on the stack, so the comparison around the CASE read the operand against the CASE's own result.

   ```sql
   SELECT COUNT(*) FROM t WHERE id > CASE 1 WHEN 1 THEN 2 ELSE 3 END;  -- 0, both engines 4
   ```

   It asked whether 1 exceeds 2. Written as `CASE ... < id` it was right, which is why it went unseen. The compiler emitted a `Dup` that `CaseCompare` does not need, since it reads the operand without taking it.

3. **A subquery two levels in.** It never saw the row its grandparent sits on.

   ```sql
   SELECT u.id, (SELECT COUNT(*) FROM users v
                 WHERE EXISTS (SELECT 1 FROM orders x
                               WHERE x.user_id = v.id AND x.amount > u.id * 10)) FROM users u;
   -- 0,0,0,0, both engines 4,3,2,1
   ```

   Two causes. `references_outer_columns` stopped at a nested subquery, so the query in the middle counted as reading nothing outside itself, was called uncorrelated and ran once for the whole parent. And `expression_references_outer_tables`, which decides whether an EXISTS may become a semi join, turned away only the names it already knew as outer, so a grandparent's name passed the guard and was then read off no row at all. The walk now goes through a nested EXISTS, scalar subquery and quantified comparison; the semi join leaves any name its own table does not own to a row at a time; and a subquery's context carries the rows above it, the nearer one winning where both name a column.

4. **Sorting by a column the SELECT leaves out.** The scan path named such a column only when it was written bare, so a qualified one rode out with the rows while the column list stayed as it was: `SELECT u.city FROM users u ORDER BY u.id` declared one column and handed back two values a row. The join path already did this correctly and its arm moved over.

Four test files, 25 tests: the join's ON condition, the CASE operand, the sort key, and the grandparent's row. 22 of them fail on `8e76d426`.

semihalev:
A second differential wave, over string functions, NULL semantics, window frames, set operations and CTEs. Same rule as before: SQLite 3.51 and DuckDB 1.2 on identical data, and only what the two of them agree on counts. Three more wrong answers came out of it, plus three questions I would rather you decide than decide myself.

1. **Sorting by an expression the SELECT list leaves out did nothing at all.**

   ```sql
   SELECT id FROM t ORDER BY ABS(f);   -- table order, both engines sort by the value
   SELECT id FROM t ORDER BY n * -1;   -- same
   ```

   The projection keeps a column for a sort key written as a name, and evaluates one for a DISTINCT ON that has to be computed, but an ORDER BY expression got neither, so the sort had nothing to read and left the rows alone. Such an expression is evaluated into a column named after it, which the sort finds the way it finds any other and which is trimmed with the rest. A subquery keeps its own path, since it is read against the parent row.

   While I was there: the two callers that name what the projection appends now take those names from the projection itself. That duplication is where the sort key riding out with the rows came from.

2. **SUBSTRING opened its window from the wrong place.** A start before the string was pulled to its first character rather than counted from there, a negative start never reached back from the end, and a negative length returned nothing rather than the characters before the start. Eleven boundaries now read the same as both engines, `SUBSTR('abcdef', -3, 2)` gives `de` and `SUBSTR('abcdef', 3, -2)` gives `ab`.

3. **LIKE read the character after its escape as itself only where that character was a wildcard.** Anywhere else the escape stayed in the pattern and was matched literally, so `LIKE '%!,%' ESCAPE '!'` looked for an exclamation mark. Escaping the escape did not work either.

Three left, and each is a decision rather than a defect:

- **Where NULLs go in a descending sort.** SQLite and DuckDB both put them last; stoolap puts them first. The three engines hold three positions: SQLite treats NULL as the smallest value, DuckDB puts NULLs last whichever way you sort, PostgreSQL treats NULL as the largest. stoolap reads as PostgreSQL does, and consistently, including the aggregate ORDER BY default added earlier in this PR. I would leave it: changing it moves rows for every query that sorts a nullable column.
- **`UNION` in a recursive CTE**, which answers `only supports UNION ALL`. A missing feature that says so out loud rather than a wrong answer. Happy to add the deduplication if you want it here.
- **A backslash in a string literal.** stoolap reads it as an escape: `LENGTH('a\\b')` is 3 and `'\'` will not parse. Both engines, and the standard, treat it as an ordinary character: 4 and 1. Fixing it is a lexer change that would alter what existing queries mean, so I would rather you called it.

Both waves together are 157 queries. stoolap now parts company with both engines on none of them, apart from the NULL ordering above and the two above it. Nine dialect splits remain where the two engines part company with each other, and stoolap sits with SQLite on all nine.

semihalev:
Head is now `951867da`. All three findings are answered, and chasing the first two turned up a fourth join bug plus four more from another round of engine comparison.

## The three findings

**1. Outer join losing rows that hold the same values.** Confirmed with your shape: `l` holding two 1s, `r` holding `(1, 0)`, `ON l.k = r.k AND r.v > 1` counted 1 instead of 2.

**2. FULL JOIN adding a row of nothing but NULLs.** Confirmed. `l = (1)`, `r = (2, 1)`, `ON l.k = r.k AND r.v > 0` returned three rows, the third all NULL.

**3. Alias shifting the hidden ORDER BY column.** Already closed by `af7d3119`. `SELECT x.g AS y, x.id FROM t x ORDER BY x.g, x.n` gives the same rows as the form without the alias.

I could not repair the post-pass for the first two. Two identical left rows each turning away n candidates leave 2n rejected output rows, and neither the left row count nor the boundaries between them survive into the output: the row id there is a counter handed out after the join. The pair has to be judged where the join still has both rows in hand.

So the hash join and the merge join now take the rest of the ON clause and test each candidate pair as they form it. A rejected pair is not a match, so it is neither emitted nor counted when either side decides which of its rows still need a NULL to pair with. `apply_residual_post_join` is gone. INNER joins keep filtering the output, where the question never arises and no candidate has to be materialized.

## A third join bug this uncovered

The merge join never applied the rest of the ON clause at all: `execute_merge_join` passed an empty filter list with a comment saying so. It is reachable, the planner picks it when both inputs are sorted on the keys and there are 500+ rows between them. On 800 sorted rows:

```sql
SELECT COUNT(*) FROM a JOIN b ON a.k = b.k AND b.v = 2   -- 400, should be 100
```

A plain INNER JOIN, silently returning every equal-key pair. Same fix closes it.

The parallel hash join has no way to reject a pair, so an OUTER join carrying such a condition now takes the streaming operator.

## Four more from another comparison round

You said to run the CLIs more, so I built two more waves against SQLite 3.51 and DuckDB 1.2, 118 queries: one on join ON conditions, one on window frames, set operations, NULL semantics and subquery placement. Where both engines agree and we differ:

**`WHERE n <= m` kept the rows where both are NULL**, and so did `>=`. Two NULLs order equal so an index can hold them side by side, and that answer reached the comparison operators.

**`WHERE NULL = NULL` kept every row, and `WHERE n = n` kept the NULL rows too.** The simplifier folded `x = x` to TRUE and `x - x` to 0 for any deterministic expression. That holds only where the expression has a value. There was an SLT case pinning the old answer with a comment calling it non-standard; it now expects 0.

**A subquery in the WHERE clause threw away what the query asked of the rows it kept.** Three fast paths answer such a query by projecting the rows they found and returning them, and they project a plain SELECT list. `SELECT COUNT(*) FROM s WHERE NOT EXISTS (...)` failed with "Function not found: COUNT", GROUP BY and HAVING were dropped, a window function was refused, and an `ORDER BY` naming a column the SELECT list leaves out was silently ignored. Each path now takes only the queries it can answer.

**A RANGE offset against a NULL sort key framed to the whole partition.** Nothing measures against a NULL, and the bound fell back to the partition edge, so the row saw every other row. It frames to its peers now, which both engines agree on.

## Where the waves stand

| wave | queries | both engines agree | we differ | dialect |
|---|---|---|---|---|
| 1 | 78 | 67 | 1 | 10 |
| 2 | 79 | 71 | 2 | 6 |
| 3 (joins) | 48 | 37 | 1 | 10 |
| 4 (frames, sets, NULLs) | 70 | 57 | 2 | 11 |

Every remaining difference is one of the open items below. All 37 dialect cases are the NULL ordering question, and in each one we match DuckDB exactly.

## Open, for you to decide

**NULL ordering.** We put NULLs last on ASC and first on DESC, which is PostgreSQL's rule read consistently. SQLite treats NULL as smallest, DuckDB puts them last both ways. This accounts for every dialect difference in all four waves, and it is the only reason `ORDER BY n DESC` differs. I have not touched it.

**Recursive CTE with UNION.** `WITH RECURSIVE r(x) AS (SELECT 1 UNION SELECT x + 1 FROM r WHERE x < 4)` gives "Recursive CTE only supports UNION ALL". Both engines answer 10. A missing feature rather than a wrong answer.

**Two window syntaxes we do not parse.** `WINDOW w1 AS (...), w2 AS (...)` fails on the second definition, and `GROUPS BETWEEN ... ` is not recognised at all. Also missing features.

**`MIN(3, 1, 2)` returns 3.** SQLite reads multi-argument MIN and MAX as scalar least and greatest and answers 1; DuckDB rejects the call. We silently ignore every argument past the first, which neither engine would do. Making it behave like SQLite means teaching five places that recognise MIN and MAX by name to look at the argument count, so I left it for you to pick between that and rejecting the call.

## Cost

The suite could not be measured today; a control run with the same binary in three slots showed a 50% swing by position alone. A dedicated probe on the shapes I changed is stable, best of five runs, 20k rows a side, interleaved against main and against `af7d3119`:

| shape | main | af7d3119 | now |
|---|---|---|---|
| INNER | 3.0 | 3.0 | 3.1 |
| INNER, condition beside the key | 4.1 | 4.1 | 4.2 |
| LEFT | 3.5 | 3.5 | 3.5 |
| LEFT, condition beside the key | 6.1 | 7.8 | 5.8 |
| FULL, condition beside the key | 5.0 | 13.0 | 5.8 |

`2e0de373` had made a FULL JOIN with such a condition 2.6x slower; this takes it back and puts LEFT below main. The 0.8ms FULL JOIN still owes to main is the price of the answer being right. The two queries the suite flagged as regressions, `IN list (7 values)` and `OR conditions (3 vals)`, measure identical on a dedicated probe: OR 15.0 / 15.0 / 14.8 and IN 5.67 / 5.68 / 5.69 microseconds for main, `af7d3119` and now.

Gates green: fmt, clippy with `-D warnings`, 5139 tests, 5141 with `test-filedb`. The eleven new tests all fail on `af7d3119` and pass here.

semihalev:
Head is `de1238bb`. All three confirmed and fixed, and the answers now match what SQLite 3.51 and DuckDB 1.2 both give.

**1. Alias inside an ORDER BY expression.** `SELECT n AS x FROM t ORDER BY x + 0` returned table order. The expression was evaluated against the source row, where the alias is not there yet, so every key came out NULL.

Rather than resolve the alias early I left such an expression to the sort, which reads the projected row and has the name. That was the path before the hidden-column work and it is the one that already handles this. It costs the fast path for these keys only, and it keeps the case both engines agree on: inside an expression a source column wins over an alias sharing its name.

```sql
SELECT n AS id FROM t ORDER BY id + 0   -- reads the table's id
SELECT n AS id FROM t ORDER BY id       -- a bare name reads the alias
```

**2. A function of a compound argument not finding its column.** `ORDER BY ABS(n - 20)` returned table order. The projection had built the column and named it after the whole call; the lookup asked for `ABS(expr)`, the name an aggregate over a compound argument would carry, and missed. It now falls back to the whole-call name.

`ABS(n)` worked before because the aggregate name and the call name agree when the argument is a bare column, which is why this only showed with something like `n - 20` inside.

**3. SUBSTRING overflow.** `SUBSTR('abc', 1, 9223372036854775807)` panicked in debug and returned an empty string in release. The sums that build the window now saturate, and the clamp that follows turns a saturated bound into the end of the string it ran past. All six extremes agree with SQLite:

```
SUBSTR('abc', 1, 9223372036854775807)                      -> abc
SUBSTR('abc', 9223372036854775807, 2)                      -> ''
SUBSTR('abc', 3, -9223372036854775808)                     -> ab
SUBSTR('abc', -2, 9223372036854775807)                     -> bc
SUBSTR('abc', -9223372036854775808, 3)                     -> ''
SUBSTR('abc', -9223372036854775807, 9223372036854775807)   -> abc
```

DuckDB rejects offsets and lengths past 2^32 rather than answering, so SQLite is the only engine that gives a value for these.

Three new tests, all failing on `951867da`. Gates green: fmt, clippy with `-D warnings`, 5143 tests, 5145 with `test-filedb`. The four comparison waves are unchanged at 67/78, 71/79, 37/48 and 57/70, with every remaining difference still one of the four open items above.

semihalev:
Head is `f9004ca2`. Both findings confirmed and fixed, and two more waves of engine comparison turned up three further bugs, one of them the worst in the PR so far.

## The two findings

**1. A key with an alias spoiling the other keys.** Confirmed: `SELECT n AS x FROM t ORDER BY ABS(n - 20), x + 0` gave `10,20,30,40` where both engines give `20,10,30,40`. Deferring one key sent them all down the same road, and the key that already had a column no longer found what it was computed from. Each key is now read from its column where it has one, and only the rest are compiled. The correlated path takes the same route.

**2. Two names read as one.** Confirmed: `ORDER BY INSTR(g, 'a'), INSTR(g, 'A'), id` gave `1,3,2` against `3,1,2`. Identifiers are case-insensitive but a string literal is not, so those are two expressions with two columns and the first was answering for both. An exact name now wins, with the case-insensitive match kept for the identifiers it was there for.

## Two more waves, three more bugs

Wave 5 covers DML, constraints and type coercion, 44 cases each bringing its own tables; wave 6 covers subquery scoping, set operations, joins, grouping and window functions, 53 queries.

**A DELETE removing nothing on a table without a primary key.** The worst one. The scanning path names each matching row by its primary key; a table without one left every match unnamed and the statement reported no rows removed, with no error:

```sql
CREATE TABLE a (id INTEGER, n INTEGER);   -- no primary key
DELETE FROM a WHERE id IN (SELECT id FROM b);   -- removed nothing
```

That path is reached by any condition the storage layer cannot answer alone, so a subquery, an expression over two columns, RETURNING, or a table with children all hit it. The same predicate as a SELECT returned the row, and UPDATE with it worked, which is why it stayed hidden. The rows are now named by the ids the scan read them under.

**A boolean column standing alone kept nothing.** `WHERE b` on a BOOLEAN column matched no rows, while `WHERE NOT b`, `WHERE b = TRUE` and `CASE WHEN b` were all right. The single-op filter paths answer a NULL as false, which is what a filter wants, and they answered an op they do not know the same way; a bare column load is not one of the comparisons they implement. They now take only the ops they implement.

**EXISTS ignoring what its subquery does with the rows it reads.** An EXISTS over one table is answered by a semi-join built from the table and the correlation in its WHERE, and the rest of the subquery was dropped:

```sql
SELECT COUNT(*) FROM d
WHERE EXISTS (SELECT 1 FROM e WHERE e.d_id = d.id GROUP BY e.d_id HAVING COUNT(*) > 1)
```

gave 3 instead of 1, and `HAVING COUNT(*) > 99` also gave 3 instead of 0. A bare aggregate went the other way: it returns its one row whether or not anything matched, so EXISTS over it should always hold, and we said it never did. A subquery carrying a HAVING, a row count, a set operation, or an aggregate without a GROUP BY is now left to the general path. Grouping alone still takes the semi-join, since a group is there exactly where a row is.

## Where the waves stand

| wave | queries | both engines agree | we differ | dialect |
|---|---|---|---|---|
| 1 | 78 | 67 | 1 | 10 |
| 2 | 79 | 71 | 2 | 6 |
| 3 joins | 48 | 37 | 1 | 10 |
| 4 frames, sets, NULLs | 70 | 57 | 2 | 11 |
| 5 DML, constraints, coercion | 44 | 42 | 0 | 2 |
| 6 scoping, sets, joins, grouping | 53 | 47 | 1 | 5 |

372 queries. Every remaining difference is one of the open items below, and all 44 dialect cases are the NULL ordering question, where we match DuckDB exactly in each one.

## Still open, for you

Unchanged from before: the NULL ordering rule (we read PostgreSQL's consistently), `WITH RECURSIVE` over UNION rather than UNION ALL, `WINDOW w1 AS (...), w2 AS (...)` failing on the second definition, `GROUPS BETWEEN` not parsed, and `MIN(3, 1, 2)` returning 3 where SQLite answers 1 and DuckDB rejects the call.

Fourteen new tests, all failing on `de1238bb`. Gates green: fmt, clippy with `-D warnings`, 5155 tests, 5157 with `test-filedb`.

semihalev:
Head is `92a01af3`. Both findings fixed, and three more waves of comparison found five more bugs, four of them fixed here.

## The two findings

**1. A DELETE skipping the keys that point at it.** Confirmed, and it was mine. The rows a keyless table's DELETE removes go into the row-id list, and the foreign key step reads the other one, so it had nothing to work on. Under RESTRICT a referenced parent row was removed; CASCADE and SET NULL never ran. The matched rows are now kept the way RETURNING keeps them, and each key reads the column it names from them.

Worth noting the shape this was hiding in: before the keyless DELETE worked at all, that same list stayed empty, so nothing was deleted and nothing was enforced. The bug only became reachable once the delete did.

**2. EXISTS answered a second way.** Confirmed. The guards went on the semi-join and the index probe kept its own road, which did not carry them. With an index on the inner column, `EXISTS (SELECT 1 FROM e WHERE e.d_id = d.id LIMIT 0)` was true again. The check now lives in one place and both roads take it, and the test runs each case twice, with and without the index.

## Three more waves

Wave 7 covers indexes, transactions and views (36 cases), wave 8 strings, numeric edges and deep nesting (60), wave 9 CTEs, correlation in every position, windows and set operations (50).

Waves 7 and 8 came back almost clean. One bug: `REPLACE('abc', '', 'x')` returned `xaxbxcx`, since an empty needle matches between every character; there is nothing there to find, so the subject is returned as it stands.

Wave 9 found four, all of them also on `main`:

**DISTINCT reading past its own projection.** `SELECT DISTINCT g FROM t ORDER BY g LIMIT 2` returned `a, a`. With ORDER BY and LIMIT the projection is deferred past the sort, which left DISTINCT hashing whole source rows, where rows that project alike are still telling apart. DISTINCT ON was already excluded from that; DISTINCT is now too.

**A subquery in the ON clause never read.** `JOIN r ON r.p_id = p.id AND r.v > (SELECT AVG(v) FROM r)` returned nothing, because the comparison against an unevaluated subquery is unknown for every pair. It is read once now, the way the WHERE clause reads one.

**A LEFT JOIN against a subquery losing its left rows.** A join whose right side is a subquery is turned around so an index on the left can be probed. Only an INNER join reads the same that way; a LEFT join went on preserving the side it was given, which after the turn was the subquery. `p LEFT JOIN (SELECT ...) s` returned only the matched rows, and `SELECT *` gave the columns in the turned-around order.

**ORDER BY over a window function, still open.** `SELECT id FROM p ORDER BY ROW_NUMBER() OVER (ORDER BY n DESC), id` ignores the key and returns table order. The sort cannot compile a window function on its own, so every key comes out NULL. Computing it would mean giving the sort a hidden column the window executor fills, which is a bigger piece of work than the rest of this PR and I would rather not start it here. Both engines answer it.

## One more I am reporting rather than fixing

The turn-around above also reorders the output columns, and for an INNER join that survives: `SELECT * FROM p JOIN (SELECT ...) s ON ...` returns `s`'s columns first where both engines return `p`'s. The values match their headers, so a client reading by name is fine and one reading by position is not. The path builds its column list in physical order on purpose, with a comment saying projections find columns by name, and `SELECT *` is the case that does not. Fixing it means expanding the star in the query's order rather than the physical one, on a hot path, which I would rather do deliberately than at the end of this PR.

## Where the waves stand

| wave | queries | both engines agree | we differ | dialect |
|---|---|---|---|---|
| 1 | 78 | 67 | 1 | 10 |
| 2 | 79 | 71 | 2 | 6 |
| 3 joins | 48 | 37 | 1 | 10 |
| 4 frames, sets, NULLs | 70 | 57 | 2 | 11 |
| 5 DML, constraints, coercion | 44 | 42 | 0 | 2 |
| 6 scoping, sets, joins, grouping | 53 | 47 | 1 | 5 |
| 7 indexes, transactions, views | 36 | 33 | 0 | 3 |
| 8 strings, numbers, nesting | 60 | 56 | 0 | 4 |
| 9 CTEs, correlation, windows | 50 | 44 | 6 | 0 |

518 queries. Of the thirteen remaining differences, ten are the NULL ordering rule showing up in a new place each time (inside a LIMIT, inside a window's ORDER BY, inside a correlated sort key), two are `WITH` inside a subquery and `WINDOW w1, w2`, and one is the window sort key above.

## Still open, for you

The NULL ordering rule, `WITH RECURSIVE` over UNION, `WINDOW w1 AS (...), w2 AS (...)`, `GROUPS BETWEEN`, `MIN(3, 1, 2)`, and now `WITH` inside a subquery, ORDER BY over a window function, and the star's column order after a turned-around join.

Nine new tests, all failing on `f9004ca2`. Gates green: fmt, clippy with `-D warnings`, 5164 tests, 5166 with `test-filedb`.

semihalev:
Head is `6e171d63`. Two more waves, one more bug.

## Wave 10: the same questions at volume

The merge join bug earlier in this PR only appeared past 500 rows, so this wave runs 52 queries over 4,000 and 12,000 rows: enough to cross the merge join threshold, the parallel hash join threshold at 10,000 build rows, and whatever the top-N and streaming aggregation paths switch on. Joins of every shape, aggregates, windows, DISTINCT, subqueries, and filters chosen to pick different plans.

49 of 52 agreed with both engines and the three differences are all the NULL ordering rule. Nothing new: the size-dependent paths answer the same as the small ones.

## Wave 11: several tables at once

45 queries over four tables: three and four table chains, INNER and LEFT and FULL mixed in one query, joins on a text column, residuals at every level, type mixing, and NULL in every subquery position.

One bug, on `main` as well as here:

**An empty set was answering as unknown rather than as no match.** A NULL asked whether it is in a set is unknown, because it might be the one that matches. A set holding nothing has no such one, so `IN` over it is false and `NOT IN` is true whatever it is asked about. That case was reaching the NULL answer first:

```sql
SELECT COUNT(*) FROM a WHERE k NOT IN (SELECT v FROM b WHERE v > 9999)
```

left out every row where `k` is NULL. A set that holds a NULL still leaves the answer unknown, which is the case just below it and is unchanged.

## Where the waves stand

| wave | queries | both engines agree | we differ | dialect |
|---|---|---|---|---|
| 1 | 78 | 67 | 1 | 10 |
| 2 | 79 | 71 | 2 | 6 |
| 3 joins | 48 | 37 | 1 | 10 |
| 4 frames, sets, NULLs | 70 | 57 | 2 | 11 |
| 5 DML, constraints, coercion | 44 | 42 | 0 | 2 |
| 6 scoping, sets, joins, grouping | 53 | 47 | 1 | 5 |
| 7 indexes, transactions, views | 36 | 33 | 0 | 3 |
| 8 strings, numbers, nesting | 60 | 56 | 0 | 4 |
| 9 CTEs, correlation, windows | 50 | 44 | 6 | 0 |
| 10 the same at 16,000 rows | 52 | 49 | 1 | 2 |
| 11 four tables, mixed join types | 45 | 40 | 0 | 5 |

615 queries. Fourteen differences remain: eleven are the NULL ordering rule turning up in one more place, two are syntax we do not parse, and one is ORDER BY over a window function.

## Still open, for you

Unchanged: the NULL ordering rule, `WITH RECURSIVE` over UNION, `WINDOW w1 AS (...), w2 AS (...)`, `GROUPS BETWEEN`, `MIN(3, 1, 2)`, `WITH` inside a subquery, ORDER BY over a window function, and the star's column order after a turned-around join.

One new test, failing on `92a01af3`. Gates green: fmt, clippy with `-D warnings`, 5165 tests, 5167 with `test-filedb`.

semihalev:
Head is `59316c22`. Two of the three confirmed and fixed; the third I could not reproduce and I want to be straight about what I found.

**2. The EXISTS check closing the correlated aggregate paths.** Confirmed, and it was mine. `extract_index_nested_loop_info` is shared with the scalar COUNT probe and the batch aggregate paths, and an aggregate without a GROUP BY is the very shape those are for, so my check turned all of them away. It moved to the EXISTS probe, beside the semi-join that already carries it. The EXISTS tests still pass with and without an index, and a correlated count now takes its own path again.

**3. The ON subquery skipped where an index answers the key.** Confirmed, and wider than reported: the index probe takes one equality out of the ON clause as its key and **the rest was dropped whatever it was**, not only a subquery.

```sql
CREATE INDEX rp ON r (p_id);
SELECT p.id FROM p JOIN r ON r.p_id = p.id AND r.v > 300;   -- returned every pair the key matched
```

That is on `main` too. The whole ON clause is now asked of each pair the probe returns. The operator already tests its filter as the pair is formed and only then counts the outer row as matched, so this is safe for a LEFT join: a pair it turns away leaves the outer row to be NULL-padded rather than dropped. Reading the subqueries in the clause also moved to the top of the join, ahead of the index path, which returns before the place I had put it.

**1. The partial delete on a foreign key refusal. I could not reproduce it.** I tried four shapes on `6e171d63` and on `main`: one child table with a CASCADE column and a RESTRICT column, in both declaration orders, and two child tables one of each, in both creation orders, against a keyed and a keyless parent. All eight refused cleanly with every child row still there. The order the constraints come back in was already putting the refusing ones first, so the cascade never ran.

The reasoning holds even though the shape does not: nothing in the code was making that ordering true. So it is now the code that does. Both paths ask every key that can refuse before the first change, the keyed one inside `enforce_delete_actions_iter` and the keyless one by taking the refusing keys first. The test is a guard rather than a reproducer, and it passes on the old head too, which I would rather say than leave you to find.

If you have a shape that does reproduce it, send it and I will check the fix covers that one as well.

Gates green: fmt, clippy with `-D warnings`, 5167 tests, 5169 with `test-filedb`. Two new tests; the index join one fails on `6e171d63`.

semihalev:
Head is `f95f9366`. Both findings confirmed and fixed, and wave 12 found two more, one of them a crash.

## The two findings

**1. A RESTRICT beneath a CASCADE.** Confirmed, and worse than my previous round claimed. The check I had borrowed from UPDATE reads the action a key takes *on update*, which for a key that cascades on delete is still RESTRICT. So it refused a plain cascade on a keyless parent (an existing test broke) and my chain test "passed" for that wrong reason rather than because the chain was checked. Your scenario is real: with `parent → child (CASCADE) → grandchild (RESTRICT)` and two parents, the first parent's child was gone before the second parent's grandchild refused.

There is now a delete-flavoured walk. It follows the same road the cascade takes, collecting the child rows that would go by primary key and asking every key pointing at them, and it runs in full for every value before anything is written, in both the keyed and the keyless path. A metadata pass skips it when the tree holds nothing that can refuse a delete. The test now runs inside `BEGIN`, deletes two parents, and reads the child table before `COMMIT`; it fails on `59316c22`.

**2. The transaction in the ON clause.** Confirmed: `JoinFilter` carried parameters but not the transaction, so `CURRENT_TRANSACTION_ID()` read as NULL inside an ON clause on the index path. It carries it now, the way `RowFilter` does. The test runs with and without the index inside a transaction.

## Wave 12: DML at volume

48 cases over 3,000 rows each, every case with its own tables, across four shapes: keyed and keyless, indexed and not. DELETE and UPDATE with expressions, subqueries, EXISTS and NOT EXISTS, on and around an indexed column, followed by a SELECT of what was left. Two bugs, both on `main`:

**The process died.** The members of an IN set are sorted to name the program that reads them, and the plain sort reads `PartialOrd`, whose NULL is neither below nor above anything, while `1 < 5`. Once the set was large enough for the sort to split it, the standard library found the order broken and panicked:

```sql
DELETE FROM t WHERE k IN (SELECT k FROM t WHERE v = 5)   -- 3,000 rows, a NULL among the keys
```

The members are now ordered through `Ord`, which puts NULL first and is total. This is the only place I found sorting `Value`s through the default order, but several others sort through `partial_cmp(..).unwrap_or(Equal)`, which is non-total in the same way if a NULL reaches them (`statistics.rs:1750` and `:1774`, `parallel.rs:1145`, `cte.rs:1584`, `query.rs:251` and `:8859`). I did not touch them; whether their inputs can hold a NULL I could not tell from the call sites, so I would rather you look at them than have me guess.

**NOT EXISTS over a NULL outer value.** `NOT EXISTS (SELECT 1 FROM r WHERE r.k = t.k)` is true where `t.k` is NULL, since no inner row can equal it. The rewrite that answers it through a set read it as `t.k NOT IN (...)`, which is unknown there, so SELECT left those rows out and DELETE left them in place. The negated set is now asked without its NULLs, and an outer NULL is kept alongside, as `(NOT IN ...) OR (t.k IS NULL)`. No new field on the node and nothing in the VM changed. I had thought this was DELETE-only at first; it was not, and wave 11 missed it because its NULL was on the inner side.

## Where the waves stand

| wave | queries | both engines agree | we differ | dialect |
|---|---|---|---|---|
| 1 | 78 | 67 | 1 | 10 |
| 2 | 79 | 71 | 2 | 6 |
| 3 joins | 48 | 37 | 1 | 10 |
| 4 frames, sets, NULLs | 70 | 57 | 2 | 11 |
| 5 DML, constraints, coercion | 44 | 42 | 0 | 2 |
| 6 scoping, sets, joins, grouping | 53 | 47 | 1 | 5 |
| 7 indexes, transactions, views | 36 | 33 | 0 | 3 |
| 8 strings, numbers, nesting | 60 | 56 | 0 | 4 |
| 9 CTEs, correlation, windows | 50 | 44 | 6 | 0 |
| 10 the same at 16,000 rows | 52 | 49 | 1 | 2 |
| 11 four tables, mixed join types | 45 | 40 | 0 | 5 |
| 12 DML at 3,000 rows, keyed and not, indexed and not | 48 | 48 | 0 | 0 |

663 queries. The fourteen remaining differences are unchanged: eleven are the NULL ordering rule, two are syntax we do not parse, one is ORDER BY over a window function.

Gates green: fmt, clippy with `-D warnings`, 5171 tests, 5173 with `test-filedb`. Four new tests, all failing on `59316c22`.

semihalev:
Head is `3272b76b`. The P1 confirmed and fixed, in both places it lived, and wave 13 found one more bug with a long tail.

## The P1: keys beneath a cascade through a UNIQUE column

Confirmed, and it was in two places. Both the walk that asks the keys before a delete acts and `cascade_delete_recursive` itself read the child rows about to go by primary key alone, and where the child had none they stepped over everything beneath it: the parent went, its child went with it, and a grandchild restricting on the child's UNIQUE column was left pointing at nothing.

Each key beneath the child is now read through the column it names. The child rows are collected once per grandchild key, by that column's value, both before anything is written and again while the cascade runs. This also fixes a case I had not looked at: a grandchild pointing at a UNIQUE column of a *keyed* child was wrong the same way, since the old code handed it the child's primary key value regardless of which column it referenced. The test covers both shapes (`child(ccode UNIQUE, ...)` and `child(id PRIMARY KEY, ccode UNIQUE, ...)`) and fails on `f95f9366`.

## Wave 13: windows at volume, conditional aggregation, three-valued logic

50 queries: twenty window shapes over 2,000 rows folded to order-free results, twelve conditional aggregates, and eighteen truth-table cases over every pairing of TRUE, FALSE and NULL in AND, OR, NOT, IS, WHERE, HAVING and CASE. The truth tables all agree with both engines. One bug:

**SUM and AVG over a boolean answered NULL.** `SUM(v > 100)`, `SUM(b)` on a BOOLEAN column and `AVG(b)` all added nothing and returned NULL, where both engines count the rows the condition holds for. A silent NULL is the one answer no reading gives: SQLite and DuckDB count, PostgreSQL refuses the call. I went with counting, since that is what both references do and it is what `SUM(CASE WHEN ... THEN 1 ELSE 0 END)` already returned.

The fix has a tail worth knowing about. The numeric reading SUM and AVG take of a value is written in **ten places**, one per path a query can take: the two aggregate functions, the compiled aggregator, the streaming aggregator in `query.rs`, the three `SimpleAgg` loops, `fast_sum_column` and `fast_avg_column` with their parallel halves, the extraction in `execute_global_aggregation`, the hot store's `version_store.sum_column`, and the sealed volume's default contribution and column stats. I found them one at a time by watching which shape still answered NULL after each fix: the expression form, then the column form, then the column form with a WHERE. Each now reads a boolean as one or nought. Gathering them into one helper is its own change and I have left it for one; but it is the reason a bug like this can hide in nine places after being fixed in the tenth.

## Where the waves stand

| wave | queries | both engines agree | we differ | dialect |
|---|---|---|---|---|
| 1 | 78 | 67 | 1 | 10 |
| 2 | 79 | 71 | 2 | 6 |
| 3 joins | 48 | 37 | 1 | 10 |
| 4 frames, sets, NULLs | 70 | 57 | 2 | 11 |
| 5 DML, constraints, coercion | 44 | 42 | 0 | 2 |
| 6 scoping, sets, joins, grouping | 53 | 47 | 1 | 5 |
| 7 indexes, transactions, views | 36 | 33 | 0 | 3 |
| 8 strings, numbers, nesting | 60 | 56 | 0 | 4 |
| 9 CTEs, correlation, windows | 50 | 44 | 6 | 0 |
| 10 the same at 16,000 rows | 52 | 49 | 1 | 2 |
| 11 four tables, mixed join types | 45 | 40 | 0 | 5 |
| 12 DML at 3,000 rows, keyed and not, indexed and not | 48 | 48 | 0 | 0 |
| 13 windows at volume, conditional aggregates, truth tables | 50 | 49 | 0 | 1 |

713 queries. The fourteen remaining differences are unchanged: eleven the NULL ordering rule, two syntax we do not parse, one ORDER BY over a window function.

Gates green: fmt, clippy with `-D warnings`, 5175 tests, 5177 with `test-filedb`. Four new tests, all failing on `f95f9366`.

semihalev:
Follow-up on the last review round and on the next three differential waves.

**Review P2s on 3272b76b (all four addressed)**

- `CompiledAggregate::accumulate_sum` now counts a boolean, so `SUM(DISTINCT b)` no longer comes back NULL.
- The version store column scan, the volume grouped accumulators (row and columnar, filtered and unfiltered) and the streaming text-key GROUP BY count booleans too.
- Volume stats sealed before the change carry zeros for boolean columns and cannot be told apart from stats sealed since, so `sum_column` bypasses the stats shortcut for boolean columns and scans instead.
- Coverage: `SUM(DISTINCT)`, plus a 20k-row grouped and filtered sum with a delete on top to force the tombstone path.

**Wave 14 (42 queries, now 42/42)**

- HAVING with a correlated subquery was processed once, unbound, so every group passed EXISTS and none passed NOT EXISTS. Each group is now the outer row of the subquery; the streaming grouped path steps aside for such a HAVING.
- An UPDATE whose correlated WHERE the semi-join rewrite refused (NOT EXISTS over two conjuncts) matched nothing. The still-correlated WHERE is evaluated per row in both the plain and the correlated SET paths.
- Found while writing the HAVING test: with as many select columns as grouped columns, a non-aggregate function or literal beside the group key was assumed to match and the raw key came back, so `COALESCE(a, -1)` gave NULL for the NULL group. A select column now only matches a GROUP BY expression column by its text; anything else is evaluated.

**Wave 15 (49 queries: windows with frames, ORDER BY with LIMIT and OFFSET, string functions, CTEs, USING joins, DML driven by a window; now 49/49)**

- JOIN ... USING renamed the join column to its bare name only under `SELECT *`, so with an explicit select list `SUM(a)` counted rows, `MAX(a)` gave 1 and `GROUP BY a` made one NULL group. The rename now applies to every select list, and `a`, `x.a` and `y.a` all resolve.
- A correlated subquery in ORDER BY read the projected row, so `SELECT id ... ORDER BY (SELECT COUNT(*) FROM y WHERE y.a = x.a)` sorted by nothing. Every base column now rides along until the sort has run. An uncorrelated subquery in ORDER BY compiled to a placeholder that could not run; it is run once up front.
- The classifier's outer-reference walk did not descend into nested subqueries, so `WHERE b > (SELECT AVG(b) FROM y WHERE y.a = (SELECT MIN(a) FROM y WHERE y.s = x.s))` counted as uncorrelated and was evaluated once for every row. It now mirrors the executor's walk.

**Wave 16 (45 queries: three-valued logic, NULL in IN and BETWEEN and CASE, FILTER aggregates, multi-column GROUP BY with HAVING on expressions, nested derived tables, set operations inside subqueries, CTE chains, VALUES sources, DML fed by subqueries)**

43/45 agree across the three engines; the remaining two are `VALUES ... AS t(v)`, which SQLite does not parse, and there stoolap matches DuckDB. No divergence to fix.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 13 | 713 | see earlier comments | |
| 14 | 42 | 42 | HAVING correlation, UPDATE refused semi-join, group key projection |
| 15 | 49 | 49 | USING column, ORDER BY subquery, nested outer reference |
| 16 | 45 | 43 (+2 SQLite dialect) | none needed |

Every new test fails on the commit before its fix and passes after. fmt, clippy and both nextest suites are green.

semihalev:
Waves 16 to 19 are in, plus the fixes they turned up.

**Wave 17 (46 queries, joins at volume: chains of outer joins, WHERE that turns an outer join inner, RIGHT and FULL joins, ON with expressions, OR and subqueries, self joins, GROUP BY, HAVING, DISTINCT and windows over a join)**

45/46. The one difference is `ORDER BY y.b DESC` with NULLs, where stoolap puts NULLs first on DESC (the PostgreSQL rule) and both engines put them last. That is the open dialect decision, not a bug, so I left it.

**Wave 18 (56 statements: indexes on the join and filter columns so the index-driven paths run the shapes earlier waves ran by scan, views over joins and aggregates, BEGIN, ROLLBACK and COMMIT around DML)**

- ORDER BY on a column of a view, a derived table or a CTE that the select list leaves out was silently dropped: the source projected the select list before the sort ran. Both paths now carry such a column along until the sort has run, as a base table and a join already did.

Now 56/56.

**Wave 19 (46 queries: aggregates and windows over empty sets, HAVING without GROUP BY, LIMIT 0 and OFFSET past the end, subqueries that yield nothing, unicode text functions, CAST, rounding, negative modulo)**

- `SELECT COUNT(*) FROM x HAVING COUNT(*) > n` with no WHERE and no GROUP BY took the row-count shortcut, which never looked at the HAVING. The MIN/MAX index shortcut had the same hole. Both step aside when there is a HAVING.
- `SELECT COUNT(*) FROM (subquery) GROUP BY s` over a text key returned the keys in place of the counts: the streaming derived-table path put out key then aggregates whatever the select list asked for. It now runs only when the select list reads exactly that.
- The simplifier folded `x * 0` to 0, so `a * 0 IS NOT NULL` counted the NULL rows too. The fold is gone, like `x - x` before it.
- `TRIM(s, chars)`, `LTRIM` and `RTRIM` with a second argument were rejected; they now strip the named characters.
- `CAST(' 7' AS INTEGER)` came back NULL; padded text is trimmed before it is parsed.

The remaining differences in this wave are dialect: SQLite has no ALL/ANY, and SQLite and DuckDB disagree between themselves on `CAST('3.9' AS INTEGER)`, `CAST('1e2' AS INTEGER)` and float equality, with stoolap siding with SQLite on the casts and on the comparison.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 15 | 804 | see earlier comments | |
| 16 | 45 | 43 (+2 SQLite dialect) | none needed |
| 17 | 46 | 45 (+1 NULL order rule) | none needed |
| 18 | 56 | 56 | ORDER BY on a hidden derived or view column |
| 19 | 46 | 40 (+6 dialect) | HAVING without GROUP BY, derived text-key counts, x * 0, TRIM chars, CAST padding |

Every new test fails on the commit before its fix and passes after. fmt, clippy and both nextest suites are green.

semihalev:
Wave 20 (53 statements: CREATE TABLE AS SELECT, DEFAULT values, ON CONFLICT upserts, RETURNING on all three DML forms, SET from CASE and subqueries, ALTER TABLE ADD COLUMN, savepoints).

- `INSERT ... ON CONFLICT (k) DO UPDATE SET ... WHERE cond` was a parse error. It is now parsed, compiled with EXCLUDED as the second row source like the SET expressions, and read against the row met; where it does not hold the row is left alone. a8ff3661.
- Everything else in the wave agrees, with two exceptions that are not fixes: DuckDB has no SAVEPOINT (dialect), and `UPDATE u SET v = k, k = k + 100` updates the primary key, which stoolap refuses with an explicit error ("cannot UPDATE primary key column"). Both engines do it. That is the PK remap limitation we decided needs a design first, so I am leaving it and noting it here; the three later differences in the wave are just that row's absence.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 19 | 997 | see earlier comments | |
| 20 | 53 | 45 (+5 dialect, +3 from the PK update limitation) | DO UPDATE ... WHERE |

semihalev:
Wave 21 (40 queries, windows in depth: RANGE and GROUPS frames, EXCLUDE, named windows, NTH_VALUE, CUME_DIST, PERCENT_RANK, ORDER BY over a window result, windows over GROUP BY, several windows in one select).

- `ORDER BY ROW_NUMBER() OVER (...)` with no such column in the select list was silently ignored: the classifier saw no window function, so the plain sort evaluated the key to NULL. A window in ORDER BY now counts as a window query; the window path computes each such key into a column aliased to its own text, the sort reads it by name and the projection drops it, beside a star included. b96c2d70.
- `OVER (w ROWS BETWEEN ...)` and `OVER (w ORDER BY ...)` were parse errors; only bare `OVER w` referred to a named window. The parenthesised form now builds on the named window with its own ORDER BY and frame. 5d7b4d5a.

Left open, as parser gaps rather than wrong results: `GROUPS BETWEEN ...` frames and the `EXCLUDE CURRENT ROW / TIES / GROUP` frame clause. Both need the frame engine to learn peer-group arithmetic, which is a feature of its own; four queries in this wave fail to parse for that reason and nothing else does.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 20 | 1050 | see earlier comments | |
| 21 | 40 | 35 (+1 SQLite dialect, +4 GROUPS/EXCLUDE gap) | ORDER BY over a window, OVER (w ...) |

semihalev:
Review of 5d7b4d5a, all seven findings addressed, plus wave 22.

**P1, correlated UPDATE skipping ON UPDATE RESTRICT.** Confirmed: `UPDATE p SET code = 2 WHERE EXISTS (SELECT 1 FROM q WHERE q.k = p.code LIMIT 1)` went through and left the child pointing at a code that no longer existed. The FK pre-check evaluated the still-correlated WHERE unbound, saw false for every row and checked nothing. It now binds each row the way the update does. Test: the update is refused and the codes stay put, while a row nothing references still changes.

**P2s**

- Boolean volume scan: the arm for `ColumnData::Boolean` is in the scan loop now. I could not reproduce the NULL on disk: 20k boolean rows written, reopened, summed, grouped, deleted from, all correct, so the scan may not have been the path that shape takes. The arm is right regardless and costs nothing.
- USING and `t.*`: the join column is no longer renamed in place. An explicit select list gets a column of its own for the bare name beside the qualified ones, so `a`, `ux.a` and `ux.*` all see it. Test covers `ux.*` and `uy.*, a`.
- Named-window extensions: `WindowExpression::Display` now writes `OVER (w ORDER BY ... frame)` when the reference carries anything of its own, so two extensions of one window are two expressions. Test sorts by the reverse numbering while selecting the forward one.
- Hidden ORDER BY windows are appended before the subqueries in windows are lifted, so `ORDER BY RANK() OVER (ORDER BY (SELECT -t.id))` reads its subquery. Test.
- A `DO UPDATE ... WHERE` that leaves the row alone is no longer an affected row: `apply_on_duplicate_update` reports whether the update took, and the callers count only those. `WHERE FALSE` gives 0, `WHERE u.v < 50` over one hit and one insert gives 2.
- The condition sees the transaction: the INSERT paths are dispatched before the dispatcher puts the transaction id in the context, so `execute_insert` now takes it from the active transaction itself. That also fixes `SET v = CURRENT_TRANSACTION_ID()` in DO UPDATE, which set NULL inside BEGIN.

**Wave 22 (42 queries: row values, `(a, b) IN (SELECT ...)`, three-level subqueries, top-1 per group through ORDER BY LIMIT, explicit NULLS FIRST and LAST, IS DISTINCT FROM, CASE as a group key, aliases in HAVING and ORDER BY, ORDER BY and LIMIT on a set operation)**

- ORDER BY, LIMIT and OFFSET at the end of a set operation were handed to the first branch: with a LIMIT the branch came back with all its columns and the union failed with "left has 4, right has 1"; an OFFSET was skipped from the first branch and then again from the whole. The first branch now runs without them and they apply to the union. Test.

36/42 agree. Four are dialect where SQLite and DuckDB disagree between themselves (row-value NOT IN and comparisons with NULLs). The last two are the NULL order rule again: `ORDER BY b DESC LIMIT 1` inside a correlated scalar picks the NULL row under NULLS FIRST on DESC, and `INTERSECT ... ORDER BY a DESC` puts NULL first; both engines put NULLs last. `ORDER BY -b LIMIT 1` matches them exactly. That decision is still yours.

Two notes for later, not changed here:
- `SELECT ux.* FROM ux JOIN uy USING (a) ORDER BY uy.id` fails to compile ("unexpected '*' in expression context"): a qualified star with an ORDER BY column from the other table goes through the extra-column projection, which cannot expand the star. It predates this branch.
- The three-level correlated queries in this wave (`WHERE b > (SELECT AVG(b) FROM y WHERE y.a = x.a AND y.b > (SELECT MIN(b) FROM x z WHERE z.a = y.a))`) take minutes at 2500 rows; correct, but the inner subquery is re-run per y row per x row.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 21 | 1090 | see earlier comments | |
| 22 | 42 | 36 (+4 dialect, +2 NULL order rule) | ORDER BY, LIMIT and OFFSET on a set operation |

Every new test fails on 5d7b4d5a and passes after. fmt, clippy and both nextest suites are green.

semihalev:
Wave 23 (43 statements: UPDATE ... FROM, WITH before UPDATE and DELETE, INSERT ... SELECT with ORDER BY and LIMIT, RETURNING with expressions, a composite conflict target, DEFAULT VALUES, a SET that reads the table it changes).

Two of the findings were silent, and both are about keys:

- `PRIMARY KEY (k1, k2)` at table level was dropped on the floor: no uniqueness, duplicates accepted, `ON CONFLICT (k1, k2)` inserting a second copy. A composite table-level key is now kept unique the way a composite UNIQUE is (the row id stays the table's own), and ON CONFLICT finds it. `PRIMARY KEY (id)` over a single column at table level was dropped too, so `CREATE TABLE t (id INTEGER, v INTEGER, PRIMARY KEY (id))` had no key at all; it is now the column's primary key. Tests cover duplicates, DO UPDATE and DO NOTHING on both.
- `UPDATE dv SET t = DEFAULT` did nothing, silently. It now writes the column's default value, its default expression, or NULL when it has none.

Two were parse errors, one of which is filled in:

- `WITH ... UPDATE` and `WITH ... DELETE` are parsed now; the CTEs are materialized first and the statement's subqueries read them, recursive ones included. The CTE loop moved out of the SELECT path into a helper the three of them share.
- `UPDATE ... FROM` is still a parse error. It is a join inside UPDATE and a feature of its own, so I left it. The six remaining differences in this wave are all that one statement's absence carried forward (I ran the later shapes, INSERT ... SELECT ORDER BY LIMIT, the column swap, DELETE ... NOT EXISTS, on fresh data and they agree).

The three dialect entries are SQLite not taking the DEFAULT keyword in VALUES and SET.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 22 | 1132 | see earlier comments | |
| 23 | 43 | 34 (+3 dialect, +6 from the UPDATE ... FROM gap) | table-level primary keys, WITH before DML, SET DEFAULT |

Every new test fails on fa60aeb2 and passes after. fmt, clippy and both nextest suites are green.

semihalev:
Wave 24 (42 queries: LIKE with ESCAPE and case, text ordering, DISTINCT and GROUP BY on text, mixed integer and float aggregates, COUNT(1), MIN and MAX of text, HAVING on text aggregates, comparisons across integer and float, text length in filters).

39/42 agree, nothing to fix. The three left:

- `LIKE 'abc'` against 'ABC': SQLite matches case-insensitively, DuckDB does not, stoolap sides with DuckDB. Dialect.
- The two ESCAPE cases fail in stoolap with "unterminated string literal", because a backslash inside a single-quoted literal is an escape character for stoolap's lexer (`'\'` never closes) while SQLite, DuckDB and PostgreSQL read it as a plain character. The lexer's escapes are relied on by the statistics serializer, which writes `\\` into SQL it reads back, so this is a deliberate dialect of ours rather than a slip. Worth a decision at some point, since it changes what a user can put in a literal, but not one I would make inside this PR.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 23 | 1175 | see earlier comments | |
| 24 | 42 | 39 (+3 dialect) | none needed |

semihalev:
Wave 25 (61 statements, index maintenance: predicates over indexed columns with NULLs, ranges, IN and NOT, the same predicates after UPDATEs that move rows across the index, DELETEs of indexed rows and re-inserts, plus views over views over views).

61/61 across the three engines, nothing to fix.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 24 | 1217 | see earlier comments | |
| 25 | 61 | 61 | none needed |

semihalev:
Wave 26 (58 statements, transactions against indexes: reads through an index inside BEGIN after UPDATE and DELETE, the same reads after ROLLBACK and after COMMIT, delete then re-insert of the same key, aggregates that mix hot and committed rows).

- Inside BEGIN, deleting a row and inserting its primary key again failed with "already exists". The key check read the committed row the transaction had only deleted locally, and the unique index check found the committed entry the same way. Both now look past rows the transaction itself has deleted, as an autocommit statement already did; a key another row still holds is refused as before, and a rollback leaves the old row. Tests with and without a secondary index.

43/58 agree across the three, the other 15 are one DuckDB dialect: DuckDB refuses to re-insert a key deleted in the same transaction and aborts the transaction, so everything after it in that block errors there. SQLite takes it, and stoolap now matches SQLite on every one of those 15.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 25 | 1278 | see earlier comments | |
| 26 | 58 | 43 (+15 DuckDB dialect, stoolap = SQLite on all) | re-insert of a deleted key in a transaction |

The new tests fail on f60e6da4 and pass after. fmt, clippy and both nextest suites are green.

semihalev:
Wave 27 (49 statements, visibility inside one transaction: rows inserted then updated then deleted in the same BEGIN, updates and deletes whose subqueries read same-transaction inserts, aggregates and joins over them, then COMMIT or ROLLBACK).

49/49 across the three engines, nothing to fix.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 26 | 1336 | see earlier comments | |
| 27 | 49 | 49 | none needed |

Running total: 1385 queries in 27 waves; every silent wrong result they turned up is fixed on this branch, and what is left is listed in the earlier comments as dialect decisions or parser gaps.

semihalev:
Three more P2s on 3fe69af7, all three confirmed and fixed, plus wave 28 and a regression pass over every earlier wave.

**P2s**

- `SET r = DEFAULT` on a column added with `DEFAULT (RANDOM())` wrote the one value computed at ALTER time to every row. The default expression now comes first, the stored value only where there is no expression, so the update draws RANDOM() per row. Test: four rows, four distinct values.
- The columns of a composite `PRIMARY KEY (k1, k2)` were nullable, so the unique index that keeps the key let `(NULL, 1)` in twice. They are NOT NULL now, and either component being NULL is refused. Test.
- Deleting `(1, 10), (2, 20)` and inserting `(1, 20), (2, 10)` in one transaction failed on the second row: the committed index still said 10 belonged to row 1, and row 1's latest local version was no longer a delete. The unique check now compares the committed entry against the row's latest local version, so an entry the transaction has overwritten with another value no longer counts. Test swaps the values and then refuses a fresh duplicate.

One thing I saw while probing that third one and did not change: inserting a duplicate of a value another row inserted *in the same transaction* is caught at COMMIT (the commit-time unique check), not at the INSERT, because the committed index does not hold local rows. Not silent, and the fix would be a scan of the local versions per insert; noting it rather than doing it here.

**Wave 28 (30 queries, composite shapes: CTEs feeding windows feeding joins, correlated subqueries over CTEs, set operations inside CTEs, windows over set operations, derived tables three deep with a window at each level, EXISTS over a window result, aggregates of window results)**

- A correlated subquery whose FROM is a derived table, a CTE or a view bound `x.a` to the source's own `a`: the outer reference was never substituted on those paths and the bare-name fallback found the inner column, so `EXISTS (SELECT 1 FROM (...) w WHERE w.a = x.a)` held for every row and `(SELECT mb FROM t WHERE t.a = x.a)` returned "more than one row". The parent's columns are now bound in scope before the filter on all three paths.
- A window over a GROUP BY result ordering by an aggregate the select list did not carry, `ROW_NUMBER() OVER (ORDER BY SUM(b) DESC, a)`, ordered by nothing: the aggregate was never computed. Aggregates inside a window's arguments, PARTITION BY and ORDER BY are collected with the rest now.

Now 30/30.

**Regression pass.** I re-ran the stoolap side of every earlier wave against the current tree. Every remaining difference is one already on the list: the NULL order rule on DESC, and the three parser gaps (`WITH RECURSIVE ... UNION` without ALL, `WINDOW w1 AS (...), w2 AS (...)` with two definitions, `WITH` inside a subquery). Nothing that had agreed disagrees now.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 27 | 1385 | see earlier comments | |
| 28 | 30 | 30 | correlated subqueries over derived sources, windows ordered by an aggregate |

Every new test fails on 3fe69af7 and passes after. fmt, clippy and both nextest suites are green.

semihalev:
Wave 29 and the two P2s on 63d408ba.

**Wave 29 (24 queries, composite shapes, second round: correlated subqueries over derived sources inside UPDATE and DELETE, references two levels up, subqueries in JOIN ON over derived sources, HAVING correlated to a derived source, IN over a windowed derived table, scalar over a join of derived tables)**

- A scalar subquery whose only tie to the parent sat in a derived table in its FROM, `(SELECT MAX(c) FROM (SELECT COUNT(*) AS c FROM y WHERE y.a = x.a GROUP BY b) t)`, counted as uncorrelated, ran once and was reused for every parent row. Both correlation walks now descend into the FROM's derived tables and join conditions. 3de322a2.
- A correlated WHERE over a join of derived tables, `p.a = x.a`, was pushed to p's side unbound, and the side's filter read `x.a` as its own `a`: every pair matched. The parent's columns are bound in scope before the join splits its predicates, and the per-row path after the join binds them the same way. 16515159.

21/24 now. Two are SQLite lacking ALL/ANY. The last is a correlated subquery inside JOIN ON, `ON t.a = x.a AND x.b > (SELECT AVG(b) FROM y WHERE y.a = t.a)`: stoolap evaluates an ON subquery once, unbound, so the join yields nothing, with a base table as well as a derived one. The same predicate in WHERE is right. Both engines take it in ON; it needs per-pair evaluation in the join filter, which is a feature rather than a fix, so I am noting it here.

**P2s**

- Aggregates in a named WINDOW definition were not collected: `RANK() OVER w ... WINDOW w AS (ORDER BY SUM(b) DESC)` ranked every group 1. `parse_aggregations` now reads `stmt.window_defs` too. Test covers ORDER BY and PARTITION BY in the definition.
- `value_to_expression` turned a TIMESTAMP, JSON or VECTOR into a plain text literal, so a bound parent value changed type: `TYPEOF(x.ts)` inside a correlated derived source said TEXT. A timestamp now carries its type as the literal's hint, the way `TIMESTAMP '...'` does when written; JSON and a vector become an explicit `CAST(... AS JSON)` / `CAST(... AS VECTOR)`, because a JSON hint alone reads as the loose one the parser puts on any text shaped like an object (my first cut coerced those and broke `JSON_VALID('{bad')`, so the CAST route it is). Test checks TYPEOF and a timestamp comparison through a derived source.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 28 | 1415 | see earlier comments | |
| 29 | 24 | 21 (+2 SQLite dialect, +1 ON-subquery gap) | derived FROM correlation, join-side binding |

Every new test fails on the commit before its fix and passes after. fmt, clippy and both nextest suites are green.

semihalev:
Wave 30 (28 statements, composite shapes, third round: correlated subqueries over derived sources and CTEs inside UPDATE and DELETE, HAVING correlated to a derived source, GROUP BY over correlated derived counts, views read by correlated subqueries, the same inside a transaction with ROLLBACK).

Two view fixes came out of it:

- A correlated subquery over a view bound `x.b` to the view's own `b` once a second parent column came in: `EXISTS (SELECT 1 FROM vy WHERE vy.a = x.a AND vy.b > x.b)` matched nothing, `x.b > (SELECT AVG(b) FROM vy v WHERE v.a = x.a)` averaged over every row. The view path now binds the parent's columns in scope before its filter, as the derived-table and CTE paths do since 9c135f5b. (My earlier view test passed on the old code through another path, which is why the view got left out then.)
- A predicate the join pushed to a view on one of its sides was dropped: `SELECT COUNT(*) FROM vy v JOIN y ON y.id = v.id WHERE v.a = 2` returned every pair. Nothing to do with correlation, the view branch of the join-side executor simply never applied the filter it was handed. It does now, the way the CTE branch always did. That one predates this branch.

28/28 now.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 29 | 1439 | see earlier comments | |
| 30 | 28 | 28 | view binding, view join-side filter |

The new tests fail on 3ccf29aa and pass after. fmt, clippy and both nextest suites are green.

semihalev:
Wave 31 (31 statements, views under load: predicates of every kind pushed to a view on a join side, views in outer joins, in set operations and as aggregate sides, views of views read by correlated subqueries, DML through EXISTS over a view).

31/31 across the three engines, nothing to fix.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 30 | 1467 | see earlier comments | |
| 31 | 31 | 31 | none needed |

semihalev:
Wave 32 (43 queries, expression evaluation: operator precedence, unary minus and NOT, IS TRUE and IS FALSE, chained and mixed comparisons, nested CASE, booleans in arithmetic and aggregates, concatenation precedence, parenthesised subexpressions in WHERE and ORDER BY).

Three fixes:

- `SELECT SUM(a + b * 2), SUM((a + b) * 2)` gave the same number twice. The aggregate stringifier dropped parentheses, so the two were one name and one computation; the same for `SUM(-(a + b))` against `SUM(-a + b)`. There were two stringifiers (utils and the executor's own), both lossy; there is one now, and it keeps the parentheses precedence needs and no others, so `a + b * 2` still reads as written. Test checks names and values for `*`, unary minus, `%` and non-associative `-`.
- `NOT (a = 1 AND b > 10)` kept rows where the conjunction was UNKNOWN: the storage filter emulates NOT's three-valued logic through `is_unknown_due_to_null`, and AND and OR answered "never unknown", on the row form and on the two slice forms of the compiled filter alike. UNKNOWN AND FALSE is FALSE, UNKNOWN AND TRUE is UNKNOWN, and the mirror for OR, on all three. Test covers NOT over AND and over OR, alone and beside another conjunct. Two assertions in `expression_pushdown_test` had encoded the old behaviour (they counted the all-NULL product under `NOT (category = 'Electronics' AND price > 500)`); they say 7 now, which is what both engines say.
- `a IN (b, b + 1, b - 1)` was a compile error ("Dynamic IN list not yet supported"). Items that are not constants are compared in turn, `a = b OR a = b + 1 ...`, with IN's NULL rule falling out of OR's. Test covers IN, NOT IN, a NULL item and an expression item.

41/43 now; the two left are DuckDB refusing `boolean + boolean`, where SQLite and stoolap add them.

| Wave | Queries | Agree | Fixed here |
|---|---|---|---|
| 1 to 31 | 1498 | see earlier comments | |
| 32 | 43 | 41 (+2 DuckDB dialect) | aggregate names, NOT over AND/OR, IN over expressions |

Every new test fails on 081e5d21 and passes after. fmt, clippy and both nextest suites are green.

PR #100, opened 2026-09-06
docs(mcp): describe the 0.4.x MCP server

semihalev:
The MCP server page still described the pre-0.4 server. This brings it in line with the `@stoolap/mcp` 0.4.x README and source.

What changed:

- Requirements: Node >= 20, Windows x64 added, note on the N-API addon and the slow first `npx` run with the global-install workaround.
- CLI options: `--path` accepts a DSN with engine options in the query string, `--version` added, `--read-only` also covers schema and engine state.
- Tools: `query` takes VALUES and WITH ... SELECT; `execute` covers COPY, DDL, SET, ANALYZE, VACUUM and ON CONFLICT upsert; `explain analyze` refuses writes; `transaction_execute` refuses DDL, TRUNCATE and COPY; `pragma` lists the real 0.4.x pragmas and actions.
- Auto-injected instructions and resources match the new compact instruction set (JSON vs Markdown noted).
- Parameter binding: `?` placeholders alongside `$1`.
- Examples: ON CONFLICT upsert and COPY ... FROM added; the vector example notes the distance function must match the index metric.
- SQL coverage and safety lists aligned with the README, including the COPY host-file warning, execute blocked during an open transaction, and clean shutdown on SIGINT/SIGTERM.

Docs only, no code changes.

### DIFF
--- a/docs/_docs/drivers/mcp.md
+++ b/docs/_docs/drivers/mcp.md
@@ -12,17 +12,23 @@ MCP ([Model Context Protocol](https://modelcontextprotocol.io)) server for Stool
 
 Works with any MCP-compatible AI client: Claude Desktop, Claude Code, Cursor, Windsurf, Cline, and others.
 
-The server provides 30 tools, 2 resources, and 1 prompt. On connection, it sends built-in instructions so the AI can write correct Stoolap SQL from the first query.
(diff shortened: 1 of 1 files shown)


semihalev:
Two follow-up fixes pushed:

- The page now states it describes `@stoolap/mcp` 0.4.1 or later and suggests pinning `@stoolap/mcp@^0.4.1`. The COPY safety bullet says outright that 0.4.0 does not apply the read-only check to COPY ... FROM.
- Building from source uses `npm run build` again, since the published package.json has no `test` script.

This PR depends on the stoolap-mcp 0.4.1 release; npm latest is still 0.4.0. Merge after that is published.

PR #109, opened 2026-09-07
perf(index): answer a series' latest rows by walking the multi-column index in key order

semihalev:
## Summary

The "latest N of one series" query over hot rows read the whole series. With `UNIQUE(exchange, symbol, time)` and `WHERE exchange = .. AND symbol = .. ORDER BY time DESC LIMIT 200`, the hot store took every row id of the pair from the index prefix, fetched and cloned all 12,500 rows, ran the WHERE on each again, and the executor sorted them to keep 200. `COUNT(*)` of the pair cost the same 35 ms, so the sort was never the problem. In production the newest candles are always hot, so this is the shape the application hits most, and it grows with every row that has not been sealed yet.

- `Index::walk_prefix_ordered`: visits the row ids whose key starts with a prefix, in the order of the next key column, within optional bounds on that column, ascending or descending, stopping when the visitor says so. The multi-column index keeps, per prefix group that a walk has asked for, a `BTreeSet<(value of the walked column, row_id)>`: built from the prefix index and the row-to-key map on the first walk (sorted once, bulk-built), kept up to date by single adds and removes, dropped by a batch of 1024 rows or more (a seal) and rebuilt on the next walk. It holds one value per row, not a copy of the key; a first draft with a BTreeMap over every composite key cost 700 MB and 2 s on 2.175M rows and doubled the seal, and is gone.
- `MVCCTable::scan_top_k`: when the WHERE is a conjunction of comparisons whose equalities pin the leading columns of a multi-column index and the ORDER BY column is the next index column, it walks that index, reads each candidate's visible version, checks the whole WHERE on the row (the shared index can be ahead of a transaction's view of a row) and stops at LIMIT + OFFSET. It declines for a transaction with its own writes (they are not in the shared indexes), a nullable or FLOAT order column, a WHERE it cannot read, or an index that cannot walk.
- `SegmentedTable::scan_top_k`: without volumes it hands the question to the hot store instead of answering None, without taking the seal fence (a seal registers its volume before it removes a hot row, so a table seen without volumes still has every row hot; a draft that took the fence made a reader wait out the whole seal, 1.9 s). With volumes it takes the hot store's top-k when there is one; the sealed copy of a hot row is then recognised per candidate through `has_row_id` instead of a skip set built from every matching hot row.
- The multi-column index's batch removal takes its locks per 8192 rows, so a reader waits for one chunk of a seal's removal, not for all of it.

Review fixes (second push):

- Bounds that cross (`t > 100 AND t < 50`) or meet with one side open (`t > 50 AND t < 50`) hold nothing: the walk answers empty before asking the set for a range, which panicked.
- A snapshot transaction is declined: it sees the versions of its start while the index holds the keys of the latest commits, so the key order is not its order. For every other transaction, a candidate whose visible order value differs from its key (changed between the index and the version seen) stops the walk and the full scan answers instead.
- Without volumes, the hot answer is returned only if the seal generation did not move while it was computed; a first seal that registered its volume meanwhile sends the query through the segment-aware path, which reads hot and volumes together.
- A group's order is built under the orders' write lock, from the prefix index and the row map, and every writer updates the orders last, after the prefix indexes: a row is either in the group when it is built or entered into it afterwards, whichever way an add or remove interleaves with the first walk. The test holds a lock the writer needs between its row-map and prefix updates and builds the group in that window.

Review fixes (third push):

- A commit updates the indexes before its versions are visible, and a key it moved can sit anywhere in the walk, past the LIMIT included, so checking only the visited candidates was not enough. The version store now counts commits that are publishing (`begin_publish`, from the index updates to the versions being visible) and the publishes completed; the walk stands down while a commit publishes and falls back to the full scan when one overlapped it. Writers are never blocked. The test stops a real `UPDATE` commit right after its index add, through an `Index` wrapper installed with `add_index`, and expects the top-k to agree with the full sort until the commit completes.
- A built group is walked under the shared lock; only a missing group takes the exclusive lock, and checks again once it holds it. The test holds a walk of one group inside its visitor and expects a walk of another group to proceed.

Review fixes (fourth push):

- The publish guard now belongs to the transaction, not to one table's commit: `TransactionEngineOperations::begin_publish` marks every table the transaction writes as publishing before the first index update, and the hold lives until the transaction is visible in the registry or undone. A reader of the first table in a two-table commit stands down until the whole transaction is visible.
- A commit that fails at the WAL after its index updates were applied takes them back: both index update paths record what they added and removed, and the WAL failure branch undoes it in reverse before the transaction is aborted. This also fixes an older gap on main, where an equality lookup on the changed column could not find the row after such an abort. The test arms the WAL sync failpoint around an UPDATE and expects the index-ordered read, the full sort, and an equality lookup on the old key to agree.

Review fix (fifth push): the walk no longer re-checks a visible version's raw deleted flag. `get_visible_version` already returns a version only when it is visible and its deletion, if any, is not; the flag can belong to an aborted or unfinished DELETE, and the extra check dropped a row an aborted DELETE had marked. The failpoint test covers an aborted DELETE and an aborted INSERT on top of the aborted UPDATE.

`examples/hot_probe.rs` and `examples/candle_bench.rs` (not committed): 2.175M hot rows, 174 series, release build, p50.

| shape (hot, unsealed) | main | this branch |
|---|---|---|
| pair latest DESC LIMIT 200 | 36.0 ms | 0.011 ms |
| pair latest, a different pair every run | 36 ms | 0.043 ms |
| pair + day range DESC LIMIT 500 | 16.6 ms | 0.03 ms |
| pair + time >= day DESC LIMIT 200 | 5.4 ms | 0.01 ms |
| fan-out 174 series on 4 threads | 2123 ms wall, p50 48 ms | 27 ms wall, p50 0.5 ms |
| all pairs last 10 min (BTREE on time) | 0.11 ms | 0.12 ms |
| pair ORDER BY id DESC LIMIT 200 (not an index column) | 35.4 ms | 33.3 ms |
| COUNT(*) WHERE pair | 34.3 ms | 37.1 ms |
| first query after the load (prefix index build) | 308 ms | 403 ms |
| PRAGMA CHECKPOINT of the loaded rows | 3.2 s | 3.0 s |
| reader latency during the seal, max | 21 ms | 23 ms |

Sealed and reopened shapes are unchanged. The two shapes that still read the whole series are outside this change: `ORDER BY id` is not the index's next column, and COUNT reads every row of the pair.

Not in this change: skipping the WHERE re-check when the index covers it. Indexes are updated at commit and are not versioned, so a reader on an older snapshot can meet a key whose visible row no longer matches; the re-check stays. It costs nothing here, since the walk reads N rows instead of the series.

`tests/hot_index_top_k_test.rs` compares the top-k with the full sort over hot rows: ASC and DESC, OFFSET, every bound form on the order column, a predicate the index does not cover, only the first index column pinned, deleted and updated rows, a transaction's own inserts and deletes, a non-unique index with duplicate keys (the t sequence must match, ids only where t is unique), a nullable order column, hot rows mixed with sealed ones before and after a reopen, bounds that hold nothing, a snapshot reading past a concurrent update of the order column, a first seal fired from inside the WHERE of a hot top-k, a real UPDATE commit stopped between its index update and its versions, a two-table transaction stopped in its second table, and a commit aborted by a WAL sync failure (in `tests/failpoint_io_test.rs`). The multi-column index gets a unit test for the build-during-add interleaving.

## Test plan

- [x] `cargo nextest run --test hot_index_top_k_test --test ordered_limited_scan_test`
- [x] `cargo fmt`, `cargo clippy --all-targets --all-features -- -D warnings`
- [x] `cargo nextest run`
- [x] `cargo nextest run --features test-filedb`

### DIFF
--- a/src/storage/index/multi_column.rs
+++ b/src/storage/index/multi_column.rs
@@ -130,7 +130,10 @@ pub struct MultiColumnIndex {
 
     /// Per prefix group, its rows ordered by the column after the prefix:
     /// built when a walk first asks for the group, kept up to date by single
-    /// adds and removes, dropped by a large batch and rebuilt on the next walk
+    /// adds and removes, dropped by a large batch and rebuilt on the next walk.
+    /// Writers touch it last, after the prefix indexes a build reads, and a
+    /// build holds its write lock from that read to the insert, so a row is
+    /// either in the group when it is built or added to it afterwards.
     walk_orders: RwLock<FxHashMap<CompositeKey, BTreeSet<(Value, i64)>>>,
 }
 
@@ -494,13 +497,6 @@ impl Index for MultiColumnIndex {
         drop(value_to_rows);
         drop(row_to_key);
 
-        if self.orders_built() {
-            if let Some(ref old_arc_values) = old_key_for_cleanup {
-                self.order_remove_arcs(old_arc_values, row_id);
-            }
-            self.order_insert(values, row_id);
-        }
-
         // Update BTree only if it was already built
         if btree_needs_update {
             let mut sorted_values = self.sorted_values.write();
@@ -558,6 +554,13 @@ impl Index for MultiColumnIndex {
             }
         }
 
+        if self.orders_built() {
+            if let Some(ref old_arc_values) = old_key_for_cleanup {
+                self.order_remove_arcs(old_arc_values, row_id);
+            }
+            self.order_insert(values, row_id);
+        }
+
         Ok(())
     }
 
@@ -594,10 +597,6 @@ impl Index for MultiColumnIndex {
             row_to_key.remove(row_id);
         }
 
-        if self.orders_built() {
-            self.order_remove(values, row_id);
-        }
-
         // Only update BTree if it was built (row_ids are sorted, use binary search)
         if self.btree_built.load(AtomicOrdering::Acquire) {
             let mut sorted_values = self.sorted_values.write();
@@ -626,6 +625,10 @@ impl Index for MultiColumnIndex {
                     }
                 }
             }
+        }
+
+        if self.orders_built() {
+            self.order_remove(values, row_id);
         }
 
         Ok(())
@@ -761,19 +764,6 @@ impl Index for MultiColumnIndex {
         drop(value_to_rows);
         drop(row_to_key);
 
-        if self.orders_built() {
-            if entries.len() >= Self::WALK_ORDER_BATCH_DROP {
-                self.walk_orders.write().clear();
-            } else {
-                for (row_id, existing_arc_values) in &updates_to_old_key {
-                    self.order_remove_arcs(existing_arc_values, *row_id);
-                }
-                for &(row_id, values) in entries {
-                    self.order_insert(values, row_id);
-                }
-            }
-        }
-
         // Update BTree only if it was already built
         if btree_needs_update {
             let mut sorted_values = self.sorted_values.write();
@@ -835,6 +825,19 @@ impl Index for MultiColumnIndex {
             }
         }
 
+        if self.orders_built() {
+            if entries.len() >= Self::WALK_ORDER_BATCH_DROP {
+                self.walk_orders.write().clear();
+            } else {
+                for (row_id, existing_arc_values) in &updates_to_old_key {
+                    self.order_remove_arcs(existing_arc_values, *row_id);
+                }
+                for &(row_id, values) in entries {
+                    self.order_insert(values, row_id);
+                }
+            }
+        }
+
         Ok(())
     }
 
@@ -869,16 +872,6 @@ impl Index for MultiColumnIndex {
             }
         }
 
-        if self.orders_built() {
-            if entries.len() >= Self::WALK_ORDER_BATCH_DROP {
-                self.walk_orders.write().clear();
-            } else {
-                for &(row_id, values) in entries {
-                    self.order_remove(values, row_id);
-                }
-            }
-        }
-
         // The sorted and prefix structures are subtracted one key at a time:
         // a removal per row is quadratic in the rows per key.
         if self.btree_built.load(AtomicOrdering::Acquire) {
@@ -906,6 +899,16 @@ impl Index for MultiColumnIndex {
                             prefix_index.remove(&key);
                         }
                     }
+                }
+            }
+        }
+
+        if self.orders_built() {
+            if entries.len() >= Self::WALK_ORDER_BATCH_DROP {
+                self.walk_orders.write().clear();
+            } else {
+                for &(row_id, values) in entries {
+                    self.order_remove(values, row_id);
                 }
             }
         }
@@ -1077,7 +1080,7 @@ impl Index for MultiColumnIndex {
         lower: Option<(&Value, bool)>,
         upper: Option<(&Value, bool)>,
         ascending: bool,
-        visit: &mut dyn FnMut(i64) -> bool,
+        visit: &mut dyn FnMut(i64, &Value) -> bool,
     ) -> bool {
         let walked = prefix.len();
         if self.closed.load(AtomicOrdering::Acquire)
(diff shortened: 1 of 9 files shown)


semihalev:
Second push addresses the four findings; each has a test that fails on the first push.

1. Crossing or open-meeting bounds (`t > 100 AND t < 50`, `t > 50 AND t < 50`, `t >= 50 AND t < 50`, `t > 50 AND t <= 50`) answer empty before the set is asked for a range. Four SQL shapes in `test_hot_top_k_with_bounds_that_hold_nothing_is_empty`, ASC and DESC.
2. A snapshot transaction is declined in `MVCCTable::scan_top_k` through `needs_snapshot_isolation`; the full sort answers it. In addition, for any transaction, a candidate whose visible order value differs from its index key stops the walk and falls back, so an update of the order column racing the read cannot reorder the answer. `test_snapshot_orders_by_the_values_it_sees` runs the snapshot past an UPDATE of the newest row's time from another handle.
3. The hot-only answer is returned only when the seal generation is unchanged after computing it; a first seal that registered its volume meanwhile falls through to the segment-aware path, which reads hot and volumes together. `test_first_seal_during_the_hot_top_k_keeps_the_series` fires the checkpoint from inside the WHERE as in the review and expects `Some([1])`.
4. The group order is built under the orders' write lock, and every writer (add, remove, both batch paths) updates the orders last, after the prefix indexes the build reads. A row is either in the group when it is built or entered into it afterwards, whichever way the add interleaves. `test_walk_order_built_during_an_add_keeps_the_added_row` holds the sorted structure so the add stops between its row-map and prefix updates, builds the group in that window, and expects both rows afterwards.

On the fan-out measurement: agreed, the bench clones the handle per worker and does not measure the driver's single-executor queueing; the wall numbers stand for the storage path only.

semihalev:
Third push addresses both findings; each has a test that fails on the second push.

1. The commit window is closed at its source rather than at the visited rows: the version store counts commits that are publishing (from the index updates to the versions being visible) and the publishes completed. `MVCCTable::scan_top_k` reads that before and after the walk and falls back to the full scan if a commit was publishing or one completed meanwhile; writers never wait. `test_top_k_during_a_commit_answers_from_the_visible_versions` stops a real `UPDATE` commit right after its index add and expects the top-k to agree with the full sort until the commit completes; the review's `read_committed_top_k_during_actual_update_commit` passes on this head as well. The synthetic `changed_key_beyond_limit_cannot_hide_visible_latest_row` still fails on purpose: it edits the index directly with no commit in flight, a state the engine cannot reach (index keys change only inside a commit's publish, which the counter covers, or at seal and recovery, which do not run against a live walk). If there is a reachable path to that state that I am missing, I would like to hear it.
2. A built group is walked under the shared lock; only a missing group takes the exclusive lock and re-checks under it. `test_built_groups_are_walked_concurrently` holds one group's walk inside its visitor and expects another group's walk to proceed; the review's `warm_independent_groups_can_be_read_concurrently` passes too.

semihalev:
Fourth push addresses both findings; each has a test that fails on the third push, and the review's own tests pass on this head.

1. The publish guard belongs to the transaction now. `begin_publish` on the engine operations marks every table the transaction writes as publishing before the first index update; the hold lives until `complete_commit` or the abort, so a reader of the first table in a two-table commit stands down until the whole transaction is visible. `test_top_k_stands_down_until_a_two_table_commit_is_visible` stops the second table's index add as the review did; `publish_guard_covers_transaction_visibility_across_tables` passes on this head.
2. A commit that fails at the WAL after its index updates were applied now takes them back. Both index update paths record what they added and removed, and the WAL failure branch undoes it in reverse before the transaction is aborted, so the indexes describe the rows that stayed visible. This also closes an older gap on main: after such an abort an equality lookup on the changed column could not find the row, since its old key had been removed. `test_wal_sync_failure_undoes_the_commit_index_updates` (in `failpoint_io_test.rs`, which CI runs with the failpoints feature) arms the WAL sync failpoint around an UPDATE and expects the index-ordered read, the full sort and an equality lookup on the old key to agree; `aborted_wal_commit_keeps_the_visible_top_k` passes on this head.

semihalev:
Fifth push addresses the finding. The walk took a visible version and then looked at its raw deleted flag, which a DELETE that aborted at the WAL leaves set with the aborted transaction's id; `get_visible_version` already answers None when the deletion is visible, so the extra check only ever dropped rows whose deletion was not. It is gone. `test_wal_sync_failure_on_a_delete_keeps_the_row_everywhere` in `failpoint_io_test.rs` aborts a DELETE and an INSERT at the WAL and expects the full sort, the index-ordered read, and the equality lookups on both indexes to agree; the review's three extended undo tests pass on this head. The test fails on the fourth push.

PR #110, opened 2026-09-08
perf(seal): remove sealed index entries by row id

semihalev:
The seal's hot removal did three things per sealed row that scale badly at volume size:

- it cloned the whole version tree (an O(1) CowBTree clone, but every later index cleanup then re-read the row data through it and rebuilt a `Vec<Value>` per row per index);
- `BTreeIndex` and `MultiColumnIndex` removed each row from its key's row list with a shift, so a key with many rows paid a quadratic removal (a `k` column with a few hundred distinct values over a million-row seal);
- the hash and bitmap indexes went through the same value-based path even though they already know each row's key.

This PR removes sealed index entries by row id. `Index` gets `remove_batch_ids`, which every built-in index implements from its own row-to-key map: BTREE, hash and the multi-column index group the ids by key and subtract each group from the key's row list in one linear pass; bitmap and PK remove directly. The default returns `None`, and `remove_sealed_index_entries` then falls back to the values from the sealed rows themselves (the same `RowVec` that built the volume), so an index type without the fast path still works. `remove_sealed_rows` no longer takes a snapshot of the version tree for the cleanup and `SealedIndexCleanup` is just the removed ids.

Nothing changes in what the seal keeps or skips: claimed rows and rows whose head changed since extraction still stay hot and are tombstoned as before.

Measured with a probe (2M rows per round, 200 distinct `k` values, UNIQUE(k, t) plus BTREE on `k` and on `t`, PRAGMA CHECKPOINT per round, release build):

| round | seal main | seal this PR |
|---|---|---|
| 1 | 4045 ms | 2524 ms |
| 2 | 4098 ms | 2489 ms |
| 3 | 4125 ms | 2561 ms |

Verification: clippy with `-D warnings`, the seal-related suites (`sealed_index_visibility_test`, `volume_checkpoint_test`, `snapshot_cold_test`, `volume_concurrency_test`, `arena_slot_reuse_test`, `hot_index_top_k_test`, `ordered_limited_scan_test`, `failpoint_io_test`), the in-memory suite and the filedb suite.

### DIFF
--- a/src/storage/index/bitmap.rs
+++ b/src/storage/index/bitmap.rs
@@ -630,6 +630,29 @@ impl Index for BitmapIndex {
         Ok(())
     }
 
+    fn remove_batch_ids(&self, row_ids: &[i64]) -> Option<Result<()>> {
+        if self.closed.load(AtomicOrdering::Acquire) {
+            return Some(Err(Error::IndexClosed));
+        }
+        let mut bitmaps = self.bitmaps.write();
+        let mut row_to_value = self.row_to_value.write();
+        for &row_id in row_ids {
+            if row_id < 0 {
+                continue;
+            }
+            if let Some(arc_key) = row_to_value.remove(row_id) {
+                if let Some(bitmap) = bitmaps.get_mut(&arc_key) {
+                    bitmap.remove(row_id as u64);
+                    if bitmap.is_empty() {
+                        bitmaps.remove(&arc_key);
+                        self.distinct_count.fetch_sub(1, AtomicOrdering::Relaxed);
+                    }
+                }
+            }
+        }
+        Some(Ok(()))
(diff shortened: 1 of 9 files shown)


semihalev:
Confirmed and fixed in the follow-up commit. A probe deleting the two highest ids under one BTREE key holding 1M rows, 200 times:

| code | total |
|---|---|
| main | 22.9 ms |
| first commit of this PR | 116.9 ms |
| with the fix | 19.8 ms |

The BTREE and hash paths now use the same helper as the multi-column index: the pass starts at the first removed id, so removing from the top of a large group touches only the tail, and removing a whole group is still one pass. The helper moved to `storage/index/mod.rs` so the three indexes share it.

PR #111, opened 2026-09-08
feat(storage): hot size limits, exact hot byte counters, PRAGMA MEMORY_STATS

semihalev:
Second step of the hot store work: the hot store gets size limits, exact byte counters and a way to see them. Nothing changes in the seal itself; this PR decides when it runs and lets a writer wait for it.

**Limits**

- `hot_max_rows` (default 262,144): a commit that takes a table's committed hot rows past it raises a seal request. The checkpoint thread, which polls every 100 ms, runs the seal cycle on its next tick instead of waiting for `checkpoint_interval`. The seal's own thresholds are bounded by the same value, so a table over the limit is always sealed by that cycle. 0 turns the trigger off and leaves today's behaviour.
- `hot_max_bytes` (default 0, off): a commit whose table holds at least this many hot bytes raises the same request and waits, before it takes the seal fence, until the seal brings the table back under. The wait is bounded at 10 seconds and counted, so a long snapshot that pins rows cannot hang writers; an in-memory database never waits because nothing can seal.
- Both are DSN keys (`hot_max_rows=`, `hot_max_bytes=`) and read/write PRAGMAs (`HOT_MAX_ROWS`, `HOT_MAX_BYTES`), mirrored into atomics so the commit path never touches the config lock.

**Counters**

The arena keeps an exact byte count of the rows it holds: 16 bytes per value in place plus what heap text and extension values own, added on every insert, moved on `update_at`, taken back on `clear_at`, `clear_batch` and `clear_all`. One atomic add per row; no per-row allocation.

**PRAGMA MEMORY_STATS**

One row per table plus a `*` row with the totals:

| column | meaning |
|---|---|
| hot_rows | committed rows in the hot store |
| hot_bytes | bytes of the rows the arena holds |
| arena_slots | slots in use, deleted and cleared ones included |
| arena_capacity_bytes | what the slot vectors reserve |
| chain_entries | previous versions kept alive by version chains |
| volume_bytes | memory of the table's loaded volumes |
| admission_waits | (`*` row) commits that waited for a seal |

**Tests** (`tests/hot_limits_test.rs`, each fails on main): DSN and PRAGMA round trip; MEMORY_STATS following inserts, longer UPDATEs, chain entries and VACUUM back to zero; `hot_max_rows` sealing 50k rows with `checkpoint_interval=3600` while every row stays readable; `hot_max_bytes` keeping the peak under the limit plus one batch with waits counted; an in-memory database ignoring the byte limit. Plus an arena unit test for the byte counter.

The default trigger changes when a bulk load seals: every 262,144 rows instead of once at the next interval. Measured with the candle load probe (2.175M rows in 10-minute batches, `checkpoint_interval=3600`, release build):

| phase | main | this PR |
|---|---|---|
| bulk load | 7976 ms | 10293 ms |
| PRAGMA CHECKPOINT after the load | 3334 ms | 801 ms |
| load + checkpoint | 11310 ms | 11094 ms |

The load pays for the eight seals it now runs; the checkpoint at the end has little left to do. RSS is not lower yet: the sealed volumes stay in the eager tier and the arena keeps its capacity for reuse, which the chunked arena and the memory budget steps address.

Verification: clippy with `-D warnings`, the storage suites (`sealed_index_visibility_test`, `volume_checkpoint_test`, `snapshot_cold_test`, `volume_concurrency_test`, `arena_slot_reuse_test`, `hot_index_top_k_test`, `ordered_limited_scan_test`, `wal_truncation_test`, `failpoint_io_test`), the in-memory suite and the filedb suite.

### DIFF
--- a/src/api/database.rs
+++ b/src/api/database.rs
@@ -546,6 +546,22 @@ impl Database {
                         })?;
                         config.persistence.target_volume_rows = rows.max(65_536);
                     }
+                    // Early seal trigger per table: hot_max_rows=262144 (0 = off)
+                    "hot_max_rows" => {
+                        config.persistence.hot_max_rows = value.parse::<usize>().map_err(|_| {
+                            Error::invalid_argument(format!("invalid hot_max_rows: '{}'", value))
+                        })?;
+                    }
+                    // Commit wait above this many hot bytes per table: hot_max_bytes=0 (off)
+                    "hot_max_bytes" => {
+                        config.persistence.hot_max_bytes =
+                            value.parse::<usize>().map_err(|_| {
+                                Error::invalid_argument(format!(
+                                    "invalid hot_max_bytes: '{}'",
+                                    value
+                                ))
+                            })?;
+                    }
                     // Checkpoint on close: checkpoint_on_close=off
                     // Set to off to simulate crashes in tests (WAL not truncated)
                     "checkpoint_on_close" => {
--- a/src/common/smart_string.rs
+++ b/src/common/smart_string.rs
@@ -242,6 +242,18 @@ impl SmartString {
         }
     }
 
+    /// Bytes a heap string owns beyond the SmartString itself: the String's
+    /// capacity, which `from_string` keeps as it came. Zero when inline.
+    #[inline]
+    pub fn heap_capacity(&self) -> usize {
+        if self.tag.is_heap() {
+            // SAFETY: heap strings hold a valid Arc<String> pointer
+            unsafe { (*self.get_arc_ptr()).capacity() }
+        } else {
+            0
+        }
+    }
+
     /// Returns the length of the string in bytes.
     #[inline(always)]
     pub fn len(&self) -> usize {
--- a/src/executor/query.rs
+++ b/src/executor/query.rs
@@ -9783,6 +9783,77 @@ impl Executor {
                     Ok(Box::new(ExecutorResult::new(columns, rows)))
                 }
             }
+            "HOT_MAX_ROWS" | "HOT_MAX_BYTES" => {
+                let config = self.engine.config();
+                let columns: Vec<String> = vec![pragma_name.to_lowercase().into()];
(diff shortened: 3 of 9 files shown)


semihalev:
All four are addressed in the follow-up commit.

1. **UPDATE waiting on its own claims**: the admission wait now leaves out any table where the transaction claimed existing hot rows (`TransactionVersionStore::holds_hot_rows`, a write set entry with a read version). New test `an_update_over_the_byte_limit_does_not_wait_on_its_own_claims`: 10.2 s before, 0.2 s now.
2. **Request consumed before visibility**: the trigger moved out of `commit_all_tables`. After `complete_commit`, the commit asks through its publish hold (`PublishHold::any_table_at`, the hold keeps the transaction's stores and their parents) whether a table it wrote is at `hot_max_rows`, and only then raises the request. The rows are visible when the checkpoint thread reads it.
3. **Lost wakeup**: the seal cycle takes the admission mutex around `notify_all`.
4. **String capacity**: `SmartString::heap_capacity` reads the String's capacity for heap strings; `row_bytes` counts that instead of the length. The arena unit test now builds a value from a `String::with_capacity(4096)` and fails with length-based counting.

Verification on the new commit: clippy with `-D warnings`, the storage suites and lib tests, the in-memory suite and the filedb suite.

semihalev:
Both fixed in the follow-up commit.

1. `PublishHold::any_table_at` now takes both limits, so a commit that became visible asks for a seal when a table it wrote is at `hot_max_rows` or at `hot_max_bytes`, whether or not it waited. New test `updates_that_grow_past_the_byte_limit_request_a_seal` (`hot_max_rows=0`, `checkpoint_interval=3600`, four growing UPDATEs): no volume within 10 s before, sealed in 0.5 s now.
2. The partial commit path calls the same hook right after its `complete_commit`.

Verification on the new commit: clippy with `-D warnings`, the storage suites and lib tests, the in-memory suite and the filedb suite.

PR #112, opened 2026-09-08
perf(volume): test dictionary equalities over a row group's raw ids in one pass

semihalev:
The dictionary pre-filter for text equality on cold rows ran row by row: `is_null` plus `get_dict_id`, each a match on the column enum, per row and per filter column. It ran that way in three places: the scanner's pre-scan in `set_filter` (both the group store and the eager branch), the row fetch path in `volume/table.rs` and the aggregation pushdown.

`ColumnData::dict_matching_offsets` takes the filter columns with a window and an expected id, walks the first column's raw id slice in one pass and reads the other columns only for its matches. The scanner's ordered walk already had this shape and now shares the helper. The three call sites apply it per row group or per range and fall back to the row-by-row test only when a filter column is not dictionary encoded. A range without dictionary filters is walked as before, no candidate vector is built for it.

Measured with a probe of a 1m to 3m candle aggregation upsert (4M one-minute candle rows for 174 pairs, sealed, release build):

| shape | main | this PR |
|---|---|---|
| upsert, lookback 7 days | 14.4-15.7 ms | 10.9-13.6 ms |
| upsert, all history | 89-104 ms | 82-93 ms |
| pair `COUNT(*)`, no time filter | 68.5 ms | 61.2 ms |

The gain is modest because the filter was not the main cost of this shape. A sample profile of the pair COUNT in a loop puts 83% of the time in `load_group_cache` decoding LZ4 groups: every needed column of every group is decoded per query, and the 64 MB decoded group cache cannot hold the working set (153 MB here), so nothing survives between queries. With `PRAGMA GROUP_CACHE_MB = 512` the same COUNT drops from 60 ms to 6.4 ms and the full upsert from 88 ms to 31 ms. That is a separate decision.

Tests: a unit test for the helper (nulls, several filters, window offsets, non-dictionary column, no filters); the volume, scan, pushdown and aggregation targets; the in-memory suite and the filedb suite. Clippy with `-D warnings`.

### DIFF
--- a/src/storage/volume/column.rs
+++ b/src/storage/volume/column.rs
@@ -27,6 +27,14 @@ use std::sync::Arc;
 use crate::common::SmartString;
 use crate::core::{DataType, Value};
 
+/// A dictionary column, the local index a caller's window starts at in it,
+/// and the id looked for
+pub type DictFilter<'a> = (&'a ColumnData, usize, u32);
+
+/// A dictionary column's raw ids and nulls, the local index its window
+/// starts at, and the id looked for
+type DictSlice<'a> = (&'a [u32], &'a [bool], usize, u32);
+
 /// Typed column data stored contiguously for cache-friendly access.
 ///
 /// Each variant stores a flat array of the native type plus a null bitmap.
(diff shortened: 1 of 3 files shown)


semihalev:
Fixed in the follow-up commit: the eager branch now hands the helper one row group at a time and checks `selectivity_cap` between groups, the same way the group store branch does, so a filter that matches most rows stops after the first groups instead of listing every match and dropping the list. Clippy with `-D warnings`, the scan targets, the in-memory suite and the filedb suite ran on the new commit.

PR #163, opened 2026-09-13
Record the single-row index undo log without allocating

semihalev:
Every committed single-row statement paid for its index undo log with fresh heap allocations that the success path never reads.

#109 introduced the publish protocol: while a commit publishes, each index change is recorded so that a failure later in the commit can take the indexes back. On the single-row path that record was built by collecting the key values a second time into a `Vec`, wrapping them in a one-element `Vec`, and cloning that for the `added` side, after the key had already been built as a `SmallVec` for the index call itself and dropped. The per-transaction undo `Vec` allocated on its first push, and `PublishHold` allocated its two vectors once per commit. Measured with a counting allocator on a three-column table with only its primary key, an ad hoc single-row `INSERT` went from 68 allocations at #108 to 75 at #109; the prepared form on main costs 20, of which the same 7 are this log. `undo_index_updates` runs only when a commit fails after its indexes were applied (`transaction.rs:513` and the `PublishHold` failure path), so on every successful commit those allocations were made and dropped unused.

`IndexUndo` is now an enum. Its `Row` form holds the `SmallVec<[Value; 2]>` the index call was made with, moved out of `completed_ops`, which now carries that key alongside the index position and direction; recording it allocates nothing, and `rollback_single_row_ops` uses the same keys instead of deriving them from the rows again. The `Batch` form is the previous struct unchanged, so the bulk path keeps moving its batches as before. The undo log itself is a `SmallVec<[IndexUndo; 2]>` and `PublishHold` keeps up to four guards and stores inline, growing past that exactly as the `Vec` did, so a commit of up to four tables, each touching up to two indexes, allocates nothing for the protocol. `undo_index_updates` replays a `Row` entry with the single-entry index calls and a `Batch` entry exactly as before.

Measured with a counting allocator, in memory, prepared statements. The first table is the guard's shape, a three-column table with only its primary key; the rest are from the 53-shape benchmark suite, whose `users` table carries secondary indexes and so recorded more undo entries per statement.

| statement | allocations, main | allocations, this branch | latency |
|---|---:|---:|---:|
| `INSERT` into the PK-only table | 20.0 | 13.0 | |
| INSERT single, suite | 32.0 | 18.0 | -13.0% |
| DELETE by ID, suite | 22.1 | 14.1 | -15.0% |
| UPDATE by ID, suite | 17.0 | 15.0 | -4.9% |
| Batch INSERT (100 rows), suite | 1150.9 | 1148.9 | -1.0% |
| COMMIT of one row into each of three tables | 20.1 | 3.1 | |

Gate: the 53-shape benchmark suite, 8 runs per build in alternating order, two independent runs on an idle machine against main b874ab38. Median latency delta -1.0 and +0.6 percent; allocations per operation equal or lower on every shape in both runs. Five read-only shapes crossed 2 percent in both runs (HAVING complex, NOT EXISTS, GROUP BY aggregation, OR conditions, JOIN+GROUP+HAVING, 2 to 6 percent), so I measured three of them directly with 6-second tight loops of the suite's own statements, four alternating pairs per binary: GROUP BY aggregation 42.3 to 41.7 us, HAVING complex 82.9 to 82.0 us, OR conditions 3.44 to 3.46 us. Equal or faster, so the suite flags are its noise on those shapes.

Cost stated: `TransactionVersionStore` grows from 224 to 376 bytes because the undo log keeps two entries inline; it is created per statement and the read probes above show no effect. `PublishHold` grows from 48 to 80 bytes; it lives on the commit stack.

Tests: `tests/commit_undo_allocs_test.rs` installs a counting global allocator and asserts that a prepared single-row `INSERT` on a PK-only table allocates at most 16 times (main fails at 20) and that a `COMMIT` of one row into each of three tables allocates at most 4 times (main fails at 20, the first revision of this branch at 7). It serializes on a static mutex because the counter is process-wide, and the target passes under both `cargo nextest` and `cargo test`. The file is compiled out under `test-filedb`, where a memory DSN opens a file database whose commit also writes the WAL and the prepared `INSERT` measures 30. The undo path itself is exercised by the existing atomicity and failpoint targets, which pass unchanged: `statement_atomicity_test`, `returning_atomicity_test`, `failpoint_io_test`, `captured_version_payload_test`, plus `durability_test`, `volume_checkpoint_test` and `upsert_cold_dupe_test`. Formatting and all-target, all-feature clippy with warnings denied pass; the full in-memory suite and the full `test-filedb` suite pass on the head.

Not changed: the bulk index update path and its rollback, the WAL and arena parts of commit, and the amount of work the undo does when it actually runs. The guard's bound of 16 is a deliberate tripwire against the commit path growing per-statement allocations again.


semihalev:
Both findings reproduced and fixed in 6f4f0640.

`test-filedb`: the prepared `INSERT` measures 30 there, as reported, because the memory DSN becomes a file database whose commit also writes the WAL. The guard file is now compiled out under that feature with a crate-level `cfg`, so the whole file goes, not just the one test, and clippy with all features stays clean. The target reports zero tests under `--features test-filedb`.

`PublishHold`: confirmed with a counting allocator. A `COMMIT` of one row into each of three tables made 7 allocations with the hold at inline capacity one (two spills per vector, 1 to 2 to 4) against 20 on main. The hold now keeps four guards and four stores inline and grows past that as the `Vec` did, 4 to 8, so up to four tables cost nothing and larger commits match the old growth. The same commit now makes 3 allocations. A second guard pins it at 4; main fails it at 20 and the first revision of this branch at 7. The hold grows from 48 to 80 bytes on the commit stack.

Retention of read-only committed transactions: I am checking that separately and will fix it in its own PR, since it predates this change and the fix is in the commit branch of `transaction.rs`, not here.

PR #165, opened 2026-09-13
Refuse a volume whose rows are out of row id order instead of sorting ids alone

semihalev:
`VolumeBuilder::finish` turned an out-of-order input into a silently corrupt volume.

It checked that row ids were strictly ascending, which the manifest and table binary searches rely on, and when they were not it sorted the row id list by itself. The column buffers stayed in insertion order, so every row of such a volume was listed under another row's id. Rows added as (3, "c"), (1, "a"), (2, "b") produced a volume whose first row is listed under id 1 and carries (3, "c"). No producer trips this today: seal, compaction and snapshot recovery all add rows in ascending row id order. The next step of the storage plan orders sealed rows by a cluster key, so rows will reach the builder in a different order, and this fallback would have scrambled them without a sign. It has to go first.

`finish` now returns `Result<FrozenVolume>` and refuses rows out of order or repeated with an internal error naming the cause. `seal_rows` returns the same result; `seal_and_persist_opts`, the chunked seal, snapshot recovery and compaction propagate it, compaction through its existing `prepare_error` path so a failed prepare is handled like any other. The success path is unchanged: the same single pass over the ids, no reordering of anything. Test and unit-test call sites unwrap.

Not measured against the gate: the change adds a `?` on the seal path and nothing on any read or write statement path; the sorted-input check is the one that was already there.

Tests: `tests/volume_row_order_test.rs` builds a volume out of row id order and asserts it is refused; with the previous `finish` the same rows produced the mismatched payload above (the assertion on main read "row 0 is listed under id 1 but carries Row { Integer(3), Text("c") }"). It also pins that in-order rows keep their own payload and that a repeated row id is refused. Formatting, all-target all-feature clippy with warnings denied, the touched integration targets and the library unit tests pass; the full in-memory suite and the full `test-filedb` suite pass on the head.

Not changed: the ascending row id contract itself and the binary searches over `row_ids()`; those move together in the clustering step.


semihalev:
The chunk cleanup finding is fixed in the follow-up PR: building and writing a chunk is one fallible step there, so a chunk refused for its row order removes the earlier chunks' files as a failed write did, with a regression test for the later-chunk case.

PR #167, opened 2026-09-13
Compile the rest of an index join's ON clause once per plan

semihalev:
The index nested loop join compiled the rest of its ON clause on every execution.

Since #95 the operator asks each pair only the part of the ON clause the probe does not answer, which is right, but `execute_join_source_capped` built that filter per pass: it lowercased both column lists into fresh maps (`CompileContext::with_second_row`), flattened and requalified the clause, and compiled the program again, arriving at the same program as the execution before. On the benchmark's self join (`ON u1.age = u2.age AND u1.id < u2.id LIMIT 100`) that was about a third of the join's time and 71 of its allocations per query; the shape had gone from 7.4 us before #94 to 11.3 us on main, with the residual compile as the largest single piece.

The compiled program now lives on the plan's compiled slot as `CompiledExecution::JoinResidual`, next to the schema epoch, the join direction, the outer column list it reads by position and the key equality it leaves out. A pass whose inputs match takes the program and wraps it in a `JoinFilter` with the execution's parameters; anything else compiles and, if the slot is free or stale, stores it. A plan with two index joins keeps the first one's program instead of swapping them every execution. An ON or WHERE clause that read a subquery this execution holds that value, and so does a WHERE a statement run with an outer row context had that row written into, so such a plan is not kept. The outer column list is in the key because a nested join on the outer side can reorder its output when an index is created or dropped on its key, which changes no schema epoch. The slot reaches the join through the same plan reference the PK fast path uses, threaded from `execute_statement_inner` down to the capped join; nested selects and set operation branches pass none, and the fast paths treat the new variant like the others they do not own.

Gate: the 53-shape suite, 8 runs per build in alternating order on an idle machine, against main d347969f.

| shape | main us | this branch | allocations, main | this branch |
|---|---:|---:|---:|---:|
| Self JOIN (same age) | 11.02 | 8.73 (-20.8%) | 374 | 255 |
| INNER JOIN | 13.55 | 12.59 (-7.1%) | 412 | 398 |

Median latency delta -0.0 percent. Self JOIN is now under its allocation count at 5f811d03 (257), the build BENCHMARKS.md was measured on; its latency is still above the 7.2 us of that build, the rest being the per-pair program run that replaced the post-pass filter, which is a separate matter. CTE + JOIN was flagged for 12 more allocations; five runs of each binary put main alone between 1592 and 1730 on that shape (its CTE aggregates in parallel), so that is the shape's own spread. Read-only shapes flagged at 3 to 4 percent in this run (GROUP BY, window, index range) have identical allocation counts and are the ones every gate run flags in one direction or the other.

Tests: `tests/join_residual_cache_test.rs` pins the allocation count of a prepared self join on a three-column table at 250 (main 283, this branch 212), checks that the cached rest of the clause still filters every pair, that a parameter inside it is read on every execution, that the program is dropped after a schema change (the table recreated with its columns in another order), that it follows a nested join whose column order flips when an index on its key is created or dropped, and that a statement run with an outer row context reads that row on every execution; the last three answered wrongly on the first revision of this branch. The file is compiled out under `test-filedb`, where the memory DSN opens a file database whose join allocates differently (402 there). It counts allocations on the measuring thread only, so a background thread waking during a loaded run does not add to it. Formatting, all-target all-feature clippy with warnings denied and the file under both runners pass; the full in-memory suite and the full `test-filedb` suite pass on the head.

Not changed: the third residual build site in `try_grouped_index_join`, which builds once per statement already, and the per-pair evaluation itself.


semihalev:
Both reproduced and fixed in the second commit.

Nested join column order: the outer column list is part of the key now and is compared on the hit path, no allocation. The create-index and drop-index cases from the review are in the guard file; both answered `[]` on the first revision and `[(1, 10)]` now.

Outer row context: a plan run with an outer row keeps no program, since the row's values are written into the WHERE before the join splits it. The `parent.v = 10` then `20` case is in the guard file too; it returned `[1]` for 20 on the first revision.

Allocation count of the prepared self join unchanged at 212; gate and both suites rerun on the head.

PR #168, opened 2026-09-13
Ask the semi-join set for a NULL instead of walking it twice per query

semihalev:
The NOT EXISTS rewrite walked its semi-join set twice on every execution to decide its NULL handling.

#95 made the rewrite keep a NULL outer value and drop NULL members from the negated set, which is right. To decide that, `transform_exists_to_in_list` scanned the whole set for a NULL, and `extract_in_hashset_info` scanned it again when the index probe read the rewritten test back. The set itself is built once and cached across executions, so on the benchmark's NOT EXISTS (7,500 cancelled orders) those two walks were about a quarter of the query, which had gone from 21.9 us before #94 to 33.7 us on main. Every `Value::Null` compares equal and hashes alike, so one `contains` of a NULL value answers the same question in constant time; both places ask that instead. Nothing else changes: the set, the rewrite and the probe are as before.

Gate: the 53-shape suite, 8 runs per build in alternating order on an idle machine, against main d347969f.

| shape | main us | this branch | allocations |
|---|---:|---:|---:|
| NOT EXISTS subquery | 34.25 | 24.22 (-29.3%) | 81 / 81 |

Median latency delta +0.2 percent; allocation counts identical on every shape. The read shapes flagged at 3 to 6 percent (DELETE by ID, GROUP BY, UNION ALL, NOT IN subquery and others) do not run the changed code, which is reached only from the EXISTS rewrite and the OR arm of the index probe's set extraction, and have the same allocation counts; they are the shapes the gate flags in one direction or the other on every run.

Tests: `tests/not_exists_null_scan_test.rs` pins the answers the NULL handling exists for: a NULL inner value matches no parent, a NULL outer value is kept by NOT EXISTS, a negated literal list holding a NULL keeps nothing, a negated IN over a subquery whose result holds a NULL keeps nothing in the OR form the probe reads (the lookup answers true and the probe stands down), and an inner result of only NULLs keeps every outer row. These pass before and after, in memory and under `test-filedb`; the change is a constant-time replacement of a scan with no observable state, so the gate is its measurement. Formatting and all-target, all-feature clippy with warnings denied pass; the full in-memory suite and the full `test-filedb` suite pass on the first commit, and the second changes only this test file, which passes under both features.

Not changed: the plain form `id NOT IN (SELECT p_id FROM c)` with a NULL in the result returns rows on main and here alike, since the plain `InHashSet` arm of the probe has no NULL check; that predates this change and gets its own PR. Also not changed: the per-execution build of the integer probe set from the cached value set in `try_in_hashset_index_optimization`, which is the remaining cost of this shape on main and before #94 alike, and a candidate for its own step.


semihalev:
Added both in the follow-up commit: a negated IN over a subquery whose result holds a NULL, in the OR form the probe reads, returns nothing (the set lookup answers true and the probe stands down), and an inner result of only NULLs keeps every outer row through NOT EXISTS.

While writing the first one I found that the plain form, `id NOT IN (SELECT p_id FROM c)` with a NULL in the result, returns rows on main as well as here: the plain `InHashSet` arm of the probe has no NULL check, so it answers "not in {1}" for a primary key. That predates this change and is not touched by it; I left that assertion out of this file and will fix it in its own PR.

PR #169, opened 2026-09-13
Let the evaluator answer a NOT IN over a subquery result that holds a NULL

semihalev:
`id NOT IN (SELECT p_id FROM c)` on a primary key returned rows when the subquery result held a NULL.

Nothing is known to be outside a set that holds a NULL, so such a NOT IN keeps no row; SQLite and DuckDB answer with none, and so does this engine on a plain column, where the compiled evaluator decides it. On a primary key two probes answered instead. The IN-subquery index path collected the subquery's values, built an exclusion set from the integers among them and returned every key outside it, never asking whether a NULL was among the values. The set path that the LIMIT form takes read the negated `InHashSet` the same way. With `p` holding 1, 2, 3 and `c.p_id` holding 1 and NULL, both returned rows 2 and 3.

Both probes now stand down when the negated set holds a NULL and leave the answer to the evaluator, which already handles it (`NotInSet` carries a `has_null` flag). The IN-subquery path counts the NULL while it collects the values, and the IN-subquery cache keeps no list that holds one: the evaluator's own run of the subquery fills the same cache, and the negated form read that list back on its next execution and answered wrongly again. A cached list is NULL-free by that rule, so no execution walks a list to ask; the set path asks the value set for a NULL, a constant-time lookup since every NULL compares equal and hashes alike. A subquery whose result holds a NULL is re-run by each execution that needs it, as the price of that rule. Found while adding review tests to #168; independent of it.

Two more routes, from the review of this branch. The IN-to-semi-join rewrite (taken when the NOT IN sits beside another predicate, `id NOT IN (SELECT p_id FROM c) OR EXISTS (...)`) collected the subquery's values through the EXISTS collector, which dropped NULLs, so it kept every row, and the matching DELETE deleted them all. The collector keeps NULL members now: a negated IN over a set holding one is answered with false before the rewrite, and the EXISTS rewrite reads such a member as matching nothing, which its NULL handling from #95 was written for and had never reached. The same rewrite rebuilt the subquery as a plain scan of its column, so `id NOT IN (SELECT p_id FROM c ORDER BY id LIMIT 1)` beside an EXISTS was judged against every `p_id`, NULL included, where the written subquery returns a single 1. The extraction now declines a subquery with a LIMIT or OFFSET, a DISTINCT ON, a grouping, a set operation, a WITH or an AS OF, which take the general IN-subquery way and run as written; on the SELECT path a subquery the rewrite refused was then handed to the compiler untouched, which the DML paths already avoided, and it is read once there now as they do. And a positive IN through a secondary index looked NULL members up as keys, so a row whose column is NULL matched `k IN (SELECT p_id FROM c)` when the result held a NULL; the three index probe loops skip a NULL member now, as the primary key probe already did.

Tests: `tests/not_in_subquery_null_test.rs` pins the primary key form with and without a LIMIT (each fails on main, and each guard needs its own check: removing either one brings its form back), the same form after the IN form ran and filled the cache (which failed on the first revision of this branch), the plain column form, the same set without a NULL keeping the rest, IN over a set holding a NULL still answering, the NOT IN beside an EXISTS with and without a LIMIT, the positive IN through a secondary index with a NULL row with and without a LIMIT, the NOT IN over a subquery with `ORDER BY id LIMIT 1`, with `LIMIT 0`, and its positive form, and over a `DISTINCT ON` and an `AS OF` subquery, each beside an EXISTS; those failed on an earlier revision of this branch. Formatting, all-target all-feature clippy with warnings denied and the file under both runners and under `test-filedb` pass; the full in-memory suite and the full `test-filedb` suite pass on the head.

Gate: the 53-shape suite, 8 runs per build in alternating order on an idle machine, against main e8c1dbc7, since the semi-join collector and the index probes serve the EXISTS, NOT EXISTS, IN and NOT IN subquery shapes: 

| shape | main us | this branch | allocations |
|---|---:|---:|---:|
| NOT EXISTS subquery | 24.65 | 24.63 | 81 / 81 |
| EXISTS subquery | 2.44 | 2.40 | 48 / 48 |
| IN subquery | 93.58 | 96.28 (+2.9%) | 47.2 / 47.2 |
| NOT IN subquery | 69.52 | 70.94 (+2.0%) | 58.0 / 57.9 |

Median latency delta +0.2 percent, allocation counts identical. The two flagged shapes probe the primary key, which the changed loops do not serve, so I measured them directly: 5-second tight loops of the suite's own statements, three alternating main/branch/branch/main rounds each, gave IN subquery 58.5 against 58.8 us and NOT IN subquery 31.0 against 31.2 us, equal within one percent. Both suites pass on the head.


semihalev:
Both reproduced and fixed in the second commit.

Semi-join rewrite: the collector keeps NULL members now. A negated IN over a set that holds one is answered with false before the rewrite, and the EXISTS rewrite reads such a member as matching nothing, which is what its NULL handling from #95 was written for; that branch had never been reached because the producer stripped NULLs first. The `OR EXISTS` case is in the guard file with and without a LIMIT; it returned all three rows on the first revision.

Secondary indexes: the three index probe loops skip a NULL member, as the primary key probe did. The `k IN (SELECT p_id FROM c)` case with a NULL row and an index on `k` is in the guard file with and without a LIMIT; it returned `[1, 4]` on the first revision.

Gate against main and both suites rerun on the head.

semihalev:
Reproduced and fixed in the third commit. The IN-to-semi-join extraction now declines a subquery whose result the rebuilt scan could not reproduce: one with a LIMIT or OFFSET, a GROUP BY or HAVING, a set operation or a WITH. Such a NOT IN takes the general IN-subquery path, which runs the subquery as written and, since the first commit here, sees its NULLs. On the SELECT path a subquery the rewrite refused was then handed to the compiler untouched ("IN over something that is not a list"), which the DML paths already avoided; it is read once there now as they do. The guard file has the `ORDER BY id LIMIT 1` form, the `LIMIT 0` form and the positive `ORDER BY id DESC LIMIT 1` form beside an EXISTS; the first two answered `[1]` and the third answered wrongly too on the second revision.

semihalev:
Both reproduced and fixed in the fourth commit: the extraction declines a `DISTINCT ON` subquery and a table source with `AS OF`, next to the LIMIT, OFFSET, grouping, set operation and WITH cases; they run as written. Both forms are in the guard file beside an EXISTS and answered `[1]` on the third revision.

PR #170, opened 2026-09-13
Locate a cold row by id whatever order its volume holds the rows in

semihalev:
First step of clustering: a sealed volume no longer has to hold its rows in ascending row id order to be looked up by id.

Every lookup of a row id in a sealed volume binary searched the volume's id list, and a segment's min and max row id were its first and last entry, so the whole cold side assumed ascending ids. Clustering seals rows in key order, so that assumption has to go before the first clustered volume exists; #165 already made the builder refuse an unordered list instead of sorting the ids away from their payload.

`FrozenVolume::locate(row_id)` answers a row id in either case: a binary search over the ids when they ascend, otherwise over a permutation of positions sorted by id, built on first use and kept in the volume's metadata so its warm and cold forms share it. The eleven lookup sites go through it (seven in the segment manager, the two segment walks and the checked lookup in the segmented table). A segment's min and max are now the smallest and largest id it holds, and the auto-increment bound taken after a reopen reads the largest id rather than the last. The builder accepts rows in the producer's own order only after `allow_any_row_order`, and then refuses a repeated id; every current producer (seal, compaction, snapshot recovery) still adds rows ascending, so for them `locate` does what the old code did: a two-compare range reject and one binary search, with the metadata answering "ascending" from the build without a scan.

The ordered primary key scan, which merges the hot rows with each volume's id list, walks a volume through the same permutation when it has one, keeping the physical position for the payload and the visibility bitmap; on the first revision it read the id list as stored and answered `ORDER BY id LIMIT 1` with the first row added. A segment's bounds come from `id_bounds`, the first and last id in either order, so registration under the seal fence stays a constant-time read rather than two passes over the ids. A reload from disk inherits the order its earlier form had decided, so it is not decided again, and the metadata size counts the permutation, four bytes per row.

No `CLUSTER BY` yet: that is the next step, on top of this one.

Gate: the 53-shape suite, 8 runs per build in alternating order on an idle machine, against main be280540, rerun on the second revision: median latency delta -1.3 percent, allocation counts identical or lower on every shape. The suite runs in memory and never seals, so it does not reach the changed lookups; it shows that the hot paths are untouched. The shapes it flagged (GROUP BY 3.0, CASE 2.5, JOIN+GROUP+HAVING 2.5, CTEs 2.1, DELETE complex 2.1 percent) have identical allocation counts and are the ones every run flags in one direction or the other. The cold lookup itself keeps its shape for ascending volumes: the same two-compare range reject and one binary search, with the "ascending" answer stored at build time.

Tests: `cold_rows_added_in_any_order_are_found_by_id` in the segmented table's unit tests registers a volume built as ids 3, 1, 2 and finds each row by id with its own payload, misses an absent id, and deletes one by id through the table; the same test walks the ordered primary key scan both ways and gets 1, 2, 3 and 3, 2, 1 (the first revision gave 3, 1, 2); `tests/volume_row_order_test.rs` pins that a producer that ordered its rows gets them located by id, may not repeat an id, and has its permutation counted in the metadata size, next to the existing checks that an undeclared unordered list is refused and that a multi-volume seal cleans up. Formatting, all-target all-feature clippy with warnings denied, the volume unit tests and the touched integration targets pass; the full in-memory suite and the full `test-filedb` suite pass on the head.

Not changed: scans, which walk a volume in physical order and never needed the ids sorted; the compaction sort by row id, which the next step replaces for clustered tables; the on-disk format, which stores the ids as given.


semihalev:
All four addressed in the second commit.

Ordered PK scan: each volume source in the merge carries the permutation when it has one and maps its cursor through it, so the id comparisons run in id order while the payload and visibility bitmap keep the physical position. The unit test now walks the scan both ways on the 3, 1, 2 volume and gets 1, 2, 3 and 3, 2, 1; it gave 3, 1, 2 on the first revision.

Reload: both reload paths call `inherit_row_order` on the loaded volume, copying the decided order from the form they replace, next to the unique index transfer they already did.

Bounds under the fence: `id_bounds` reads the first and last id, through the permutation when there is one, so registration is a constant-time read for volumes that come out of `finish`, whose order is decided at build; the two min/max passes are gone.

Memory accounting: `memory_size` counts four bytes per row for the permutation; a test pins the difference against the same rows added ascending.

PR #171, opened 2026-09-14
Let a table declare CLUSTER BY and seal and compact its rows in that order

semihalev:
Second step of clustering: a table can declare the order its sealed rows are kept in.

`CREATE TABLE t (...) CLUSTER BY (exchange, symbol, time)` names columns of the table; the key is resolved to column indices at creation (an unknown or repeated column is an error, an empty list is a parse error), kept on the schema, written after the defaults section of the serialized schema so a reader without it stops where it used to, read back on reopen, and shown by `SHOW CREATE TABLE`. `CLUSTER` is not a reserved word, so it stays usable as a name.

The seal sorts the rows it collected by that key, then by row id, before building volumes, and `seal_rows` lets the builder keep that order (#170 gave lookups their permutation; #165 refuses an undeclared unordered list). Compaction orders the live rows it merges the same way, reading the key columns through each input volume's column mapping, so a rewrite of clustered volumes keeps them clustered; an unclustered table sorts by row id as before. Snapshot recovery still builds its volume in row id order, which the lookups accept; the next seal or compaction of those rows puts them in key order. Every current path stays as it was for a table without a key: the sort branch is not taken, and `finish` keeps refusing an unordered list.

From the review, in the second commit: a dropped column renumbered the ones after it and left the key's indices behind (the key then named another column, or one past the end, which `SHOW CREATE TABLE` tripped on and a reopen refused as corrupt); a key column cannot be dropped now and the key follows the columns that move down. The snapshot codec did not carry the key, so `PRAGMA SNAPSHOT` then `PRAGMA RESTORE` silently unclustered the table; it writes and reads the same trailing section. Dropping a referenced parent rebuilt the child's schema without its key; the rebuild keeps it. A column whose type changed after rows were written holds more than one kind of value and comparing across kinds by value is not transitive, so the sort decides by kind first (NULL, booleans, numbers, text, timestamps) and by value within a kind; JSON and VECTOR columns have no order and are refused in a key. Compaction holds one column handle per key column per input volume and compares key cells in place through `ColumnData::compare_cells`, reading the typed arrays and dictionary strings where they are, so the only per-row work of the ordering is the index sort; a column a volume predates is its default for every row. The engine's schema validation checks the key wherever a schema enters (each column exists, appears once, has an order), so a key built through the schema API is refused at creation rather than on the next reopen, and the rule that a key column keeps a type with an order lives in `Schema::modify_column`, the one method every path changes a column through, so neither the statement nor the table API can turn a key column into JSON or VECTOR. Two NULL cells are equal before either is read (a key column holding only NULLs has an empty dictionary), integers and floats compare exactly through `cmp_i64_f64`, and a cell against a column a volume predates is compared in place against the precomputed default.

Docs: `docs/_docs/sql-commands/schema-management.md` gains a Clustered Tables section (syntax, what it buys, the type and DROP COLUMN rules, SHOW CREATE TABLE, snapshots) and `docs/_docs/architecture/storage-engine.md` says what order a volume holds its rows in and that the seal and compaction keep it.

Not in this step: `ALTER TABLE ... CLUSTER BY`, the window key, group-local dictionaries and variable groups; the query side reads the clustered volumes through the zone maps it already has, and the live measurement (next step) says what more they need.

Gate: the 53-shape suite, 8 runs per build in alternating order on an idle machine, against main 057c619a: two runs, median latency delta -0.2 and +0.2 percent, allocation counts identical on every shape (CTE + JOIN's count moves on its own, it aggregates in parallel). Both runs flagged a similar set of short read shapes 3 to 7 percent slower (index range and exact, COUNT DISTINCT, DISTINCT, derived table), none of which runs any changed code, so I measured three of them directly: 4-second tight loops of the suite's own statements, three alternating main/branch/branch/main rounds each.

| shape | main | this branch |
|---|---:|---:|
| SELECT by index (range) | 4.96 ns per row | 4.95 ns per row |
| SELECT by index (exact) | 1.460 us | 1.435 us |
| COUNT DISTINCT | 0.367 us | 0.380 us |

Equal within the noise of a sub-microsecond loop; the suite's flags on these shapes are its own spread, which the calibration notes put at 5 to 10 percent for them. The second commit was gated again at idle: median +0.3 percent, deltas between -5.1 and +6.1 on the same short shapes, allocation counts identical except the two shapes whose counts move on their own; the third commit likewise, median +0.7 percent, deltas between -5.3 and +5.9; the fourth, median +0.5 percent, deltas between -4.4 and +9.7 (the largest on SELECT by ID, a sub-microsecond shape with an identical allocation count).

Tests: `tests/cluster_by_test.rs` pins the clause (parsed, kept, shown), the errors, a sealed volume holding key order (`seal_rows` on a clustered schema keeps ids 3, 1, 2 and locates each), and a file database with 3,000 rows whose key order is the reverse of their id order: after `PRAGMA CHECKPOINT` the sealed volume file, read back from disk, holds its 3,000 rows in key order with 2999 first (where an unclustered table gives 1), while point lookups, the ordered PK scan both ways, a key-prefix query, an update and a delete by id all answer, and the key, the update and the delete survive a reopen. A fifth test seals three volumes of 1,000 rows each and lets a compaction with `compact_threshold=2` merge them: the single merged volume file, read back from disk, holds all 3,000 rows in key order, and lookups by id still answer; this test found two things: the compaction path built its volumes without declaring the order and was refused by #165's check, and it took a merged volume's bounds from the first and last reference of its chunk, which in key order are not the smallest and largest id, so a lookup by id was pruned away; both are fixed here (the bounds now come from `id_bounds`). Further tests pin that a key column cannot be dropped and that the key follows a dropped preceding column across a reopen, that a snapshot restore keeps the key, that dropping a referenced parent keeps the child's key, that JSON and VECTOR key columns are refused, that a key column holding integers, text and a NULL orders by kind then value, that a key built through the schema API past the last column, twice, or on a JSON column is refused when the table is created, that a key column cannot become JSON while a non-key column can and a key column can become TEXT, through the statement and through the table API, that compacting volumes whose keys are all NULL works, and that key cells compare integers and floats exactly and two NULL text cells as equal. Formatting, all-target all-feature clippy with warnings denied, the parser, volume and engine unit tests and the file under both runners and under `test-filedb` pass; the full in-memory suite and the full `test-filedb` suite pass on the head.


semihalev:
All seven addressed in the second commit; each of the five behavioural findings has a guard that fails on the first revision.

1. DROP COLUMN: `Schema::remove_column` refuses a key column and moves the key's indices down past a removed one. Test drops `a` from `(id, a, k, z) CLUSTER BY (k)`, checks the key still names `k` (index 1) in the schema, in `SHOW CREATE TABLE` and after a reopen, and that dropping `k` is an error.
2. Snapshot: the snapshot codec writes and reads the same trailing section as the WAL schema. Test: `PRAGMA SNAPSHOT` then `PRAGMA RESTORE` keeps the key.
3. Parent drop: `strip_fk_references` carries the key into the rebuilt child schema. Test: `DROP TABLE p` with a clustered child keeps `[2]`.
4. Comparator: values order by kind first (NULL, booleans, numbers, text, timestamps) and by value within a kind, which is a total order over the mixed representations a type change leaves behind. Test sorts `'15'`, `10`, `'a'`, `2`, NULL and expects NULL, 2, 10, '15', 'a'.
5. Unorderable types: JSON and VECTOR are refused in a key at creation.
6. Compaction: one `Vec<Value>` per key column over the live rows plus one index vector, sorted by index; no per-row vector.
7. Tests: the physical order tests read the sealed `.vol` files back through `read_volume_from_disk` and check ids and key order directly, both after a seal and after the compaction merge.

semihalev:
All three addressed in the third commit.

1. Engine boundary: `validate_schema` checks the key wherever a schema enters the engine (each column exists, appears once, has an order), so `SchemaBuilder::cluster_by(vec![1])` on a one-column table is refused at `create_table`, not on the next reopen. Guards for past-the-end, repeated and JSON keys through the builder.
2. MODIFY COLUMN: the statement refuses a key column becoming JSON or VECTOR, and the engine's own `modify_column` and `modify_column_with_dimensions` check the same, so the schema API and WAL replay cannot bypass it. Guard: `MODIFY COLUMN k JSON` is an error, `v JSON` and `k TEXT` go through and the key stays.
3. Compaction: no Value per row any more. One `&ColumnData` per key column per input volume, and `ColumnData::compare_cells` orders two cells in place (typed arrays, dictionary strings by `&str`, kind then value, NULL first); the only per-row work is the index sort. A column a volume predates is compared through its default, the one place a Value is built.

Both new guards fail on the second revision.

semihalev:
All four addressed in the fourth commit.

1. Two NULL cells are equal before either payload is read; the three single-row NULL-key volumes with `compact_threshold=2` are a test now and merge without a panic.
2. Mixed numeric cells compare through the exact `cmp_i64_f64`, both in `compare_cells` and against a value; the 2^53 + 1 case is pinned.
3. The type rule lives in `Schema::modify_column`, which every path changes a column through; the statement-level and engine-level checks were duplicates and are gone. Guard: `tx.get_table("t")?.modify_column("k", Json, true)` is an error, `Text` goes through.
4. A cell against a Default source is compared in place through `compare_cell_with_value`, kind then value, against the precomputed default; no `get_value` in the comparator any more. Only two defaults against each other use value comparison.

The three new guards fail on the third revision.

PR #172, opened 2026-09-14
Let ALTER TABLE set a table's CLUSTER BY key

semihalev:
An existing table can now get its clustering key, or a new one, without being recreated.

`ALTER TABLE t CLUSTER BY (exchange, symbol, time)` resolves the columns against the table, hands the key to the schema, which checks it the way `CREATE TABLE` does (each column exists, appears once, has an order), sets it on the version store's schema and the engine's cache, bumps the schema epoch, and records it to the WAL as ALTER TABLE operation 6, so a reopen replays it. `SHOW CREATE TABLE` shows the new clause. `CLUSTER` is not a reserved word; the ALTER parser accepts it as an identifier where it expects an action. A second `ALTER TABLE ... CLUSTER BY` replaces the key.

The volumes sealed before the change keep their order and stay readable through the lookups #170 gave every volume; the next seal takes the key order, and a compaction that merges older volumes writes its output in key order. Nothing is rewritten at the moment of the ALTER. What this step does not do: compaction only picks volumes below the target size, with tombstones, or oversized, so a full, untouched old volume is never rewritten and stays in id order. The next step makes a volume not in the table's key order a compaction input on its own, so a checkpoint after the ALTER reclusters the table gradually; until then the ALTER declares the key and orders new data only.

The key validation moved from the engine into `Schema` (`validate_cluster_key`), where `set_cluster_key` and the engine's schema check both call it.

From the review, in the second commit: another ALTER could run between the key change and its WAL record, so the log replayed the two in the wrong order (a DROP COLUMN replayed before the key that allowed it was refused against the older key, and the column came back); one DDL statement at a time now holds the engine's DDL guard from its schema change to its WAL record. And a table handle applied every column change twice, to the version store's schema and to its own cached copy, so a key set elsewhere after the handle was taken left the copy behind and a change the live schema accepted was refused against the copy; the handle changes the live schema and takes it whole, and a new column's id comes from the live schema too.

Third commit: a checkpoint re-records every table's CREATE after the point it truncates the log to and read the catalog without the guard, so an ALTER landing between the read and the records replayed before the CREATE that carried the older key and was refused. The re-recording holds the DDL guard from its read of the catalog to its last record, and CREATE and DROP of tables, indexes and views take the guard as ALTER does.

Gate: the 53-shape suite, 8 runs per build in alternating order, against main c1076309: median latency delta +0.2 percent, allocation counts identical on every shape, on the one run of three that had the machine to itself (the other two, one under a system daemon at a full core, gave -0.7 and +2.5 with flags in both directions on shapes that run no changed code). The change touches ALTER TABLE and the table handle's column changes only; no read or write path runs any of it.

Tests, in `tests/cluster_by_test.rs`: a file database seals 1,000 rows before the ALTER and 1,000 after; read back from disk, the first volume is in id order and the second in key order; a third seal triggers a compaction that merges all three into one volume in key order, a lookup by id still answers, and the key survives a reopen through WAL replay. The error cases pin an unknown column, a repeated one, a JSON one and an empty list, and that a second ALTER replaces the key. A retained table handle follows a key change made elsewhere (modify and drop the freed column, refuse the new key column), and two threads interleaving 40 key changes with 40 add and drop column statements leave a schema that a reopen replays identically. Formatting, all-target all-feature clippy with warnings denied, the parser unit tests and the file under both runners and under `test-filedb` pass; the full in-memory suite and the full `test-filedb` suite pass on the head. A checkpoint started while a statement holds the DDL guard does not re-record the catalog until the guard is released; on the previous commit it completes at once.

Docs: the Clustered Tables section of `docs/_docs/sql-commands/schema-management.md` gains the ALTER form and what happens to the volumes already sealed; the sentence about older volumes will be sharpened with the next step.

Not in this step: removing a key (`CLUSTER BY` with no columns is an error), and reclustering the volumes that exist at the ALTER, which is the next step.


semihalev:
Both addressed in the second commit.

1. Ordering: one DDL statement at a time holds the engine's DDL guard (`ddl_guard`, a mutex on the engine) from its schema change to its WAL record, so the log replays the changes in the order they were applied. A test runs two threads interleaving 40 `CLUSTER BY` changes with 40 add and drop column statements and checks that a reopen replays the schema the statements left.
2. Retained handles: the table handle no longer applies a mutation to its cached copy a second time; it changes the version store's schema and takes it whole, and a new column's id comes from the live schema as well. The review's scenario is a test: key (a), handle taken, key moved to (b) elsewhere, then `modify_column(a, Json)` and `drop_column(a)` go through with both copies agreeing, and `modify_column(b, Json)` is refused.

semihalev:
Addressed in the third commit. `rerecord_ddl_to_wal` holds the DDL guard from its read of the catalog to its last record, so an ALTER either completes before the read (and the re-recorded CREATE carries its key) or records after the re-recorded entries. The same holds for CREATE and DROP of tables, indexes and views, which now take the guard as ALTER does; a CREATE INDEX landing in that window would have replayed before its table the same way. The guard is taken by PRAGMA CHECKPOINT and the background cycle alike, outside the seal fence, and no DDL path runs a checkpoint, so there is no cycle. Guard test: a checkpoint started while a statement holds the guard does not finish until the guard is released; on the previous commit it finishes at once.

semihalev:
Gate on the current head (786a5c2b), the 53-shape suite, 8 runs per build in alternating order, against main c1076309, on an idle machine on mains power: median latency delta +0.5 percent, deltas between -10.5 and +7.4, allocation counts identical on every fixed-count shape. The flagged shapes (LEFT JOIN + GROUP BY +7.4, INNER JOIN +5.8, GROUP BY +5.6, DELETE by ID +5.3) run none of the changed code, which is the DDL statements and the checkpoint's catalog re-recording; the same shapes moved the other way on the two earlier runs of this branch (median -0.7 and +0.2), so the spread is the suite's own. The earlier runs of the day were on a drained battery and are not comparable in absolute numbers.

PR #173, opened 2026-09-14
Recluster the volumes sealed before a table's key through compaction

semihalev:
The volumes that exist when a table gets its key are reclustered by compaction, one pass over each.

Compaction picked only volumes below the target size, with tombstones, or oversized, so a full, untouched volume sealed before `ALTER TABLE ... CLUSTER BY` stayed in id order for good, and the key ordered new data only. On the live table that is every row.

A volume not in its table's key order is a compaction input on its own now. Whether it is in order is decided once per volume by one typed pass over the key columns (`FrozenVolume::in_key_order`, through `compare_cells`, about a millisecond per million rows), read through the volume's column mapping, so a column dropped since the seal or added after it is resolved the way compaction resolves it: a dropped earlier column no longer shifts the check onto the wrong columns, and a key column the volume predates holds one value for every row and orders nothing. The verdict is kept on the segment manager since volumes are immutable, leaves with its segment, and the volumes compaction itself writes are recorded in order without a pass. A cold volume is loaded to decide, and a cycle decides at most `compact_threshold` volumes, so what one cycle loads and rewrites is bounded; a large table is reclustered over as many cycles as that takes. A single such volume is rewritten alone, where a single small volume would otherwise wait for company.

Compaction took the copy of a row from the volume with the highest segment id, while readers take it from the volume latest in the manifest and a rewrite takes the position of the batch's first member; a rewritten older volume therefore had a higher id than an untouched newer one, and the next compaction that held both let the older copy win. Compaction dedups in manifest order now, and a batch is closed over overlap: a volume left between two members that shares a row with the batch joins it, and one deferred behind a snapshot holds the table for the cycle. A cycle selects at most `compact_threshold` volumes for the rewrite whether their verdict was decided now or kept from an earlier cycle, and the closure counts against the same limit: while the batch holds more volumes beyond the ones size or tombstones call for than the threshold, the selection loses its last volume and the closure is taken again; a last selection that still exceeds the limit beside base work goes alone and the base work waits for the next cycle, since a lone member closes over nothing (the volumes on either side keep their places relative to its rewrite) while base work that is a single small volume would wait for company for good (`close_batch_over_overlap`, unit tested on its own).

The check runs while a recluster request is open. Requests are counted: every manager starts with one, setting a key opens another, and a cycle closes the request it read before it read the schema, once it found every volume in order with none left undecided or deferred behind a snapshot. A key set while a cycle runs therefore stays pending for the next one. Planning reads the manifest for metadata and releases it before any volume is read, since a seal holds the table's fence while it waits for the manifest and DML waits on the fence. Nothing changes at the moment of the ALTER; the checkpoint cycles that follow do the rewrite through the compaction machinery that exists, outside the seal fence like any compaction.

The next-seal test of #172 changes with this: after the ALTER and a checkpoint, the volume sealed before the key is in key order too, not only the new one.

Gate, on the first revision of this branch: the 53-shape suite, 8 runs per build in alternating order on an idle machine, against main c1076309 (this branch carries #172 as well): median latency delta +0.5 percent, deltas between -6.4 and +13.0, allocation counts identical on every shape. The +13 was SELECT by index (range), which runs no changed code, so I measured it directly: 4-second tight loops of the suite's statement, three alternating main/branch/branch/main rounds, 5.00 against 5.08 ns per row for the range and 1.564 against 1.479 us for the exact form. Equal within the noise of the loop; the suite's flag is its own spread on that shape, as in the two previous runs that flagged it. The gate on the current head follows in a comment once the machine is idle enough to measure.

Tests, in `tests/cluster_by_test.rs`: a volume sealed before the ALTER is rewritten in key order by the next checkpoint (read back from disk; the file set changes) and left alone by the checkpoint after (the file set does not); an unclustered table's full volume is left alone across checkpoints; a key declared, the database closed without a checkpoint and reopened, reclusters the older volume on the next checkpoint; a column dropped before the key is declared does not hide the older volume from the rewrite; a key holding a column added after the seal reclusters the older volume instead of failing the checkpoint; three at-target volumes with a threshold of two are rewritten two in the first cycle and one in the second, with the third cycle touching nothing; and a row updated after its volume was sealed keeps its newer value through a batch that rewrites the older copy and a compaction that then holds both copies. A volume out of order tied by a chain of overlapping ordered volumes to a dirty volume at the far end is rewritten on the first cycle rather than waiting behind the dirty work. Unit tests in `manifest.rs` pin the request counting (a request opened during a check stays open), the check through the column mapping, and that verdicts leave with their segments. The two recluster tests and the changed next-seal test fail on #172's sources, the dropped-column and added-column tests on the first revision of this branch, the newer-copy test on the second (the value goes back to the older copy), and the tied-selection test on the fifth (the volume out of order is never rewritten); the unclustered-table guard passes on both, as it should. Formatting, all-target all-feature clippy with warnings denied, the volume unit tests and the file under both runners and under `test-filedb` pass; the full in-memory suite and the full `test-filedb` suite pass on the head.

Docs: the Clustered Tables section says what happens to the older volumes now.

Not in this step: reclustering while a snapshot older than the volumes is open, where the existing gate defers the rewrite as it does every compaction and keeps the request open for the cycle after the snapshot closes; that window cannot be produced through the public API (the seal's cutoff and compaction's floor come from the same registry), so it has no guard test. The merge of many sub-target volumes in one cycle is unbounded as it was before this branch.


semihalev:
All six addressed in the second commit.

1. The key is resolved through the volume's column mapping before the check (`SegmentManager::volume_key`): a dropped earlier column maps to the volume's own column, and a column the volume predates is one value for every row and is left out. Verdicts are keyed by the volume columns checked. Two guard tests: a column dropped before the key hid the volume from the rewrite; a key with a column added after the seal failed the checkpoint. A unit test pins the mapping.
2. Planning collects (segment id, row count, rewrite) under the manifest guard and drops it before any verdict is decided or volume loaded; the volumes to rewrite are fetched by id afterwards.
3. A cycle decides at most `compact_threshold` volumes (a decision is the one place a cold volume is loaded), and only the volumes decided out of order are rewritten, so what one cycle loads and rewrites is bounded by the threshold. The volumes compaction writes are recorded in order, so they cost no decision later. A test with three at-target volumes and a threshold of two sees two rewritten in the first cycle, one in the second, nothing in the third. The merge of many sub-target volumes in one cycle is as unbounded as before this branch; not changed here.
4. A deferred volume, and a volume left undecided by the bound, both keep the request open (`all_checked`); the request closes only when every volume was found in order. The bound test covers the undecided path; the deferral window cannot be produced through the public API, so it has none of its own.
5. Requests are counted (`recluster_request`, `request_recluster`, `recluster_done`): compaction reads the open request before it reads the schema and closes that request, so a key set during the pass leaves a newer request open. Unit test: a request opened during a check stays open, a stale close changes nothing.
6. Verdicts are removed in the four removal and replacement paths and cleared with the manager; a unit test checks each path.

The docs say a cycle reclusters at most `compact_threshold` volumes.

semihalev:
Both addressed in the third commit.

1. The precedence rule: readers take a row from the volume latest in the manifest, and a rewrite takes the position of the batch's first member, so compaction now dedups in manifest order instead of by segment id. That alone leaves the case of a volume between two members of a batch that shares a row with them, so a batch is closed over overlap (every such volume joins it; one deferred behind a snapshot holds the table for the cycle). Guard test: a row updated after its volume sealed, the older volume reclustered in a batch that leaves the newer one behind, then a tombstone and the newer volume's turn put both copies into one compaction; on the previous commit the value goes back to the older copy.
2. Selection counts against the threshold as decisions do, kept verdicts included, so a failed rewrite does not grow the next cycle's batch. No guard test: the write failure cannot be produced through the public API.

The same id-based precedence was reachable before this branch through the sub-target merge (two small volumes merged around an untouched larger one that held a newer copy); the fix covers that path too.

semihalev:
Addressed in the fourth commit. The closure and the bound live in one function now (`close_batch_over_overlap`): after the closure, while the batch holds more volumes beyond the base (the ones size or tombstones call for) than the threshold, the selection loses its last volume and the closure is taken again. The floor is one selection: a single volume's overlap has to be rewritten with it or the precedence bug returns, so a lone volume that shares rows with several others still rewrites them all. A reduced selection keeps the request open. Unit tests cover the three cases: two selections around an overlapping ordered volume at threshold two (reduced to one, the middle left alone), the same at threshold three (all three), and one selection whose overlap exceeds the threshold (kept), plus the deferred-overlap hold. No integration guard: the scenario needs a failed rewrite to leave kept verdicts behind.

semihalev:
Gate on the current head (1e3ded98), the 53-shape suite, 8 runs per build in alternating order, against main c1076309, on an idle machine on mains power: median latency delta +0.2 percent, deltas between -4.6 and +5.0, allocation counts identical on every fixed-count shape (the one flag is NOT IN subquery at 57.90 against 58.00, a shape whose count varies run to run). The largest deltas are UPDATE complex (+5.0), IN subquery (+3.8) and NOT EXISTS subquery (+3.8), none of which runs compaction planning or the manifest's verdict code; the same shapes moved the other way on the earlier runs of this branch.

semihalev:
Addressed in the fifth commit. A last selection that still exceeds the limit beside base work is dropped from the cycle: the base work goes alone, the selection waits for a cycle of its own, and the request stays open. There is no floor left: a lone member closes over nothing, because the volumes on either side of it keep their places relative to its rewrite, so a cycle rewrites at most the threshold beyond the base work. Unit tests: one selection beside a distant dirty volume with an overlapping chain between them at threshold two leaves the batch at the dirty volume alone; a lone selection takes nothing with it.

semihalev:
Addressed in the sixth commit. When a last selection cannot join the base work within the limit, the selection goes alone and the base work waits for the next cycle, the other way round from before: a lone member closes over nothing, so the batch is within the limit, and base work is never held for good, whereas a single small base volume waits for company under the existing rule and would have held the selection indefinitely. Guard test, the scenario from the finding: one volume out of order, three ordered volumes each holding a newer copy of a row of the one before, a tombstone in the last one, threshold two; on the previous commit the out-of-order volume is still there after the first checkpoint, with the fix it is rewritten on that cycle and the table settles in the cycles after. The unit test for the case flipped accordingly.

semihalev:
Gate on the current head (ec55a770), the 53-shape suite, 8 runs per build in alternating order, against main c1076309, on mains power, two runs on an idle machine: median latency delta -0.1 and +0.2 percent, allocation counts identical on every fixed-count shape (the two alloc flags, Correlated in SELECT and DELETE complex, are shapes whose count varies run to run, and both moved the other way on earlier runs). Both runs flag the same three short shapes, SELECT by ID (+9.6, +11.4), SELECT by index exact (+19.4, +11.3) and index range (+5.0, +5.6), none of which runs any changed code, so I measured them directly: tight loops of the suite's statements, three rounds of main, branch, branch, main on an idle machine.

| shape | main, mean of 6 | branch, mean of 6 |
|---|---|---|
| SELECT by index (exact) | 1.594 us | 1.490 us |
| SELECT by index (range) | 9.36 us | 9.41 us |
| SELECT * by id, prepared | 0.1623 us | 0.1615 us |

Equal within the loop's own drift (every shape slowed by about 10 percent from the first round to the third on both builds). The suite's flags on these shapes are its own, as on the previous runs of this branch.

PR #175, opened 2026-09-14
Rule row groups out by their zone maps before the dictionary filters decode them

semihalev:
The scanner decided which row groups its zone maps rule out after the dictionary filters had already decoded the key columns of every group to list the matching rows, and the fast path over that list never consulted the decision. On a clustered volume a query for one series therefore decoded the key columns of all row groups to find the one or two that hold it. The decision is taken first now, in `prepare_filter`, and both dictionary passes (the compressed-store pass and the resident-column pass) leave the groups it rules out alone; the linear path already did. No new metadata and no format change: the zone maps a volume already carries, consulted earlier.

Second commit, from the review: the resident pass took chunks of a row group's length from wherever the range began and skipped a chunk by the group its start lay in, so a range starting inside a pruned group lost the start of the next one; a chunk ends at its group's boundary now. And the filter's columns were resolved by name in the volume, which may still hold a dropped column of that name, so the zone maps read the old column's bounds once they were consulted first (the dictionary filter and the typed predicates read the old column the same way); they resolve through the column mapping now, which carries the schema's column names for that, and a column the volume does not hold is left to the full filter on the materialized row. The volume-level pruning in `prune_volume` resolves by name too and prunes a whole volume by a dropped column's bloom filter; that is older than this branch and is left to a follow-up.

Measured on the untracked HTAP harness (one volume of 1,044,000 rows in 16 row groups, 174 series arriving interleaved, 16 MiB group cache, one series' full history projected and aggregated; decode bytes and blocks from an instrumented build, latency from a native build, one run per cell):

| layout | | first cycle blocks | first cycle decoded | cache after | native first | native two repeats |
|---|---|---:|---:|---:|---:|---:|
| clustered | main | 39 | 14.57 MB | 14.57 MB | 11.4 ms | 4.9 ms |
| clustered | branch | 11 | 5.44 MB | 5.44 MB | 2.8 ms | 2.3 ms |
| unclustered | main | 342 | 170.1 MB | 16.4 MB (full) | 58.6 ms | 101.3 ms |
| unclustered | branch | 342 | 170.1 MB | 16.4 MB (full) | 54.9 ms | 102.2 ms |

The 11 blocks are the two key columns of the two groups the zone maps admit plus the columns of the group that holds the series; the second admitted group holds the end of one exchange and the start of the next, so its exchange bound and its symbol bound each admit the key although no row of it carries both. Independent per-column bounds do not carry the composite key's relation, which is why the result is 11 and not 7. The fixed-window query (first six hours of one series) goes from 39 blocks and 9.8 ms to 11 blocks and 2.1 ms on the clustered layout for the same reason. On the unclustered layout every group holds every key, nothing is ruled out, and the numbers do not move; the second commit leaves the clustered numbers where the first put them (11 blocks, 5.44 MB). The two repeated cycles on the clustered layout decode nothing on either build and still make 24,800 allocations, which is the next thing to look at, not this change.

Gate: the 53-shape suite, 8 runs per build in alternating order, against main fe2cf12b on an idle machine: median latency delta +0.3 percent on the first commit and +0.2 on the second, allocation counts identical on every fixed-count shape; the suite runs an in-memory database and never reaches this code.

Tests, in `tests/row_group_prune_test.rs` (a file database of twelve series in a clustered volume of six row groups, reopened so the columns are read a group at a time; the tests share the process-wide decode counter and run one at a time): a series held by one group decodes 7 blocks (15 without the change); a series straddling two groups is read whole, 12 blocks (18); a group whose independent bounds admit a key it does not hold is visited and the answer stands (15 before); an OR filter rules nothing out and returns the union; a filter matching half the rows takes the fallback path and still skips the other exchange's groups; a group of NULL keys is left out of an equality and counted by IS NULL; a range starting inside a pruned group on the columns as sealed keeps the 6,000 rows at the start of the next group; a column dropped and added back with a default is not pruned by the dropped column's bounds. The first four bounds fail on main, the last two on the first commit of this branch. Formatting, all-target all-feature clippy with warnings denied, the file under both runners and under `test-filedb`, the full in-memory suite and the full `test-filedb` suite pass.


semihalev:
Both fixed in the second commit.

1. The resident pass chunks the range by row group now: a chunk ends at its group's boundary, so a range starting inside a pruned group no longer carries the skip into the next group. Guard: your scenario, 196,608 rows on the same handle after the checkpoint, `id > 10000 AND symbol = 'b'` and the `n > 10000` form both return 6,000 (0 on the first commit).
2. The filter's columns resolve through the column mapping now; `ColumnMapping` carries the schema's column names for that, and a column the volume does not hold (`ColSource::Default`) is left to the full filter on the materialized row. This covers the zone maps and, since they resolve the same way, the dictionary filter and the typed predicates. Guard: your scenario, `x INTEGER` dropped and `x FLOAT DEFAULT 5.0` added, `k = 'needle' AND x >= 5.0` returns 10 rows with sum 50, and `x < 5.0` returns none.

Found on the way and left out: `prune_volume` in `volume/table.rs` resolves by name too, and a text column dropped and added back with a default (`k TEXT DEFAULT 'z'`, `WHERE k = 'z'`) prunes the whole volume by the dropped column's bloom filter, on main as well. It has thirteen call sites and predates this branch; follow-up PR.

PR #176, opened 2026-09-14
Resolve a filter's columns through the column mapping when pruning a whole volume

semihalev:
Stacked on #175 (base `perf/row-group-prune`); the follow-up named in its review.

The volume-level pruning in `prune_volume` resolved a filter's columns by name in the volume, which may still hold a dropped column of that name. A text column dropped and added back with a default therefore had the whole volume ruled out by the dropped column's bloom filter and bounds: `k TEXT` dropped, `k TEXT DEFAULT 'z'` added, `SELECT id FROM t WHERE k = 'z'` returned no rows on main for the sealed rows, which all carry `'z'` now. The pruning takes the segment's column mapping at each of its twelve call sites and resolves through it, with the same helper the scanner uses since #175 (`ColumnMapping::volume_column`): the identity mapping resolves by name, an evolved one through its sources, and a column the volume does not hold prunes nothing and is left to the full filter on the materialized row. The parallel-scan site that already skipped pruning for evolved volumes keeps its guard.

Second commit, from the review: two renames that swap two names leave every column where it was, so the mapping stayed the identity and carried no names, and a filter on a swapped name resolved by name in the volume to the column that used to carry it (on main as well). The mapping carries the names whenever they differ from the volume's by position, the scanner keeps such a mapping for resolving a filter's columns, and rows go through the mapping only when the columns are not in place.

Tests, in `tests/row_group_prune_test.rs`: a volume of 70,000 rows sealed with `k`, reopened, `k` dropped and added back with default `'z'`: `k = 'z'` returns every row (none on #175's head), `k = 'needle'` none, and `k = 'z' AND x > 69990` ten, so the mapped column and a real column prune together. A sealed table with `x` renamed to `z` and `y` to `x`: `x = 9` returns every row (none before), `x = 1` none, `z = 1` every row, `z = 9` none. Formatting, all-target all-feature clippy with warnings denied, the file under both runners and under `test-filedb`, the full in-memory suite and the full `test-filedb` suite pass.

Gate: the 53-shape suite, 8 runs per build in alternating order, against main fe2cf12b on an idle machine: median latency delta +0.2 percent on the first commit and -0.1 on the second, allocation counts identical on every fixed-count shape; the suite runs an in-memory database and never reaches this code (the flags, COUNT DISTINCT and the DISTINCT shapes at 7 to 8 percent, are the suite's short-shape spread).


semihalev:
The swapped-name case is fixed in the second commit rather than left for another PR, since it is the same guarantee: the mapping carries the schema's column names whenever they differ from the volume's by position, not only when the shape differs, so an identity-by-position mapping after two renames still resolves a filter's `x` to the column the schema means. The scanner keeps such a mapping for resolving filter columns and reads rows through it only when the columns are not in place (`row_mapping`), so the identity row path is unchanged. Guard: your scenario, `x` renamed to `z` and `y` to `x` on a sealed table, `WHERE x = 9` returns the 1,000 rows (none on the previous commit and on main), `x = 1` none, `z = 1` all, `z = 9` none.

semihalev:
Gate on the current head (497a380d), the 53-shape suite, 8 runs per build in alternating order, against main fe2cf12b on an idle machine on mains power: median latency delta -0.1 percent, allocation counts identical on every fixed-count shape. The flagged shapes (HAVING complex +5.6, Complex JOIN+GROUP+HAVING +5.4, NOT IN subquery +5.1) run an in-memory database and never reach the volume pruning; they moved the other way on the earlier runs of this branch. An earlier run of this head on a draining battery gave a uniform +6.1 with base latencies two to three times the usual and is discarded.

PR #177, opened 2026-09-14
Resolve every column a read path names in a sealed volume through the mapping, and prepare the cold collector's filter

semihalev:
Stacked on #176 (base `fix/prune-volume-mapping`); the four wrong results its second review found, plus two more on the same paths. All of them are on main.

Four read paths named a column of a sealed volume by its name in the volume, which may be a column renamed since the seal: the ordered `LIMIT` path took its bounds and stop key from it (`ORDER BY x ASC LIMIT 1` after `x` renamed to `z` and `y` to `x` returned the row with the smallest old `x`), the MIN and MAX shortcut took its extrema from it, the cold row collector took its dictionary filter from it, and the mapping itself followed only one rename record back, so a column renamed twice since the seal (a cycle of four renames) was lost when the mapping was computed again by a later `ADD COLUMN`. They resolve through the column mapping now (`ColumnMapping::volume_column`), the mapping follows the rename records until a name the volume holds (`renamed_source`), and the SQL `RENAME COLUMN` refreshes the schema cache before the mapping is recomputed, as `DROP COLUMN` already did, so the mapping is computed from the renamed schema rather than one statement behind.

Second commit, from the review: the rename records had no order against the drops, so a temporary name used twice, a name dropped and added back and renamed on, and a name freed by a drop and taken by a rename each resolved to the wrong column or to none, and the cases cannot be told apart from the records alone (a rename into a name and a drop of that name are the same records whichever came first). A rename now records its schema version as well, kept in the drop list under a name no column can carry (a NUL byte and the rename's ordinal in the rename list, so it is short whatever the names and pairs with its record by position), so the manifest format stays at 6 and an older reader never matches the marker against a column; a column's source is found by walking the changes made since the seal, newest first (`schema_events_since`, `source_before_changes`): a rename into the name takes the name it had before, a drop of the name ends the identity there and the column takes its default. A history whose renames carry no marker is read as it was before the markers, one rename record back with a dropped name resolving to nothing, since it has no order to walk. The plan's K7 (catalog identity that survives DDL) remains the permanent form; the markers go with it.

Third commit, from the review of the second: the schema epoch started over at zero on an open while the versions the manifests carry persisted, so a change made after the open ordered before every seal made before it (this also affected the drop masks, which compare the same versions); the epoch now restarts above every persisted version, before the log is replayed. A rename replayed from the log was recorded under the epoch before its increment where the statement records it after, so the same history took different versions on the two paths; the replay increments first now. And a change recorded in the version a later seal carries happened before that seal and is not walked back over its rows.

Fourth commit, from the review of the third: the seal read the schema before it built the volume and the schema version when it registered it, so a rename made in between gave a volume of the old names the rename's own version; the version is read under the lock the schemas are read under, and compaction reads it with the schema the same way. A rename or drop replayed from a log kept past the checkpoint that wrote the manifest (a snapshot holding the truncation back) was recorded in the manifest again under a newer version, so the volume sealed between the change and the checkpoint was read as if it predated the change; the manifest keeps the log position its renames and drops reach (another marker in the drop list, `\0lsn`), the statement notes its record's position there, and a replay at or below it changes the schema without recording the change again. And the loader merged every rename into a volume's own column names, which turned the names the history walks back to into other columns' slots after a reopen (on main as well, since 497a380d's names): a volume keeps the names it was sealed with and the mapping alone resolves the renamed ones.

Fifth commit, from the review of the fourth: a checkpoint whose truncation a snapshot held back could write a manifest out between a rename's manifest record and the note of its log position, so a reopen replayed the rename over a manifest that held it already; the manifest is written out under the DDL guard now (at the checkpoint's write and at compaction's, both outside the seal fence, and no DDL path writes a manifest), so it never shows a statement halfway, and the statement records itself in the log before the manifest takes the change together with that record's position. The schema epoch moved after the schema cache's write lock was released, so a seal could take the new schema with the old epoch; the epoch moves under the lock on every path that replaces a schema. And a volume sealed under an older schema was registered with the identity mapping until the next change recomputed it; a sealed volume's mapping is computed against the schema current at registration, and compaction recomputes its outputs' mappings when the epoch moved during the rewrite.

Sixth commit, from the review of the fifth: a sealed volume's mapping was computed from a schema cloned and released before the segment was published, and compaction published its outputs with the identity mapping and corrected them in a later step, so a change completing in between left a published segment with a mapping of the wrong shape until the next change. Publication holds the schema cache's read lock while the mapping is computed and the segment made visible, for the seal's registration and for compaction's replacement alike (`mapping_for`, taken under the manifest lock the publication holds): every change replaces the schema under that lock's write side and propagates its history after, so a publication is either wholly before a change, whose propagation then covers the new segment, or wholly after it. The post-publication correction is gone.

Seventh commit, from the review of the sixth: compaction recorded the order it wrote its outputs in by the key's positions in the rewrite schema, resolved through the mapping current at the time, so a column dropped ahead of the key during the rewrite moved the key's position and the record named another column as ordered; a later `CLUSTER BY` on that column could then skip its rewrite. The outputs are built in the rewrite schema's column order, so the key's positions in that schema are their physical columns; the record keeps those, and a lookup resolves the schema's key through the mapping to compare. A manifest unit test drops a column ahead of the key before the record is made and checks that the key is still the ordered column and the column that took its position is not.

Two more faults surfaced on the same paths while writing the guards. The cold row collector evaluated the filter it was handed as written, and a comparison rejects every row until it is prepared for the schema, so an anti-join over a sealed inner table with a non-correlated filter let every outer row through, renamed or not (`NOT EXISTS (SELECT 1 FROM r WHERE r.id = p.id AND r.x = 'old')` returned every `p` once `r` was sealed); the collector prepares the filter for the schema now, as the scanner path does. And the MIN and MAX shortcut answered from the sealed extrema while the transaction held writes of its own (`MAX(x) ... LIMIT 1` after an uncommitted insert of a larger value answered the sealed maximum); it declines when the table has local changes, as the aggregate pushdown does.

Tests, in `tests/column_rename_resolution_test.rs`: on a table of 70,000 sealed rows with `x` and `y` swapped by two renames, `ORDER BY x ASC LIMIT 1` and `DESC LIMIT 1` return the rows the renamed column orders; `MIN(x)` and `MAX(x)` with `LIMIT 1` inside a transaction that inserted a row return 1 and the inserted 80,000; the anti-join with `r.x = 'needle'` on the renamed text column returns no outer row and with `r.x = 'old'` returns it; a cycle of four renames followed by `ADD COLUMN` projects `(2, 3, 1, 0)` and matches all 1,000 rows on the three renamed columns; and an anti-join over a sealed inner table without any rename keeps the outer row that has no match and drops the one that has. All five fail on the first commit's predecessor. Four more, each read once after the history and once after a reopen so the order comes back from the manifest: a temporary name used twice for two columns gives each its own data; a name dropped, added back with a default and renamed on carries the default; a name freed by a drop and taken by a rename reads the renamed column; a name dropped and added back reads the default. The first three fail on the first commit; the fourth passes there as well and pins the walk's drop rule. Three more from the review of the second commit: a rename after a reopen of a table sealed at a high version, two renames made just before a seal followed by two after it, and three renames replayed from the log after an open without a checkpoint; all three fail on the second commit. A writer unit test walks a versioned history, reads an unversioned one as before, and leaves a rename made in the seal's own version alone. Two more from the review of the third commit: a rename replayed from a log kept past the checkpoint that wrote the manifest is not recorded twice (rows sealed before, between and after read their values after two reopens), and two renames survive a full checkpoint and a reopen through the manifest alone; both fail on the third commit. The seal race itself (a rename between the schema read and the registration) has no deterministic guard; the fix reads the version under the schemas' lock. A checkpoint run while a rename statement is in progress (the statement's steps under the guard it holds, as the executor runs them) seals before or after the rename and writes the manifest only once the statement is complete; the rows read their values live and after two reopens. This guards the arrangement rather than a regression: the interleaving the review produced is between the steps of one statement, which the guard now excludes, so it cannot be reached through the executor. Formatting, all-target all-feature clippy with warnings denied, the file under both runners and under `test-filedb`, the full in-memory suite and the full `test-filedb` suite pass.

Gate: the 53-shape suite, 8 runs per build in alternating order, against main fe2cf12b on an idle machine: median latency delta -0.1 percent on the first two commits, +0.7 on the third, +1.8 then -0.3 on two runs of the fourth +1.1 then +0.6 on two runs of the fifth, +1.3 on the sixth and +0.4 on the seventh, allocation counts identical on every fixed-count shape (the one alloc flag, DELETE complex at 52.50 against 52.57, is a shape whose count varies run to run). The suite runs an in-memory database and reaches none of these paths; the flags, SELECT by index range +5.1 and HAVING complex +4.8, are its short-shape spread.


semihalev:
Gate on this head (92c1463a), the 53-shape suite, 8 runs per build in alternating order, against main fe2cf12b on an idle machine: median latency delta -0.1 percent, allocation counts identical on every fixed-count shape (DELETE complex at 52.50 against 52.57 is a shape whose count varies run to run). The flagged shapes (SELECT by index range +5.1, HAVING complex +4.8, CTE + JOIN +3.8) run an in-memory database and reach none of the changed paths. Two earlier attempts on this head were voided by machine load and are not reported.

semihalev:
All three fixed in the second commit, by giving the renames an order against the drops rather than by tuning the walk: the cases cannot be told apart from the records alone (your third case and "rename a to b, drop b, add b with a default" are the same records either way round). A rename records its schema version now, kept in the drop list under a name no column can carry (old NUL new), so the manifest format stays at 6 and an older reader never matches it; a column's source is found by walking the changes since the seal newest first, a rename into the name taking the name it had before and a drop of the name ending the identity there. Guards for your three histories and for the drop-and-add-back one, each read after the history and again after a reopen; your three fail on the first commit, the fourth passes there too and pins the walk's drop rule. Legacy renames without a marker are treated as the newest change, which is the order the list was read in before, so old manifests keep their old behaviour. The DDL order change stays. The permanent form is the plan's K7 (column identity); the markers go with it.

semihalev:
Gate on the second commit (6c7dc6e1), same method, idle machine: median latency delta -0.1 percent, allocation counts identical on every fixed-count shape. The flags (CTE + JOIN +10.5, UPDATE complex +6.2, index range +6.1) are shapes whose latency varies run to run on an in-memory database that never reaches the mapping.

semihalev:
All five addressed in the third commit.

1. The schema epoch restarts above every version the manifests carry (segments, drops, markers), right after the manifests load and before the log is replayed. This also mends the drop masks, which compared the same versions across an open.
2. A change recorded in the version a later seal carries happened before that seal: the walk takes only changes strictly after the volume's version, and the `name_now` fallback is gone.
3. The replay path increments the epoch before it records the rename, as the SQL path does; the three-rename swap replayed from the log resolves the same as the live statements.
4. A history whose renames carry no marker is read exactly as before the markers (one record back, a name dropped at or after the seal resolving to nothing), not as "newest change"; your legacy case gives the default again.
5. The marker is a NUL byte and the rename's ordinal in the rename list, so it is short whatever the names and pairs with its record by position; the `taken` matching is gone with it.

Guards: your first three histories as integration tests (each read live and after a reopen; all three fail on the second commit), and a writer unit test for the versioned walk, the unversioned reading and the seal-version boundary. Not changed: the duplicated rename records when a manifest persisted while the log still held the statements, which predates this branch; with ordinal markers each copy carries its own version, and the walk cancels a repeated rename pair but not a repeated swap.

semihalev:
Gate on the third commit (7611a853), same method, idle machine: median latency delta +0.7 percent, allocation counts identical on every fixed-count shape and no allocation flag. The flags (NOT EXISTS subquery +7.8, SELECT by ID +7.0, IN subquery +4.3) are the suite's short-shape spread on an in-memory database that never reaches the mapping or the open path.

semihalev:
Both fixed in the fourth commit, and the older loader finding with them.

1. Replay: the manifest keeps the log position its renames and drops reach (a `\0lsn` marker in the drop list, so still no format change); the statement path notes its record's position after writing it, under the DDL guard; a rename or drop replayed at or below that position changes the schema and recomputes the mappings but records nothing. Guard: your scenario with a snapshot holding the truncation back, rows sealed before, between and after the rename, read after two reopens; the middle row came back NULL on the third commit.
2. Seal: the schema version is read under the same lock the schemas are read under, and the volume is registered with that version rather than the one current at registration; compaction reads its version with the schema the same way. No deterministic guard for the race itself.
3. The loader no longer merges renames into a volume's `column_name_map`: a volume keeps the names it was sealed with and the mapping alone resolves the renamed ones. Guard: `a` to `t`, `b` to `a`, full checkpoint, reopen: (t, a) = (1, 2), which was (2, NULL) on main.

Unchanged and stated in the body: a history with a rename recorded before the markers has no order to walk and is read one record back, as before.

semihalev:
Gate on the fourth commit (63e13a3b), same method, two runs on an idle machine: median latency delta +1.8 percent and -0.3 percent, allocation counts identical on every fixed-count shape (the alloc flags, LEFT JOIN + GROUP BY and CTE + JOIN, are the shapes whose counts vary run to run). The first run's flags (DISTINCT +13.6, SELECT by ID +9.1) did not repeat; the suite runs an in-memory database and reaches none of the seal, replay or load paths this commit changes.

semihalev:
Both fixed in the fifth commit, and the seal's first mapping with them.

1. The manifest is written out under the DDL guard, at the checkpoint's write (step 4, after the fence) and at compaction's two writes, so a persisted manifest never shows a statement halfway; the statement records itself in the log first and the manifest takes the rename together with that record's position in one write. The interleaving you produced sat between the steps of one statement, which the guard now excludes; the test that replaces it runs a checkpoint concurrently with the statement's steps under the guard and checks the rows live and after two reopens. Lock order: the checkpoint takes the guard only after its seal steps, compaction holds no fence, and no DDL path writes a manifest, so there is no cycle with a statement waiting on the fence or on admission.
2. The schema epoch moves under the schema cache's write lock on every path that replaces a schema (`refresh_schema_cache` included), so a reader holding that lock sees the schema and its epoch together.
3. A volume is registered with its mapping computed against the schema current at registration rather than the identity, so a seal that completed under the older schema resolves a renamed filter column from its first read; compaction recomputes its outputs' mappings when the epoch moved during the rewrite. Your `x=9 AND z=1` case reads 30,000 rows before any further change.

semihalev:
Gate on the fifth commit (b50f8387), same method, two runs on an idle machine: median latency delta +1.1 and +0.6 percent, allocation counts identical on every fixed-count shape and no allocation flag on either run. The first run's CTE + JOIN +20.6 did not repeat (that shape's latency varies run to run); the second run's flags (INSERT single +7.5, LIKE prefix +6.8, Window ROW_NUMBER +6.4) are the suite's short-shape spread on an in-memory database, which takes the DDL guard once at setup and never reaches the seal, persist or replay paths this commit changes.

semihalev:
Both closed in the sixth commit by making the mapping and the segment appear together. Publication holds the schema cache's read lock while the mapping is computed and the segment made visible, for the seal's registration and for compaction's replacement (`replace_segments_atomic_multi` takes the schema and computes the outputs' mappings under the manifest lock it already holds, through the same `mapping_for` registration uses); the correction step after replacement is gone. Since every change replaces the schema under that lock's write side and propagates its history after, a publication is either wholly before a change, whose propagation then recomputes the new segment's mapping, or wholly after it, with the current schema and history. Lock order is schema cache, then manifest and segments, which is the order the DDL paths use; registration inside the seal fence holds no lock a DDL statement takes while waiting for the fence. No deterministic guard for either window; the existing file and the unit tests pass.

semihalev:
Fixed in the seventh commit. `record_key_order` keeps the physical columns the writer hands it, and compaction hands it the key's positions in the rewrite schema, which are the outputs' physical columns since they were built in that schema's column order; a lookup still resolves the schema's key through the mapping (`volume_key`) to compare. Unit test with your sequence: (id, d, k, v) sealed with k ordered, d dropped and the mappings recomputed before the record, then the record by position 2: k at its new position 1 reads as the ordered column, v at position 2 reads as unchecked and a decision on it says not ordered; on the previous commit the record claimed v.

semihalev:
Gate on the seventh commit (18cda5be), same method: median latency delta +0.4 percent, allocation counts identical on every fixed-count shape (CTE + JOIN, 1676.30 against 1702.30, is a shape whose count varies run to run). The flags (CTE + JOIN +7.8, INSERT single +7.0, SELECT by ID +6.3) are the suite's short-shape spread on an in-memory database; none of them reaches the compaction record this commit changes. The sixth commit's own run gave +1.3 on a machine that was not idle and is superseded by this one.

PR #202, opened 2026-09-19
Verify a LIMIT read against the seal generation, and collect again when a seal landed in it

semihalev:
The two LIMIT collectors on a segmented table read the hot store and the cold view at different moments, and a seal that lands between the two moves rows from one to the other. Reproduced on the #201 head with a checkpoint run from the `after_cold_volumes_taken` hook, two sealed volumes and 1,000 hot rows:

| seal lands | path | rows / distinct ids |
|---|---|---:|
| after the cold view, before the hot read | `collect_rows_with_limit` | 2,000 / 2,000 |
| after the hot read, before the cold view | `collect_rows_with_limit_unordered` | 4,000 / 3,000 |

The first loses the sealed rows (in neither side), the second shows them twice (in both). The order of the two reads is the same on main before #199; #201 did not change it.

## What changes

Both collectors run under `settled_across_seals`: the seal generation is read under the seal's read fence before the collection and again after it, and the collection is accepted only when the two are equal. A seal registers its volume and removes the hot rows under the write fence and moves the generation there, so the reads under the read fence never land inside a seal, and an equal generation means no seal ran between them. A changed generation drops the result and collects again, and every attempt is verified by the same rule; none is trusted for its number. The fence is never held across a collection, so a read never delays a seal. Under seals landing faster than a collection runs, the read keeps collecting until one is clear; a firm bound on that needs a hot and cold capture taken together, which this PR does not attempt.

The hot read of the unordered path moved inside the verified span, since its window is between that read and the cold view. Nothing else in the two collectors changed; the cold view, the mapping and the file lifetime stay as #201 left them.

A `test-failpoints` hook, `after_hot_rows_taken`, fires in the unordered path between its hot read and its cold view.

## Tests

`tests/limit_seal_consistency_test.rs`, in the CI failpoint step. Each fails on the #201 head with only the hook call added, and passes with the fix:

- `an_ordered_limit_keeps_the_rows_a_seal_moves_after_its_cold_view`: 3,000 rows, every id once (2,000 without the fix).
- `an_unordered_limit_shows_once_the_rows_a_seal_moves_after_its_hot_read`: 3,000 rows, every id once (4,000 with 1,000 repeated without the fix).
- `a_retry_a_seal_lands_in_again_is_verified_too`: the first seal drops attempt one; a batch inserted meanwhile and a second seal land in attempt two, which is dropped the same way; the answer has all 4,000 ids and both seals were counted.
- `every_attempt_a_seal_lands_in_is_dropped_however_many_in_a_row`: five attempts in a row each hit by a seal that moved a fresh batch; the sixth answers all 7,000 ids. With a fifth attempt trusted unverified, it answers 8,000 rows with 1,000 repeated and the test fails.

The hooks run their checkpoints on the reader's own thread, so every interleaving is deterministic and no test sleeps. A hook a read never reached is disarmed at the end of the test, since dropping a database handle inside the thread's destructor aborts the process instead of failing the test.

## Measured

A probe on a file database with four sealed volumes of 65,536 rows and 10,000 hot rows, `fast` profile, three runs per build in contiguous blocks, the #201 head first and then this branch. Phase A loops one collector alone with `k = 3 LIMIT 10`; phase B inserts single rows for six seconds with a checkpoint every second on another thread; phase C adds one reader thread per collector to phase B; phase D seals everything phase C left (about 3.1 million rows in 32 volumes) and loops the ordered collector alone, then adds 200,000 hot rows and loops it again. Microseconds unless marked; each cell is the range over the three runs.

| point | #201 head | this branch |
|---|---|---|
| A ordered alone, p50 / p99 | 964 to 1,014 / 1,023 to 1,096 | 970 to 986 / 1,023 to 1,086 |
| A unordered alone, p50 / p99 | 0.4 to 0.5 / 0.5 | 0.5 / 0.5 to 0.6 |
| B INSERT alone, p50 / p99 / max, 6 seals | 2.2 to 2.3 / 7.8 to 7.9 / 55.9 to 57.6 ms | 2.3 / 7.8 / 53.6 to 58.9 ms |
| C INSERT with two readers, p50 / p99 / max | 5.4 to 5.7 / 16.1 to 17.0 / 62.5 to 67.1 ms | 5.2 to 6.2 / 15.7 to 17.1 / 59.1 to 64.4 ms |
| C ordered under inserts and seals, p50 / p99 / max | 35.5 to 37.0 / 65.7 to 86.9 / 83.7 to 93.9 ms | 35.1 to 36.5 / 117.5 to 119.7 / 126.9 to 130.8 ms |
| C unordered under inserts and seals, p50 / p99 | 0.9 / 1.6 to 1.7 | 0.8 to 0.9 / 1.8 to 1.9 |
| D ordered, all rows cold, p50 / p99 | 21.7 to 24.0 / 32.4 to 36.2 ms | 21.1 to 21.8 / 32.5 to 37.3 ms |
| D ordered, same cold plus 200,000 hot, p50 / p99 | 35.1 to 36.6 / 39.1 to 40.5 ms | 35.4 to 36.4 / 40.0 to 41.2 ms |

INSERT latency: the ranges overlap in both phases, alone and with the readers. A call no seal lands in: the ranges overlap in phases A and D. The ordered collector under phase C: p50 unchanged, p99 and max higher on this branch, since six seals over about 160 calls of 35 ms land inside a few of them and those are collected twice; this is the cost of dropping a collection a seal moved rows through.

Where the 35 ms goes (phase D): the authority map over about 3.1 million cold ids in 32 volumes takes 21 to 24 ms, and the hot id set over 200,000 hot rows adds 14 ms on top of it, on both builds alike. Per id that is about 7 ns on the cold side and 70 ns on the hot side; the hot id collection is the more expensive structure per row, which is a separate item and not touched here.

Suites: `cargo nextest run --no-fail-fast`, `cargo nextest run --features test-filedb --no-fail-fast`, the CI failpoint step with the new target, clippy with all targets and features, `cargo fmt`.

## Not in this PR

- A read under seals landing faster than its collection runs keeps collecting; the cost is the dropped collections, bounded here for a seal per second and not in general. A firm bound needs the hot rows and the cold view captured together.
- `scan_top_k` and the scan constructor keep their own guards; the two LIMIT collectors are the only paths this PR verifies.


semihalev:
Both findings are in the new head.

P1: the fenced last attempt is gone. `settled_across_seals` now loops: generation under the read fence, collection outside it, generation under the read fence again, accept on equality, otherwise collect again. Every attempt is held to the same rule, and the fence is never held across a collection. The body says plainly that under seals landing faster than a collection runs the read keeps collecting, and that a firm bound needs a hot and cold capture taken together.

P2: the fourth test now lands a seal in five attempts in a row and expects the sixth to answer all 7,000 ids. With a fifth attempt trusted unverified it answers 8,000 rows with 1,000 repeated and fails; with the fix it passes. The other three are unchanged.

Measurement redone as three contiguous runs per build, base first, and the wording changed to overlapping ranges. Phase D splits the ordered collector's 35 ms: 21 to 24 ms is the authority map over about 3.1 million cold ids, 14 ms is the hot id set over 200,000 hot rows, about 7 ns per cold id against 70 ns per hot id on both builds. The hot side is the more expensive structure per row; that is a separate item.

Suites on the new head: memory 5,610 passed, filedb 5,603 passed, the CI failpoint step with the new target, clippy and fmt clean.

PR #205, opened 2026-09-20
A paged secondary index for a sealed volume's column, with its ledger and cursor

semihalev:
First of the deliveries the paged secondary access design names (`secondary-paged-access-design.md` in the plan directory): the paged index infrastructure and its measurement. No query path changes here; the engine does not build or read these files yet. That comes with the next delivery, together with publication and file lifetime, bounded fetch, `needed` projection and the visibility guards.

## What is built

`src/storage/volume/secondary.rs`: a paged index over one integer column of a sealed volume.

- **Layout.** A side file holds, per indexed column, the distinct keys in order and for each key its ascending row positions, in pages bounded in bytes (32 KiB, header and checksum included): a key page carries keys with the position index each key ends at, a position page carries positions; one key with a million positions spans 123 position pages. A small directory names every page with its first and last key, offset, length, key index and position index, and stays resident; the file ends with the directory, its length and CRC, and the magic. The header carries an immutable generation.
- **Corruption checks.** Magic, version and footer magic; the directory's CRC and the consistency of its entries (order, bounds against the file length, page lengths against the bound); every page's CRC and its count against its length and order. A corrupt page fails the read of that page alone; a corrupt directory fails the open.
- **Generation.** A page read re-reads the header and compares its generation with the one the directory was read at; a file replaced under an open reader (a backfill) answers `generation changed`, which `is_generation_changed` names, and the caller reopens. Cache keys carry the generation, so the old reader's directory and the new file's pages never meet.
- **Ledger and page cache.** One process-wide ledger counts every byte the index holds, charged before the allocation is made, at the capacity it takes, and released when its last owner lets go: a directory for the life of its `IndexFile`, a page for as long as any holder keeps it, the raw bytes of a page or a directory while they are parsed, a build's sort run, page buffers, output buffer, merge readers and page metadata, a cursor's window. The page cache evicts unheld pages while a reservation would go over the budget; when nothing unheld is left the reservation is still granted and counted as over budget. An eviction removes the map entry only; a held page's bytes stay charged until its last holder lets go. The ledger is atomics, the cache has one lock, and nothing is dropped or reserved while that lock is held.

  This is accounting with best-effort eviction, not a memory limit: held pages, resident directories and concurrent builds have no maximum aggregate overshoot. Admission, the rule that refuses or delays an allocation the budget cannot take, is a named prerequisite of the query integration and is not in this PR.
- **Cursor.** Equality reads one key page and yields the key's position index range; a range reads the boundary key pages for the exact range, and `candidate_bound` gives the directory's upper bound without a page read. The cursor yields positions in windows of at most `window` entries, one page loaded at a time, each window sorted (the window is the whole workspace, reserved for the cursor's life).
- **Build.** `(position, key)` pairs of any order are sorted in runs that fit the build's own budget, spilled and merged upward as they accumulate: a level holds fewer than `fan_in` runs, and the run that fills it is merged with the others into one run of the next level, so the files alive and the names held stay bounded by the fan-in and the levels (eleven to thirteen files for 1,172 or 4,096 runs), not by the input. The fan-in comes from the workspace left after the writer's buffers, the output and spill buffers and the merge heap, capped at 32 open runs. Every byte the build holds is taken from its budget before it is allocated: the directory's column slots and their metadata charges, the sort run, the page buffers, the output, spill and merge buffers, the readers, the heap, the run names of a merge, the page metadata as it grows, and the encoded directory at the end; earlier columns' metadata stays counted while later columns build. A take the workspace cannot hold fails the build with `InvalidInput`, as does a workspace under the minimum (about 290 KiB, `MIN_WORKSPACE_BYTES`, covering one column's slot; every further column adds `COLUMN_SLOT_BYTES`, and the page metadata an input needs comes on top, `metadata_allowance`); an input without an upper size bound keeps a quarter of what is left for its metadata and fails if that is outgrown, and a bound avoids it. Every build works in a directory of its own beside the target, `<name>.build-<pid>-<generation>`, created by the build and refused if it already stands: the temporary output and the runs live there, a failure removes the directory with everything in it and nothing outside it, success renames the output out and removes the empty directory. A relative target without a parent builds beside itself in the current directory. The report names keys, positions, pages, file bytes, runs spilled, merges, the largest fan-in, the most files alive, the build's own peak and the time.

## Tests

Fifteen in the module, each against expected sets; the lib test binary carries a per-thread counting allocator (unless the `mimalloc` feature supplies one), and the workspace tests check the allocator's peak beside the ledger's:

- keys map to their positions through pages: equality, absent keys, a closed range, an empty range, and the bound never below the exact count;
- a key with a million positions spans pages and a walk over it with an eight-page budget never goes over budget, by the ledger's high-water mark;
- a page is charged at the capacity of its vectors, its raw bytes only while it is parsed (the high-water mark shows the raw page), and the directory's charge equals its vectors' capacity;
- a build beyond its workspace spills runs, merges them upward with a fan-in under the ceiling and at most a dozen files alive, and writes a byte-identical file (past the generation) to the in-memory build; a workspace under the minimum is refused;
- the three shapes of the review, held to the workspace by the build's own ledger and checked against the allocator: 1.2 million unique keys under the minimum plus their metadata (1,172 runs), three columns of 400,000 rows under 512 KiB, and 4.2 million rows without a size bound under the minimum (refused up front, nothing allocated past the workspace, nothing published) and with the bound given (built);
- a fixed workspace of 512 KiB holds while the input grows from 50,000 to 800,000 rows;
- a failed spill after a successful one (a test hook fails the second spill), and a failed final rename, leave no temporary file and keep the published file untouched;
- a failed build touches nothing but its own directory: a published file whose name looks like a build's leftovers still answers, a stray file with the target's name as a prefix keeps its bytes, and a directory already standing where the build's would be makes the build refuse with `AlreadyExists` and leaves it;
- a relative target without a parent builds beside itself in the current directory, and a refusal after its directory was created leaves nothing there;
- the directory's column slots are taken from the budget before they are allocated: 2,000 empty columns under the minimum are refused before anything beyond the inputs is allocated, and with room for their slots they build within the budget, checked against the allocator;
- a corrupt page fails alone and a corrupt directory fails the open, as does a truncated file;
- a page read after the file was replaced reports the generation, and the reopen answers from the new file;
- a held page stays charged after its eviction, the budget is honoured while something unheld can be evicted, a load with nothing to evict is counted as over budget, and everything is released at the end;
- statistics, reservations, loads and evictions from four threads for half a second all stop when told to, with a ten-second watchdog;
- a range bound from the directory is never below the exact count.

## Measured

The probe in the plan directory (`htap-harness/secondary-paged-205`, `fast` profile, three runs, ranges over them) exercises the module alone with a counting global allocator beside the ledger, the allocator's mark and the ledger's high-water reset taken at the same point: builds under a large workspace and under one that forces spills, first access with an empty cache under a 16 MiB budget and under a 64 KiB one, a working set of 1,000 scattered keys under both, a reader holding a page with the budget set below it, one key with a million positions walked under an eight-page budget, the three shapes of the review, and the V4 volume read a positive answer on a metadata-only volume still pays.

### Build

| what | ms | keys | key pages | pos pages | file bytes | runs spilled | merge passes | max fan-in | ledger peak bytes | allocator peak bytes |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 500k, workspace 64 MiB | 18 to 22 | 499993 | 184 | 62 | 8011289 | 0 | 0 | 0 | 8188664 | 8195164 |
| 500k, workspace 1 MiB | 29 to 42 | 499993 | 184 | 62 | 8011289 | 20 | 0 | 20 | 649816 | 650826 |
| 4.2M, workspace 64 MiB | 150 to 167 | 4194297 | 1537 | 513 | 67203137 | 2 | 0 | 2 | 66450424 | 66451178 |
| 4.2M, workspace 8 MiB | 170 to 174 | 4194297 | 1537 | 513 | 67203137 | 9 | 0 | 9 | 7730168 | 7730978 |

### First access with an empty cache (directory open, one equality, its first window)

| file | budget | open ms | directory bytes | lookup ms | page loads | ledger peak bytes | allocator peak bytes | warmed p50 ms |
|---|---:|---:|---:|---:|---:|---:|---:|---:|
| 4.2M | 65536 | 0.023 to 0.023 | 86160 | 0.028 to 0.029 | 2 | 168072 | 168165 | 0.0278 to 0.0284 |
| 4.2M | 16777216 | 0.042 to 0.052 | 86160 | 0.030 to 0.032 | 2 | 211752 | 211925 | 0.0001 |
| 500k | 65536 | 0.013 to 0.013 | 10392 | 0.034 to 0.045 | 2 | 92304 | 92399 | 0.0296 to 0.0304 |
| 500k | 16777216 | 0.042 to 0.048 | 10392 | 0.050 to 0.106 | 2 | 135984 | 136459 | 0.0001 |

### Working set of 1,000 scattered keys, 2,000 timed lookups after a warming round

| file | budget | page loads | hits | evictions | over budget | p50 us | p99 us | max us | ledger peak bytes | allocator peak bytes |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 4.2M | 65536 | 6000 | 0 | 5999 | 15000 | 28.6 to 28.8 | 33.9 to 39.7 | 60 to 75 | 168072 | 184549 |
| 4.2M | 16777216 | 5538 | 462 | 5103 | 0 | 29.2 to 30.3 | 34.5 to 35.8 | 56 to 63 | 16771928 | 16852389 |
| 500k | 65536 | 6000 | 0 | 5999 | 6000 | 28.3 to 29.5 | 32.7 to 36.3 | 54 to 106 | 92304 | 108783 |
| 500k | 16777216 | 244 | 5756 | 0 | 0 | 0.3 to 0.3 | 0.7 to 0.8 | 1 to 8 | 10034960 | 10083467 |

### A held page under a budget below it (200 lookups)

| held bytes | charged with hold | budget set below it | charged after squeeze | loads | evictions | over budget | charged after | charged once released |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 32760 | 162600 | 129840 | 118920 | 399 | 398 | 998 | 151680 | 0 |

### One key with a million positions, budget of eight pages

| build ms | file bytes | pos pages | walk ms | ledger peak | allocator peak | budget above the base | loads | evictions | over budget |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 11 to 12 | 4005177 | 123 | 1.9 to 1.9 | 315920 | 316728 | 327680 | 124 | 115 | 0 |

### The review's shapes over the workspace

| shape | workspace | outcome | ledger peak | allocator peak | runs spilled | passes | max fan-in | max live runs | ms |
|---|---:|---|---:|---:|---:|---:|---:|---:|---:|
| 1.2M rows, one column, minimum plus its metadata | 320952 | built | 290108 | 291792 | 1172 | 1170 | 2 | 11 | 349 to 388 |
| 400k rows, three columns, 512 KiB | 524288 | built | 510036 | 510958 | 135 | 19 | 8 | 13 | 68 to 72 |
| 4.2M rows, no size bound, minimum | 290232 | refused: side index build: 115392 bytes of workspace left for column 1 cannot hold 2 merg | - | 66298 | - | - | - | - | - |
| 2,000 empty columns, minimum | 290232 | refused: side index build: 192000 bytes for the directory's columns would take the build  | - | 232 | - | - | - | - | - |
| 2,000 empty columns, minimum plus their slots | 1154232 | built | 424748 | 383446 | 0 | 0 | 0 | 0 | 4 to 5 |
| 4.2M rows, size bound given, minimum plus its metadata | 413112 | built | 375641 | 352824 | 4096 | 4095 | 2 | 13 | 1310 to 1424 |

### V4 first access a positive answer on a metadata-only volume still pays

| rows | volume file bytes | read_volume_from_disk ms (three reads per run) |
|---:|---:|---:|
| 500000 | 10074268 | 3.9, 3.1, 3.0 |
| 500000 | 10074268 | 4.1, 3.3, 3.0 |
| 500000 | 10074268 | 4.2, 7.8, 5.2 |

What the numbers say:

- **The ledger's high-water mark matches the allocator.** A lookup's first access: 168,072 charged against 168,165 allocated at the same baseline; the million-position walk 315,920 against 316,728 under a budget of 327,680; a 500,000-row build 8,188,568 against 8,195,047. The working-set rows differ by the strings of 6,000 file opens (up to 10% under the 64 KiB budget, 0.5% under 16 MiB), which the ledger does not count and the body says so.
- **The build's own budget bounds the whole build.** 1.2 million unique keys under the minimum plus their metadata (320,952 bytes): 1,172 runs, at most 11 files alive, ledger peak 290,108, allocator 291,792. Three columns of 400,000 rows under 512 KiB: ledger 510,036 with earlier columns' metadata counted, allocator 510,958. 4.2 million rows without a size bound under the minimum: refused before any pair is read, 66 KB allocated; with the bound, under the minimum plus its metadata (413,112): 4,096 runs, 13 files alive, ledger 375,641, allocator 352,824. 2,000 empty columns under the minimum: refused before anything but the inputs is allocated (232 bytes); with room for their slots they build with a ledger peak of 424,748 against 383,446 allocated.
- The file takes 16 bytes per row at unique keys (12 per key, 4 per position), 8 MB per 500,000 rows; the directory resident with it is 10 KB, and 86 KB for 4.2 million rows.
- First access with an empty cache: directory open 0.013 to 0.050 ms, one equality with its first window 0.03 to 0.11 ms and two page loads; warmed, 0.0001 ms under budget and 0.03 ms when every lookup reloads its two pages.
- A working set that fits (500,000 keys under 16 MiB): 244 loads for 3,000 lookups, p50 0.3 microseconds. One that does not: every lookup loads, p50 28 to 30 microseconds, p99 33 to 37. Under a 64 KiB budget every reservation is over budget once the directory alone exceeds it, and the ledger counts each.
- A page a reader holds stays charged when the budget is set below it; the 200 lookups around it evict everything else, count their reservations as over budget, and the ledger returns to zero once the holder and the file let go.
- The V4 side: a positive answer on a metadata-only 500,000-row volume still reads its 10 MB file through the CRC, 2.8 to 4.2 ms, a hundred times the directory open; an empty answer avoids it. The next delivery probes the index before the reload.

## Review rounds

First round, four findings, all in:

- **Lock order.** Statistics took the ledger then the cache, reservations the cache then the ledger, and two concurrent calls stalled. The ledger is atomics now and the cache has one lock that is never held across a drop or a reservation. The concurrency test fails on a ten-second watchdog if any worker waits on another.
- **Accounting.** The ledger charged a page at its encoded length while the parsed vectors took more, and reserved directories and cursor windows after allocating them. Every allocation is reserved before it is made at the capacity it takes.
- **Bounded merge.** The merge opened every run at once with a 64 KiB reader each. The fan-in is bounded by the workspace and a ceiling of 32.
- **Temporary files.** A failed build left its runs and its `.tmp`. Every failure path removes what the build created.

Second round, three findings, all in:

- **Run bookkeeping.** The names of every spilled run were held in memory outside the ledger, 1,042 of them for 1.2 million rows under the minimum workspace, more than the workspace itself. Runs are now merged upward as they accumulate, a level at a time, so at most a dozen files are alive and named at once whatever the input, and the names a merge holds are taken from the budget.
- **The whole build.** The workspace was applied per column, earlier columns' metadata was not counted against later ones, an input without a size bound was taken for zero rows, and the final directory was not checked. The build now has its own ledger over the workspace: every take is checked against it, earlier columns stay counted, an unbounded input keeps a quarter of what is left for its metadata and fails if that is outgrown, and the encoded directory is taken like everything else.
- **Peak measurement.** Peaks were sampled between calls and the allocator's baseline was set after the directory was opened. The ledger keeps a high-water mark taken at every charge, the build's report is its own budget's peak, and the probe sets the allocator's mark and the ledger's reset at the same point. The earlier "within 0.1%" and "within 1%" claims are withdrawn: the numbers above are the corrected ones, and the working-set rows differ by the strings of file opens.

Third round, three findings, all in:

- **Cleanup by name pattern.** A failed build removed every file whose name began with the target's name and looked like a run, which a published `alpha.sidx.run-backup.sidx` did. Every build now works in a directory of its own, created by the build and refused if it already stands; a failure removes that directory and nothing else.
- **Two vectors sized by the column count outside the budget.** The directory's columns and their metadata charges were allocated for every input column without a take; 2,000 empty columns built under the minimum with an allocator peak above it. Both are taken from the budget before allocation, one slot per column, and the minimum names one column's slot.
- **A relative target without a parent.** `parent()` gave an empty path, the cleanup listed nothing and left the temporary output. An empty parent is the current directory now, and the build directory lives there; tested with a bare name.

Also from the reviews: the ledger is accounting with best-effort eviction, not a memory limit, and the body says so; admission is a named prerequisite of the query integration. The build's own budget is a bound, since it refuses.

## Not in this PR

- The engine does not build or read the side files yet; no `PRAGMA` sets the budget (the API does); no `MEMORY_STATS` row. All with the next delivery.
- Admission control: a reservation the budget cannot take is granted and counted, never refused or delayed; the rule that refuses or delays, with a defined outcome for an insufficient budget and a progress rule that never waits under a lock, comes before the query integration.
- Keys other than INTEGER and TIMESTAMP nanos, multi-column indexes.
- The page size is 32 KiB; the design names 1,024 and 16,384 keys per page as the alternatives to try when the query path exists.
- Backfill and `DROP INDEX`; compaction's rebuild.

Suites: `cargo nextest run --no-fail-fast` 5,625 passed, `cargo nextest run --features test-filedb --no-fail-fast` 5,618 passed, clippy with all targets and features, `cargo fmt`.


semihalev:
All four are in the new head, in the order given.

P1 lock order: the ledger is atomics and the cache has one lock; nothing is dropped or reserved while it is held (an eviction takes the page out under the lock and drops it after). A new test drives statistics, reservations, loads and evictions from four threads for half a second and fails on a ten-second watchdog if any worker stays blocked; the previous head's opposite orders are gone by construction.

P1 accounting: every allocation is reserved before it is made, at the capacity it takes. A page charges its parsed vectors for its life and its raw bytes only while they are parsed; the directory charges its raw bytes while read and its vectors for the file's life; a cursor's window, the build's sort run, page buffers, output and spill buffers, merge readers, heap and page metadata are all charged. The probe now runs with a counting global allocator: a build's ledger peak is within 0.1% of the allocator's peak (8,188,568 against 8,195,078 bytes at 500,000 rows) and never above the workspace; a lookup's allocator peak is the ledger's end state plus the one raw page held while parsing.

P1 merge: the fan-in comes from the workspace left after the writer's buffers, the output buffer, the spill buffer and the heap, capped at 32, in passes; a workspace under `MIN_WORKSPACE_BYTES` (about 300 KiB), or one that cannot hold the input's page metadata beside a minimal run, is refused with `InvalidInput`. 500,000 rows under 1 MiB spill 11 runs and merge in one pass with a fan-in of 11, peak 1,033,976 bytes; the test holds a fixed 512 KiB while the input grows from 50,000 to 800,000 rows.

P2 temporary files: a guard removes what the build created on every failure path and keeps the file published before; tested with a failed second spill and a failed final rename.

The contract is stated as you put it: the ledger is accounting with best-effort eviction, not a memory limit; admission is named in the body and the design as a prerequisite of the query integration.

Suites on the new head, eleven module tests, clippy and fmt: in the body.

semihalev:
All three are in the new head.

P1 run bookkeeping: runs are merged upward as they accumulate, a level at a time, so the files alive and the names held stay bounded by the fan-in and the levels: 1.2 million unique keys under the minimum plus their metadata spill 1,172 runs with at most 11 alive, and the names a merge holds are taken from the budget. The `TempFiles` guard no longer remembers paths; every file the build creates carries the target's name as a prefix and a failure lists the directory.

P1 the whole build: the build has its own ledger over the workspace. Every take is checked against it and refused with `InvalidInput` when it does not fit: the sort run, page and output buffers, spill and merge buffers, readers, heap, run names, the page metadata as it grows, and the encoded directory at the end; earlier columns' metadata stays counted while later columns build. An input without an upper size bound keeps a quarter of what is left for its metadata and fails if that is outgrown; 4.2 million rows under the minimum are refused before any pair is read, and with the bound given they build under the minimum plus their metadata (4,096 runs, 13 files alive, peak 375,545 in 413,016). Three columns of 400,000 rows under 512 KiB peak at 509,490.

P2 peaks: the ledger keeps a high-water mark taken at every charge (`peak_bytes`, `reset_peak`), the build reports its own budget's peak, and the probe sets the allocator's mark and the ledger's reset at the same point: the million-position walk reads 315,920 against the allocator's 316,728, your audit's number. The "within 0.1%" and "within 1%" claims are withdrawn in the body; the working-set rows differ by the strings of the file opens, which the ledger does not count.

Tests: twelve in the module, the three shapes checked against a per-thread counting allocator in the lib test binary. Suites, clippy and fmt: in the body.

semihalev:
All three are in the new head.

P1 cleanup: every build works in a directory of its own beside the target, `<name>.build-<pid>-<generation>`, created by the build and refused with `AlreadyExists` if it already stands; the temporary output and the runs live there, a failure removes that directory with everything in it and nothing outside it, success renames the output out and removes the empty directory. The test publishes `alpha.sidx.run-backup.sidx` and a stray `alpha.sidx.tmp`, fails a build of `alpha.sidx` after its first spill, and checks the lookalike still answers, the stray keeps its bytes, and a pre-existing directory at the build's name is left with its contents.

P1 column vectors: the directory's columns and their metadata charges are taken from the budget before either vector is allocated, one `COLUMN_SLOT_BYTES` per input column; the minimum covers one column's slot and every further column adds its own. 2,000 empty columns under the minimum are refused before anything beyond the inputs is allocated (allocator-checked), and with room for their slots they build within the budget.

P2 relative target: an empty parent is the current directory; the build directory is created there and removed on refusal. Tested with a bare name: nothing stays after the refusal, and the bare name builds and opens.

The failed-spill test now fails the second spill through a test hook instead of a directory in the way, since the runs live in the build's own directory. Fifteen tests in the module; suites, clippy and fmt in the body. The `test_classification_caching` failure in Coverage is the executor's classification cache test racing another test on a process-wide cache; it failed the same way on an unrelated PR earlier and passed on rerun, and this PR does not touch that code.

PR #206, opened 2026-09-20
Build a volume's secondary index side file with the volume, under a budget, and own it through its life

semihalev:
The engine now owns the paged secondary index side files from #205: a volume of a table with a single-column B-tree index on an INTEGER or TIMESTAMP column gets `vol_<id>.sidx` beside it when it is sealed or compacted, and the file follows the volume through renames, retirement, DROP TABLE, TRUNCATE and reopen. Queries do not use the side files yet; that is the next PR (design 8.4). This one is the lifecycle, the budgets and the admission rule.

## What changed

**Budgets (design 8.1, 8.2).** Three ledgers instead of one:

- `INDEX_CACHE_MB` (default 16): the pages readers hold or cache, raw parse buffers and cursor windows. `try_reserve` evicts every unheld page first and then refuses what still does not fit; page loads and cursors return a `WouldBlock` error the caller answers with its fallback (`is_refused`). `reserve` and `over_budget` keep their meaning for accounting. Lowering the budget never revokes a holder.
- `INDEX_BUILD_MB` (default 64): the whole workspaces of the builds in flight and the decode of their input. `build_side_file` admits its workspace against it before it creates anything; the engine's `build_side_for` admits beside it what decoding one row group of the input allocates at most (the compressed block, its decompressed bytes and the decoded column, from the store's block lengths), holds one decoded group at a time, and takes its workspace from the same share. A refused admission leaves no file. The per-build `Budget` is unchanged inside.
- Directories: reported, never refused. The resident directory and the raw directory buffer are charged there rather than to the page ledger.

`PRAGMA INDEX_STATS` reports the three ledgers with budget, charged, peak and refused, plus the cache counters, the builds that failed with an I/O error and the side files discarded before publication. `PRAGMA MEMORY_STATS` keeps its per-table shape.

**Ownership (design 8.3).** `ColdSegment` carries `side: Option<Arc<IndexFile>>`. `IndexFile` opens through the file's `VolumeFile` handle (`open_through`) and reads every page through it, so a rename cannot land between resolving a path and opening it, and a captured reader keeps a retired side file until it lets go. The side file is built and opened after the volume is written and before the fence or the publication locks; the seal passes it through `register_volume_with_id_and_seal_seq`, the compaction through `replace_segments_atomic_multi` beside the file owners. Cleanup retires side files by path with their volumes (`retire_side_of`), `delete_volume` and `delete_all_volumes` retire them on DROP TABLE and TRUNCATE, and reopen attaches the side file when it reads and leaves it in place, logged, when it does not.

**Definitions.** Each side file column records the identity of the index definition it was built under (column name, index type, uniqueness; `definition_of`), format VERSION 3. The seal and the compaction capture the definitions with the schema and check them again under the fence, before `replace_segments_atomic_multi` for the compaction: an index created or dropped meanwhile discards the prepared side files and the volumes are published uncovered, counted in `sides_discarded`.

**Outcomes.** A refused admission (counted by the ledger) or a side file that cannot be written or read back (`builds_failed`, logged, temporary files removed) publishes the volume without a side file; the next compaction covers it again. An error reading the volume itself is not a side file outcome: it fails the seal the way a failed unique prebuild does and fails the compaction's preparation, with the written files removed. Positions come from the written volume's column, decoded one row group at a time through the group cache when the column is not resident. Side files found stale under the seal fence are collected there and discarded after it, so no registry lock or file removal runs under the fence. On reopen a side file that does not open is left in place and logged, and the volume is uncovered.

Docs: the persistence page describes the side files, their budgets and outcomes; the PRAGMA page lists `index_cache_mb`, `index_build_mb` and `index_stats`.

## Tests

`tests/secondary_side_file_test.rs` (the failpoint variants are in the CI failpoint step):

- a seal writes one side file beside each volume, indexing the integer column alone under its definition, positions checked against the rows; a table without such an index gets none; workspaces released; directories charged
- a compaction replaces the inputs' side files with the outputs'; `INDEX_BUILD_MB = 0` publishes the next volume uncovered with the refusal counted and every row still readable; the next compaction covers it again
- DROP TABLE, TRUNCATE and ALTER TABLE RENAME take the side files with their volumes
- reopen attaches the side files (directories charged) and leaves one it cannot read in place, only the readable one's directory resident, the volume staying
- failpoints: a captured reader keeps the retired side files of a compaction until it lets go; DROP INDEX between the build and the publication discards the prepared side file on the seal path and on the compaction path

Unit tests in `secondary.rs`: a refused page load under a held page is counted and answered again once the budget allows; a build is refused while another holds the builds budget and creates nothing; the definition round trip and `covers`; the directory charged to its own ledger; on a file-backed two-group volume with the group cache off, a budget under the input's decode refuses the build before anything is decoded, and with a budget of exactly the decode and the least workspace the counting allocator's peak stays within what the ledger admitted; a volume whose file was truncated fails its side build as the caller's error, leaves no side file and is not counted as a side failure; a side file that does not open at reopen is left in place and opens again once readable.

The guards were checked in place: without the definitions check under the fence, without retiring the side files through the registry, without the side retirement in `delete_all_volumes`, and without the input's admission and the group drop, the matching tests fail.

## Suites

Run locally at the merged head: `cargo nextest run --no-fail-fast` (5634 passed, 3 skipped), `cargo nextest run --features test-filedb --no-fail-fast` (5627 passed, 3 skipped), `cargo test --test secondary_side_file_test --features test-failpoints -- --test-threads=1` (6 passed), `cargo clippy --all-targets --all-features -- -D warnings` clean. The seal-fence and cache-budget follow-ups (098604b0) reran the touched unit and integration targets and clippy.

## Measurement

Maintenance round on the harness probe, main ce3a194b against this branch with builds refused (`INDEX_BUILD_MB = 0`) and with builds, three runs per column, ranges in ms. Two shapes: phases (500,000 rows in four seals of 125,000, a fifth seal under a single-row writer, a compaction of the five volumes, a reopen), and a fixed 3-second window in which a writer runs single statements while this thread seals 25,000-row batches back to back, with statements per second over the window and latency percentiles over every statement in it. The system load was higher in this run than in the first (main's seal moved from 86.6 to 89.5 up to 87.6 to 101.5); the ranges are given as measured.

| point | main | builds refused | builds |
|---|---:|---:|---:|
| seal of 125,000 rows | 87.6 to 101.5 | 98.6 to 126.1 | 102.6 to 109.1 |
| compaction, five volumes plus the writer's rows into one | 99.5 to 119.5 | 112.6 to 122.7 | 143.8 to 148.8 |
| reopen | 188.3 to 226.3 | 214.5 to 228.7 | 219.3 to 230.2 |
| RSS after reopen and a count (MB) | 981.8 to 985.6 | 982.8 to 984.2 | 985.1 to 987.5 |
| side file after compaction | none | none | one, 9.7 MiB |
| window, INSERT writer: statements per second | 170,707 to 189,291 | 167,948 to 177,479 | 178,654 to 185,535 |
| window, INSERT writer: p50 / p99 | 0.003 / 0.01 | 0.003 to 0.004 / 0.01 | 0.003 / 0.01 |
| window, INSERT writer: max | 33.3 to 53.1 | 35.8 to 66.5 | 33.7 to 36.3 |
| window, INSERT writer: seals in the window, seal p50 | 26 to 27, 41.2 to 41.4 | 23 to 25, 40.3 to 43.5 | 25, 48.1 to 48.6 |
| window, cold-key UPSERT writer: statements per second | 138,115 to 147,624 | 129,076 to 133,286 | 121,283 to 140,642 |
| window, cold-key UPSERT writer: p50 / p99 | 0.005 / 0.02 | 0.006 / 0.02 to 0.03 | 0.005 to 0.006 / 0.02 to 0.03 |
| window, cold-key UPSERT writer: checkpoints in the window, checkpoint p50 | 13 to 14, 159.9 to 169.4 | 13, 170.2 to 176.7 | 11, 203.3 to 244.6 |

One integer column adds 7 ms to a 25,000-row seal and 8 to 15 ms to a 125,000-row seal in this run, and 24 to 49 ms to the compaction of 625,000 rows. In the fixed window the INSERT writer's p50 and p99 are the same in every column and its rate ranges overlap; what changes is the number of seals the window holds. The claim is that narrow: the build outside the fence does not add to a statement's wait, while each checkpoint takes longer. Under the cold-key UPSERT writer every checkpoint carries a compaction, and the branch rebuilds the growing output's side file each time: checkpoint p50 goes from 160 to 169 up to 203 to 245, the writer's rate band sits lower with the top ends overlapping, and its percentiles stay in the same band. That per-compaction rebuild is this delivery's known cost; merging the inputs' side files instead of rebuilding is separate work. The builds-refused column's seal range sits above main's in this run although that path returns before any build; the measurement does not separate that from system noise. Reopen is 219 to 230 against 188 to 226 with 12.7 KiB of resident directories.

## Not in this PR

The query path (design 8.4), DROP INDEX physical cleanup of side files, backfill of uncovered volumes outside compaction, and the hot B-tree removal under the fence.


semihalev:
Four points addressed in 78c7465a (see the commit for the diff):

1. Input decode inside the admission. `build_side_for` now charges INDEX_BUILD_MB with what decoding one row group of the input allocates at most (compressed block at the LZ4 bound, decompressed bytes, decoded column; from the store's block lengths and group size), takes the workspace out of the same share, and `GroupPairs` drops the group it is done with before decoding the next and at exhaustion. Guard: a file-backed two-group volume with the group cache off; a budget under the decode refuses the build before anything is decoded or written; with a budget of exactly the decode plus the least workspace the counting allocator's peak stays within the admitted bytes plus the directory and slack. Verified in place: with the input charge and the group drop removed the test fails. Your `budget_probe` at 1 MiB now refuses the build instead of overrunning it.

2. Discard after the fence. The seal collects stale sides under the fence and calls `discard_side` once the guard is dropped; nothing takes the registry or removes a file under it. The compaction's check was already before its publication locks.

3. Source read errors propagate. `build_side_for` returns `Result<Option<..>>`: an error left by the input iterator comes back as `Err`, the short side file is retired, `builds_failed` is not touched; the seal removes the written volumes and their sides and returns the error (as the unique prebuild path does), the compaction sets `prepare_error` and breaks out. Guard: a truncated volume file fails the build as the caller's error and leaves no side file.

4. Reopen keeps the file. `open_side_for` logs the error and returns uncovered; the file stays. Guard: a truncated side file is left in place and opens again once its bytes are restored; the integration test now asserts both files survive and only the readable one's directory is resident.

Decisions taken as you set them: INDEX_STATS stays separate; unreadable side files are not removed at reopen. Docs updated for both.

Measurement: the round was rerun with a fixed 3-second window and percentiles (INSERT writer, and a writer alternating inserts with upserts of keys sealed long ago), see the PR body. Narrowed claim: in the window the INSERT writer's p50/p99 are the same in every column and its rate ranges overlap, while each checkpoint takes 7 ms longer per 25,000-row seal. Under the cold-key UPSERT writer every checkpoint carries a compaction and the branch rebuilds the growing output's side file each time: checkpoint p50 160 to 169 against 203 to 245 ms, the writer's rate band lower with the top ends overlapping. That per-compaction rebuild is this delivery's known cost.

Two items I take to 2b rather than this PR: the definition fingerprint is not the monotonic definition generation of design 8.3 and its check is not atomic with publication (index DDL runs under the DDL guard, not the seal fence); the eligibility protocol gets settled before the query path reads a side file. The concurrent-build test still exercises the admission arithmetic, not two decoders running at once.

semihalev:
Both P2s in 098604b0:

1. `stale_sides` takes its capacity (`sealed_sides.len()`) before the fence; the `extend` under it allocates nothing.
2. The decoded group cache's budget is now set through one guard in `group_cache::test_budget::hold(bytes)`, shared by the row reader test and the admission test: it holds the one lock and puts the default back when it drops, on a panic too. The writer test's own mutex and `Restore` are gone. Both tests were run together in one `cargo test` process.

Agreed on the rest: the UPSERT cost stays open as this delivery's known bill (every checkpoint under a cold-key upsert load is a compaction, and the output's side file is rebuilt each time), 2a is not a live performance upgrade on its own, and 2b carries the definition identity and the eligibility check atomic with publication.

PR #207, opened 2026-09-20
Give each index definition the identity of its catalog record, decide side file coverage by it under the DDL guard, and let an admitted reader finish on its own buffers

semihalev:
Delivery 2b-1 of the paged secondary index (design 8.4): the identity of an index definition, its persistence through the catalog, the publication protocol under the DDL guard, the per-use eligibility check on a captured view, and the reader's working reservation with the refusal contract. Queries still do not read side files; 2b-2 brings the scanner caller, the fetch and the scale measurement. The 2a fingerprint (`definition_of`) is gone.

## Identity

An index definition's identity is the LSN the WAL assigned to its `CreateIndex` record. `CREATE INDEX` records as before and takes the identity from the record's LSN (`record_ddl_lsn`); the live index keeps it in the version store beside the index, keyed exactly as the index map keys the name, so two indexes whose names differ only in case are two definitions with two identities. The record's first write carries no identity, which means "this record's LSN"; every later copy of the catalog carries it as a trailing `IndexMetadata` field: the checkpoint's re-recording after WAL truncation, the snapshot's `ddl-*.bin`, the restore's fallback metadata. Replay keeps an explicit identity and takes `entry.lsn` when the record has none, so several legacy indexes stay distinct, each its own record's LSN. An index that reaches a checkpoint's catalog copy without an identity (a legacy restore, which creates indexes in bulk and records the catalog after) takes its copy's LSN, one record per index. No counter is added to the manifest or the engine; nothing is lost when the last index is dropped, since LSNs only grow. A memory engine issues from a process counter and has no side files. `MVCCEngine::index_identity(table, index)` reads it.

The side file records the identity per column (format VERSION 4; the VERSION 3 fingerprint files do not open and count as uncovered until a compaction rewrites the volume). Coverage is per column: adding, dropping or recreating index B changes nothing about column A's coverage. `secondary_index_identities` lists only indexes the catalog has issued an identity to.

## Publication

The seal reads the identities under the DDL guard, builds the side files outside it, then takes the guard, compares each side file column by column with the catalog (`still_covers`), sets aside a file none of whose columns still stands for a current index, takes the fence, registers, and holds the guard to the end of the seal's hot cleanup, so a `CREATE INDEX` cannot copy the rows being sealed into its new index after the cleanup took its list of indexes. Index and column DDL and a checkpoint's manifest write wait for a seal's whole fence section meanwhile; that cost is stated in the persistence page. Set-aside files are discarded after the fence and the guard. The compaction prepares its publication outside the guard (`SegmentManager::prepare_replacement`: owners taken, visibility decided), then under the guard compares, takes stale side files out (`PreparedReplacement::uncover`) and commits (`commit_replacement`), and releases the guard before the input files are retired. A preparation the segments moved under since (a seal registered, a column change rebuilt the mappings, a reload or an eviction swapped the map) is handed back by the commit untouched rather than rebuilt under the guard: the freshness decision is made under the publication's own write locks, before anything changes, so nothing between a check and the publication can slip in; the compaction drops the guard, prepares again outside it (`into_parts`), and compares the identities again before the next commit. `commit_publication` keeps its rebuild only for the plain `replace_segments_atomic` path. Lock order is the DDL guard before the fence, as the DDL paths keep it.

## Eligibility and the reader

`ColdSegment::side_for(column, identity)` resolves the schema column through the segment's captured mapping alone to the physical column and returns the attached side file only if it covers that column under that identity; anything else sends the volume through the scan. No global mutex is added for readers: the identity and the column binding are taken together with the cold view, the existing schema generation checks cover both captures, and the seal check of the hot and cold union stays as it is. An old file with a matching held identity may be read to the end; a new identity with an old file goes to the scan, which is what closes reopen after a drop and recreate.

`IndexFile::reader(column, window)` takes one working reservation from `INDEX_CACHE_MB` before the first row: the window, a raw page buffer and a parsed page of each kind. Refused, the caller serves the scan (`is_refused`). Admitted, the reader always progresses: a page comes from the cache when the cache admits it and through the reader's own buffers (`read_page_into`, uncached) when it does not, so a cache that admits nothing and a budget lowered after admission never stop it. A real read or checksum error is the walk's error, and progress is recorded only for a window returned whole: a window whose second page fails reports nothing and leaves the walk where the window began. The search and the walk are written once, over a page source (`Pages`) that the cache-only `Cursor` and the `Reader` both implement; the old cursor keeps its cache-only semantics for the tests and the harness, and 2b-2 decides whether it stays.

## Tests

Storage level, each checked in place:

- an index dropped and created again with the same definition gets a larger identity; the checkpoint's catalog copy and the reopen keep it; the old side file, attached at reopen, covers the column for the old identity and not for the new; the next seal covers the new index
- a memory engine issues distinct identities
- two first-write records replayed without a checkpoint resolve their own LSNs, distinct, and the next checkpoint copy keeps them (fails with replay resolving a constant)
- a legacy restore: the snapshot's `ddl-*.bin` is rewritten without identities, `PRAGMA RESTORE` creates the indexes without any, each takes its own catalog copy's LSN, distinct, and the reopen keeps them (fails without the assignment in the re-recording)
- failpoints, after the comparison under the guard: a DROP and CREATE from another thread waits for the guard, the side file attaches for the index it matched and does not cover the recreated one, on the seal path and the compaction path; a CREATE INDEX arriving while the seal is between registration and its hot cleanup holds no entry for a sealed row (B-tree entry count zero) and the rows answer from the cold side. The second is the invariant's guard: with the guard released right after registration the seal's row removal won the race in three runs, so that mutation does not reproduce; the rule stands on the ordering argument
- the earlier build-time hook stays: a DROP INDEX during the build discards the prepared side file
- a column added between a compaction's preparation and its commit makes the preparation stale: the commit hands it back, the compaction prepares again outside the guard, and the output is published whole with its side file and the new column on every row (a probe panicking on the hand-back path fails the test, so the path is taken)
- an eviction placed after the comparison under the guard and before the commit (no DDL coordination of its own) makes the preparation stale under the publication's write locks: handed back before any change, prepared again, published whole; the same probe fails it, so that path is the one taken
- two indexes whose names differ only in case get two identities, and dropping one leaves the other's
- unit tests: a column records its identity, a zero identity covers nothing, a VERSION 3 file does not open; a reader refused its working reservation is refused before any row and counted; an admitted reader completes through its own buffers once the budget drops to what is held, answering the same positions as the cache path (fails with the fallback removed); a checksum failure mid-walk is the reader's error and it does not advance past the page; a window wider than a page whose second page fails returns nothing and keeps its place, and comes whole from the start once repaired, for the reader and the cursor alike (fails when partial progress is kept)

## Suites

Run locally at this head: `cargo nextest run --no-fail-fast` (5643 passed, 3 skipped), `cargo nextest run --features test-filedb --no-fail-fast` (5636 passed, 3 skipped), `cargo test --test secondary_side_file_test --features test-failpoints -- --test-threads=1` (15 passed), `cargo clippy --all-targets --all-features -- -D warnings` clean.

## Not in this PR

The scanner caller and the fetch (2b-2), the scale measurement, DROP INDEX physical cleanup, backfill of uncovered volumes, merging the inputs' side files at compaction.


semihalev:
Three findings addressed in a5e1f07a:

1. Identity keys match the catalog. `set_index_identity`, `index_identity`, the removal in `remove_index` and `secondary_index_identities` use the name exactly as the index map keys it; no lowercasing anywhere in the identity path, and general name sensitivity is unchanged. Guard: `Mixed` and `mixed` get two identities, and dropping `Mixed` leaves `mixed`'s.

2. Stale preparation outside the guard. `commit_replacement` checks under the read locks that the segments and the manifest order are still the ones the preparation was built on and hands the preparation back untouched (`Err(Box<PreparedReplacement>)`) otherwise; the compaction drops the guard, prepares again outside it (`into_parts` then `prepare_replacement`), compares the identities again and commits. `commit_publication`'s rebuild branch is no longer reached from the compaction. Guard: a column added between the preparation and the commit (the failpoint before the guard spawns and joins the DDL) makes the preparation stale; the output is published whole with its side file and the new column on every row, and a temporary panic on the hand-back path fails the test, so the path is taken.

3. Progress only for a whole window. The walk keeps its cursor local and writes it back only after the window is filled; a window whose first page reads and whose second fails returns the error and stays where it began. Guard: one key over three position pages, the second page corrupted, an 8,192-wide window: `remaining` stays 20,000 after the error, and once repaired the same window comes whole from position 0; the cache-only cursor keeps its place the same way. The test fails when partial progress is kept.

Duplication: the search and the walk are now written once (`equal_with`, `range_with`, `next_window_with`) over a page source trait implemented by the cache-only `Cached` (behind `IndexFile::equal`, `range`, `Cursor`) and by `Reader`; the two differ only in what `with_page` does on a refused page. The cache-only cursor stays for the tests and the harness; 2b-2 decides whether the query path leaves it any caller, and removes it otherwise.

Suites at this head: `cargo nextest run --no-fail-fast` 5643 passed, `cargo nextest run --features test-filedb --no-fail-fast` 5636 passed, the failpoint target 14 passed, clippy clean.

semihalev:
The remaining P2 in 64085bdb: the freshness decision now lives in `commit_publication` itself, under the manifest and segments write locks the publication uses, before anything is changed. A stale preparation comes back as `Err` from there (`rebuild = false` for the compaction; `publish_with` keeps `rebuild = true` for the plain `replace_segments_atomic` path), `commit_replacement` hands the whole replacement back, and the compaction prepares again outside the guard. The read-lock pre-check is gone, so there is no window between a check and the publication.

Guard: after the comparison under the guard, the failpoint evicts the three inputs to metadata-only through `cold_volumes_for_test`, which swaps the segments map without any DDL coordination; the commit finds the preparation stale under its write locks, hands it back, the compaction prepares again and publishes the output whole with its side file and the right counts. A temporary panic on the hand-back path fails this test, so the path is the one taken; the ADD COLUMN test still covers the staleness before the guard.

Suites at this head: `cargo nextest run --no-fail-fast` 5643 passed, `cargo nextest run --features test-filedb --no-fail-fast` 5636 passed, the failpoint target 15 passed, clippy clean.

PR #208, opened 2026-09-20
Serve an equality or a range on an indexed column from each volume's side file, decided before the volume is touched and read window by window

semihalev:
Delivery 2b-2 of the paged secondary index (design 8.5): the query path. A statement that bounds an indexed INTEGER or TIMESTAMP column with one equality, or a lower and an upper bound, now serves each covered volume from its side file instead of the scan, under the identity, budget and visibility rules fixed in 2a and 2b-1. Nothing changes for other predicates, for the hot side, or for a volume the side file cannot serve.

## The path

**Decision per volume, before any data is touched.** `SegmentedTable::side_decision` runs after the zone map and bloom filter pruning and before a metadata-only volume is reloaded, on the captured segment alone: the comparisons must bound one column the catalog indexes (`Table::secondary_index_identities`, read once per statement with the cold view, only when the statement has comparisons), and `ColdSegment::side_for` must find the side file column eligible under the index's current identity (else `ineligible`, scan). The directory decides what it can first, without a page read or a reservation (`IndexFile::candidate_estimate`): no key page in the range skips the volume (`misses`), and an estimate of the candidates (the pages inside the range exactly, the boundary pages by key interpolation, one at least) above one twentieth of the volume's rows sends the volume through the scan (`cost_scans`), so an index the statement will not use costs it no page. A volume that one position page holds (8,190 rows) is scanned about as fast as its pages are read from the file, so it is probed only when the key pages the range spans and the position pages under them are cached or fit the cache beside the walk's reader without an eviction, a missing page counted at its parsed size with the largest raw page of a read on top (`IndexFile::pages_admissible`, else `page_scans`); larger volumes are probed regardless. Otherwise a transient probe reader must be admitted (else `refused`, scan), and the exact candidate count from the boundary key pages must be at most the same share (else `cost_scans`); an exact empty answer skips the volume too. The estimate is a heuristic on the key distribution inside a boundary page: it can send a skewed narrow range to the scan, and the exact count still guards every walk. A TIMESTAMP bound outside the range nanoseconds hold has no exact key: the column goes to the scan, where the typed narrowings leave such a bound to the timestamp comparison itself.

**Fetch, window by window.** The probe hands on a `SidePlan` (file, column, key range); the walk starts when the volume's turn comes, takes the reader's one working reservation then, iterates the reader's own window (4,096 positions, no second copy) and lets the reader go when the walk ends, so the readers of a statement never hold reservations together. It is wired into the four cold read loops: the scanner path (`VolumeScanner::set_side_plan`, where the walk replaces the range and every other rule stays: the visibility bitmap, the committed tombstones under the snapshot rule, the transaction's pending deletions, the hot ids that shadow their cold copies, the typed predicates, the filter, and the projection that materializes only the columns the statement needs), `collect_cold_rows` (the candidate list becomes the walk), `collect_rows_with_limit_unordered` (the position loop becomes the walk, so a LIMIT stops after the window that satisfies it) and the ordered `collect_rows_with_limit` (the authority map and the offset contract kept), so a residual predicate beside the indexed comparison, with a LIMIT, keeps the index. The executor calls that collector again with a larger offset for each batch, so it drains and sorts the walk's positions per volume: its order is the scan's whatever the admission decides per call, and a batch refused after a batch walked never shifts the offset. The positions are admitted against `INDEX_CACHE_MB` beside the reader before they are allocated and charged while they live; refused, the scan reads the same order. Stated cost of that order: every window of the volume is read before the first row, LIMIT or not. A window that fails to read is no window (the reader clears its buffer, the walk retries where it began), and a closed scanner lets its reader and reservation go. On the indexed path the scanner's dictionary prescan and the collection path's whole-column dictionary loads are skipped: a text equality beside the key is checked on the candidates by the row filter. A metadata-only volume with candidates is reloaded after the probe (`reloads`).

**Counters.** `PRAGMA INDEX_READ_STATS` reports probes, misses, candidates, rows, windows, refused, ineligible, cost_scans, page_scans and reloads. Docs: the persistence page describes the path, the PRAGMA page the counters.

The cache-only `Cursor` kept from 2b-1 has no caller in the query path; it stays for the module's tests. Removing it is a small follow-up.

## Tests

`tests/secondary_index_query_test.rs`, every answer against the fixture's formula (20,000 rows, eight rows per key, key uncorrelated with the row order), each guard checked in place:

- equality, closed range, open range and a range past the last key answer the same hot, sealed, reopened; the sealed lookups are probed, none refused or ineligible, every candidate of an untouched volume visible
- (failpoints) a metadata-only volume is probed before any reload: an absent key reloads nothing; a range inside the zone map that the side file finds empty is probed, missed, reloads nothing; a hit reloads the volume once (fails with the reload moved ahead of the probe)
- a volume sealed before its index existed is scanned while later ones are probed
- the hot version shadows the cold copy whether or not it still matches: a sealed row updated in place and one updated out of the key are answered once from hot or not at all, before and after the next seal (fails with the visibility rules skipped on the side path)
- a snapshot keeps seeing the old key after a commit moves a sealed row, and not the new one (same mutation)
- a deleted sealed row is hidden from its own transaction first and from everyone after commit (same mutation)
- `INDEX_CACHE_MB = 0` refuses every working reservation and the answers come through the scan, no candidate walked; an index dropped and created again leaves the old side file ineligible, a volume sealed under the new index is probed while the old one is scanned
- a range over four fifths of the rows goes to the scan by cost and a narrow one to the index
- a LIMIT fetches fewer candidates than the range holds
- (failpoints) a compaction inside the lookup window leaves the answer unchanged and the output is probed; a seal inside the window answers every row once; both assert that the hook fired inside the read and that the volume count changed
- sixteen volumes answer every shape through their side files
- a TIMESTAMP bound past the year 2262 answers every row hot and sealed without a probe, while bounds inside the range are probed (fails with the side decision converting the bound, and with the scan's conversion falling back to zero)
- a residual predicate with a LIMIT is probed and reads at most the key's candidates (fails with the ordered collector's decision removed)
- `INDEX_CACHE_MB = 1` over sixteen volumes: no reader refused, every volume probed (fails with the reservation taken when the plan is set)
- a text equality beside the key on a four-group volume decodes at most one group per column on the scanner path and on the collection path (the vector order), and loads no column whole (fails with the prescan kept, and with the collection path's dictionary loads kept)

Unit and side file lifecycle targets are unchanged and green.

## Measurement

Harness round on the prototype's shapes and matrix (`secondary-query-2b2`), repeated after each review round: main 998a1b3a (2b-1 merged, no query path) against 61eb7d15, both rebuilt from the same probe, indexed on both, three processes per point, every answer checked against the fixture's formula on every iteration. The fixture carries a text column, and two shapes beside the prototype's: `text` (the range plus a text equality) and `limit` (`k = $1 AND ABS(id + 1) > 0 LIMIT 3` on the eight-row key); one phase runs every shape under `INDEX_CACHE_MB = 1`. Warmed p50 ranges in ms; "first" is the first access of the metadata-only start.

| axis / volumes | shape | main | branch | probes | first |
|---|---|---:|---:|---:|---:|
| fragmented / 1 (500k rows) | hit | 2.00 to 2.05 | 0.003 | 1 | 3.7 to 3.8 |
| fragmented / 1 | range (16 keys) | 2.90 to 3.00 | 0.004 | 1 | 0.52 to 0.53 |
| fragmented / 1 | text | 1.95 | 0.004 to 0.005 | 1 | 0.88 to 0.98 |
| fragmented / 1 | limit | 20.7 to 21.8 | 1.38 to 1.52 | 1 | 7.4 to 8.0 |
| fragmented / 1 | wide (a fifth of the keys, scan) | 8.04 to 8.14 | 7.77 to 8.30 | 0 | 11.7 to 11.8 |
| fragmented / 256 | hit | 0.062 to 0.066 | 0.012 | 7 | 0.17 to 0.20 |
| fragmented / 256 | range | 2.84 to 2.95 | 0.107 to 0.109 | 256 | 0.76 to 0.82 |
| fragmented / 256 | text | 2.09 to 2.13 | 0.113 to 0.115 | 256 | 0.84 to 0.97 |
| fragmented / 256 | limit | 2.05 to 2.09 | 1.58 to 1.94 | 8 | 2.8 to 4.1 |
| fragmented / 256 | wide (scan) | 7.67 to 7.91 | 7.92 to 8.26 | 0 | 26.9 to 28.7 |
| growing / 64 (4.2M rows) | hit | 0.48 to 0.51 | 0.004 | 2 | 1.1 |
| growing / 64 | range | 54.2 to 56.6 | 0.025 to 0.027 | 64 | 7.2 to 7.6 |
| growing / 64 | text | 61.4 to 62.0 | 0.029 to 0.032 | 64 | 5.4 to 5.9 |
| growing / 64 | limit | 91.9 to 92.9 | 32.3 to 32.7 | 8 | 38 to 43 |
| growing / 64 | wide (scan) | 95.9 to 99.1 | 97.5 to 100 | 0 | 120 to 123 |
| `INDEX_CACHE_MB = 1`, fragmented / 256 | hit | 0.059 to 0.061 | 0.066 to 0.067 | 0 (256 page scans) | 0.45 to 0.55 |
| `INDEX_CACHE_MB = 1`, fragmented / 256 | range | 2.85 to 2.94 | 3.05 to 3.19 | 0 (256 page scans) | 14.1 to 16.8 |
| `INDEX_CACHE_MB = 1`, fragmented / 256 | text | 2.10 to 2.11 | 2.24 to 2.33 | 0 (256 page scans) | 16.8 to 20.0 |
| `INDEX_CACHE_MB = 1`, fragmented / 256 | wide (scan) | 7.65 to 7.86 | 7.87 to 8.54 | 0 | 18.7 to 23.5 |
| `INDEX_CACHE_MB = 1`, growing / 64 | range | 54.2 to 56.6 | 1.19 to 1.20 | 64, none refused | 5.8 to 11.4 |

The 16 volume points and the refused phase match the earlier rounds and are in the harness tables. Every shape whose result is a few rows moves from the scan to a few page reads, the text residual with them: two to three orders of magnitude on one volume and on 64. At 256 volumes a range costs 256 probes and about 0.1 ms, a thirtieth of the scan; routing above the volumes is not needed for this delivery. The wide shape is decided from the directory without a page in every column (probes 0) and the branch's scan overlaps main or sits a few percent above it. Under a 1 MiB budget every volume is served in turn with no refusal at 16 and 64 volumes, where the numbers equal the 16 MiB ones; at 256 small volumes the pages do not fit that budget, so the page-aware decision sends every volume through the scan (256 page scans) with no file read: the range is 3.05 to 3.19 against 2.85 to 2.94, the text 2.24 to 2.33 against 2.10 to 2.11, the hit 0.066 to 0.067 against 0.059 to 0.061, the difference being the admission check per volume (a directory lookup and a cache lookup, 256 times), where the earlier rounds paid a file read per volume (3.8 ms); with the default budget the same table probes (0.107 to 0.109). One cost stays and is stated: the residual LIMIT shape takes the index (8 probes, 8 candidates) but the ordered collector's authority map still walks every cold id for the newest-copy rule, and its sorted order reads every window of a volume before the first row, so it gains 2 to 15 times rather than orders of magnitude (1.4 to 1.5 ms on 500k rows, 32 to 33 ms on 4.2M); moving that rule onto the segment visibility bitmap the scanner uses is a follow-up. With every reservation refused the answers come through the scan at main's cost. Reopen and seal times overlap on both sides. Full tables and the report are in the harness directory.

## Review round 1 (e41283d9)

- **TIMESTAMP bound overflow (P1)**: the side decision multiplied the seconds unchecked when `timestamp_nanos_opt` failed. Now a bound with no exact key sends the column to the scan. The scan had the same defect on main in three places (the typed pre-filter of the collection path fell back to zero, so `ts < '2500-01-01'` answered no rows on a sealed volume; the sorted binary search and the equality lookup multiplied unchecked and panicked in debug); the three now saturate through one helper, and the test covers both.
- **Second window outside the budget (P2)**: the walk iterates the reader's window in place.
- **Readers reserving before their turn (P2)**: the probe uses a transient reader; the walk's reader is taken at the first row of its volume and released when the walk ends. Sixteen volumes under a 1 MiB budget: zero refusals.
- **Dictionary prescan before the indexed read (P2)**: skipped when a plan is set, on the scanner path and in the collection path.
- **Residual filter with LIMIT (P2)**: the ordered collector takes the side decision too.
- **Race tests not firing the hook (P2)**: the scanner path calls `cold_volumes_taken` after the view is captured, and both tests assert the hook fired and the volume count changed.
- Identities are collected only when the statement has comparisons; the per-row counter is batched per volume.

## Review round 2 (fd4b6123)

- **Residual LIMIT batches (P1)**: the ordered collector's walk order differed from the scan's, so a batch refused after a batch walked applied the executor's offset to another order and a row came twice. The walk's positions are sorted per volume now; the test refuses the second batch from the hook and gets the row once (`[904, 904]` with the sort removed).
- **Saturated TIMESTAMP bound (P1)**: round 1's saturation made `ts = '3000-01-01'` match a row at the end of the range and strict bounds drop the end rows. The five typed narrowings leave an unconvertible bound to the timestamp comparison; the test holds rows at both ends of the range, hot and sealed, SELECT and COUNT.
- **Index I/O before the scan decision (P2)**: the directory decides first (`candidate_estimate`), with no page and no reservation; the wide range reads no page now (probes 0 in its test).
- **Failed window served (P2)**: the reader clears its buffer on a failed window; the retry yields 5,904 positions once, not 9,998.
- **close keeps the reservation (P2)**: close drops the side state; the reservation is released without the drop.

## Review round 3 (8153ced1)

- **Position vector outside the budget (P2)**: the ordered collector's sorted positions are admitted with `try_reserve` before allocation and charged for their lifetime; refused, the scan reads the same order, counted under `refused`. `reader_bytes` computes the reader's reservation in one place. Test: a budget with room for the reader and not for the positions answers through the scan with one refusal; a wide budget's peak carries reader plus positions.
- **Remainder after a failed refill (P2)**: a refill failing after a served window reset the reader's buffer but not the walk's cursor, so `remaining` underflowed. The cursor is reset with the buffer. Test: 4,096 positions served, CRC error on the next page, remaining 15,904, and every position once after the repair.

## Review round 4 (cc4c1c33)

- **Narrow range under a small budget (performance acceptance)**: 256 volumes of two thousand rows under 1 MiB paid a file read per volume on every statement (3.8 ms against the scan's 2.9). A small volume is now probed only when its pages are cached or fit the cache without an eviction; else it is scanned (`page_scans`). Test: sixteen small volumes under a budget with room for a reader and a few volumes' pages answer with the volumes that fit probed and the rest scanned, no refusal, and every one probed once the budget has room; page_scans stays zero with the rule removed. The re-measurement is in the table.

## Review round 5 (26b714bb)

- **Admission counted encoded lengths (P2)**: a key entry takes sixteen bytes parsed for twelve on disk, and a read holds the raw page while it parses, so a budget with room for the encoded pages passed the check and then loaded and evicted the same two pages on every lookup. The check counts a missing page at its parsed size and adds the largest raw page. Test: a one-page volume under a budget with room for the reader, the parsed pages and one raw page serves three lookups with two loads and no eviction; short of the raw page it goes to the scan with nothing loaded or evicted. Fails with the encoded sum, and with the raw page left out.

## Suites

Run locally at 61eb7d15: `cargo nextest run --no-fail-fast` (5663 passed, 3 skipped), `cargo nextest run --features test-filedb --no-fail-fast` (5656 passed, 3 skipped), `cargo test --test secondary_index_query_test --test secondary_side_file_test --features test-failpoints -- --test-threads=1` (21 and 16 passed), `cargo clippy --all-targets --all-features -- -D warnings` clean.

## Not in this PR

Serving ORDER BY through the side file (`scan_top_k`), the filtered aggregation fast path (`COUNT(*)` with a WHERE), IN lists, the ordered LIMIT collector's newest-copy rule on the segment visibility bitmap, DROP INDEX physical cleanup, backfill of uncovered volumes, merging side files at compaction, and the removal of the cache-only cursor.


semihalev:
Round 1 fixes are in, on top of e41283d9; every guard was checked in place by mutating the fix and watching the test fail.

- **P1 TIMESTAMP**: a bound `timestamp_nanos_opt` cannot hold sends the column to the scan. The scan itself had the same overflow on main in three places (typed pre-filter fallback to zero, sorted binary search, equality lookup): `ts < '2500-01-01 00:00:00'` answered no rows on a sealed volume at 998a1b3a too. They saturate now through one helper. Test: 600 hourly rows, `ts < '3000-01-01'` and `ts > '1000-01-01'` answer every row hot and sealed with no probe; the bounds inside the range are probed. Fails with the decision converting the bound (probes appear) and with the fallback to zero (no rows).
- **P2 second window**: the walk iterates the reader's own window; no copy.
- **P2 early reservations**: the probe reads the boundary pages through a transient reader; the walk's reader is taken at the volume's first row and released at the walk's end. Test: sixteen volumes under `INDEX_CACHE_MB = 1`, zero refusals, sixteen probes; sixteen refusals with the reservation moved back to plan time.
- **P2 dictionary prescan**: skipped on the indexed path in the scanner and in the collection path. Test: four-group volume with a text equality beside the key, at most one group per column decoded on the scanner path and on the collection path (vector order), no column loaded whole; five groups with the prescan back, and a whole column resident with the collection path's loads back.
- **P2 residual filter with LIMIT**: the ordered collector takes the side decision, authority map and offset contract unchanged. Test: `k = $1 AND ABS(id + 1) > 0 LIMIT 3` probes and reads at most the key's candidates; no probe with the decision removed.
- **P2 race tests**: the scanner path fires `cold_volumes_taken` after the view is captured; both tests assert the hook fired inside the read and the volume count changed; both fail with the hook call removed.
- Identities are read only when the statement has comparisons; the per-row counter is batched per volume.

Measurement repeated on both binaries (main 998a1b3a, fd4b6123) with the text residual, the residual LIMIT and a 1 MiB phase added; the table is in the body. The earlier numbers stand, and the text residual joins them (1 volume 1.8 to 2.1 ms to 0.005 ms, 256 volumes 2.1 to 2.5 to 0.10 ms). Two costs remain and I state them rather than hide them: the residual LIMIT shape now probes and reads eight candidates, but the ordered collector's authority map still walks every cold id (1.6 ms on 500k, 48 to 51 ms on 4.2M, against 21 to 24 and 99 to 108 on main); moving that rule onto the segment visibility bitmap is the follow-up, not a second visibility implementation. Under 1 MiB there is no refusal at 16, 64 or 256 volumes, and 16 and 64 equal the 16 MiB numbers; at 256 the boundary pages no longer stay cached and each probe reads from the file, 3.8 ms against the scan's 2.9 to 3.3. The cost decision does not know the cache's pressure; I list it beside upper routing as an open item.

The snapshot LIMIT note (a hot row moved out of the key after the snapshot started) is main's raw-membership rule in the unordered collector, as you classified it; I left it out of this PR and will take it with the LIMIT paths follow-up. Docs and the PR body are updated; suites in the body.

semihalev:
Round 2 fixes are in on top of fd4b6123; each guard was checked in place by reverting its fix and watching the test fail.

- **P1 residual LIMIT batches**: the ordered collector drains and sorts the walk's positions per volume, so its order is the scan's whatever the admission decides per call; the authority map and the offset contract are unchanged. Test: the 100k fixture, the first batch walks and arms the hook that sets the page budget to zero for the second; result `[904]`, and `[904, 904]` with the sort removed.
- **P1 TIMESTAMP**: the saturation is gone. The five typed narrowings (aggregate pre-filter, sorted binary search, equality lookup, scanner predicates, stop key) leave a bound nanoseconds cannot hold to the timestamp comparison itself. Test: rows at i64::MAX and i64::MIN nanoseconds plus 2020, bounds in 3000 and 1000, hot and sealed, SELECT and COUNT; with a saturating conversion back, `COUNT(*) WHERE ts = '3000-01-01'` is 1 again.
- **P2 index I/O before the decision**: the directory decides first, without a page read or a reservation: no key page in the range is a miss, and an estimate (interior pages exact, boundary pages by key interpolation) above the scan share is a cost scan. The exact count from the boundary pages still guards every walk, so the heuristic can only send a skewed narrow range to the scan, never admit a wide one. Test: the wide range is decided with zero probes. The 1 MiB / 256 volume wide shape is in the re-measurement below.
- **P2 failed window**: the reader clears its buffer on a failed window; the same walk, after the file is repaired, yields 5,904 positions once (9,998 with the clear removed).
- **P2 close**: close drops the side state, releasing the reader, the reservation and the batched counter (125,592 bytes stay charged with the drop removed).

Measurement repeated on both binaries (main 998a1b3a, 8153ced1), same matrix and shapes; the table is in the body. The wide shape now reads no page in any column: at 1 MiB and 256 volumes it is 7.92 to 8.02 ms against main's 7.75 to 7.88, where fd4b6123 paid 11.9 to 12.2. The other numbers stand. The two stated costs remain as before: the residual LIMIT's authority map over every cold id (1.4 ms on 500k, 33 to 39 ms on 4.2M, 2 to 15 times over main) and the narrow range under 1 MiB at 256 volumes, whose boundary pages no longer stay cached (256 file reads, 3.8 ms against the scan's 2.9); both are in the follow-up list. Suites at 8153ced1: memory 5659, filedb 5652, failpoint targets 18 and 16, clippy clean.

The Windows failure in `test_classification_caching` is on an unchanged test, an `Arc::ptr_eq` on a shared cache under `cargo test`; I have not touched it in this PR and will look at it apart.

semihalev:
Round 3 fixes are in on top of 8153ced1, both checked in place by reverting the fix and watching the test fail.

- **P2 position vector**: the ordered collector admits its sorted positions against the page budget with `try_reserve` before allocating them, the reservation lives as long as the vector, and a refusal sends the volume through the scan in the same physical order (counted under `refused`). The reader's working reservation is computed in one place now (`reader_bytes`). Test: with room for the reader and not for the 3,200 bytes of positions the query answers through the scan with one refusal; with a wide budget the ledger's peak carries reader plus positions. With the reservation zeroed the refusal count drops to zero. The cost you named stands and is written into the body: on that path every window of the volume is read before the first row, LIMIT or not.
- **P2 remainder after a failed refill**: the walk's cursor is reset together with the reader's cleared buffer, so a refill failing after a served window keeps the remainder (15,904 after 4,096 served) and, repaired, yields every position once. The new unit test covers exactly that sequence; with the reset removed it overflows in debug.

The narrow range under 1 MiB at 256 volumes is unchanged: the pages cannot stay cached in that budget, so each probe is a file read. It stays an open item with the page cache aware decision; I have not claimed it closed.

Measurement repeated on both binaries (main 998a1b3a, cc4c1c33); the numbers match round 2 within their ranges, the growing / 64 point wider on both sides this time. Suites at cc4c1c33: memory 5661, filedb 5654, failpoint targets 19 and 16, clippy clean.

semihalev:
The remaining acceptance item is in, on top of cc4c1c33, checked in place by removing the rule and watching the test fail.

- **Index page cost before the decision**: a volume that one position page holds (8,190 rows) is scanned about as fast as its pages are read from the file, so it is probed only when the key pages the range spans and the position pages under them are cached or would fit the cache beside the walk's reader without an eviction; otherwise it goes through the scan, counted under a new `page_scans` column of `INDEX_READ_STATS`. The check is a cache lookup on the directory's page numbers, before any page read or reservation. Larger volumes are probed as before, a file read being far cheaper than their scan. Stated limit: the check counts room without eviction, so a cache full of another table's useful pages sends small volumes to the scan rather than evicting for them.
- Test: sixteen small volumes under a budget with room for a reader and a few volumes' pages; the volumes that fit are probed and the rest scanned, no refusal, and every one is probed once the budget has room. With the rule removed page_scans stays zero.

Measurement repeated on both binaries (main 998a1b3a, 26b714bb). Under 1 MiB at 256 volumes the range is 2.96 to 3.04 ms against main's 2.84 to 2.96 (3.76 to 3.80 before), the hit 0.062 to 0.065 against 0.061 to 0.062, the text 2.20 to 2.22 against 2.14 to 2.17, all 256 page scans and no file read; with the default budget the same table probes as before (0.107 to 0.110). Everything else matches the previous rounds within its ranges. Suites at 26b714bb: memory 5662, filedb 5655, failpoint targets 20 and 16, clippy clean. The `page_scans` column is in the PRAGMA docs and the persistence page.

semihalev:
The admission fix is in on top of 26b714bb, checked in place with two mutations.

- **Parsed size and raw page**: a missing page is counted at its parsed size (sixteen bytes per key entry, four per position) and the largest raw page among the missing ones is added on top, since a read holds the raw bytes while it parses them; the walk's reader stays in the sum as before. Your three budgets now decide as the table says they should: the encoded sum and the parsed sum are refused, the parsed sum plus the raw page is admitted and stays warm.
- Test: a volume of one position page of rows with a key per row; under a budget with room for the reader, the parsed pages and one raw page, three lookups probe with two loads and no eviction; under a budget short of half a raw page the lookups go to the scan with nothing loaded or evicted. The guard asserts `loads` and `evictions` beside `page_scans` and `probes`. With the encoded sum back, and with the raw page left out, the short budget is admitted and the guard fails.

Suites at 61eb7d15: memory 5663, filedb 5656, failpoint targets 21 and 16, clippy clean. Measurement repeated on both binaries (main 998a1b3a, 61eb7d15): the numbers match the previous round within their ranges; under 1 MiB at 256 volumes the range is 3.05 to 3.19 against main's 2.85 to 2.94 with no file read, the remaining difference being the admission check itself, 256 times; the table in the body says so.

PR #209, opened 2026-09-21
Keep the HNSW graph true to the table across sealed updates, older snapshots, inferred creation, reopen and NULL vectors

semihalev:
The index path audit at 61eb7d15 reproduced five wrong answers on the HNSW path, none introduced by #208 and all older than it. This PR fixes them with a test each, failing before its fix.

## The defects and the fixes

- **A sealed row's update left its old vector in the graph.** The cold claim carries no hot read version, so commit took the insert branch of the index maintenance, and the graph ignored an insert of a live row id. A live node whose vector changed is now tombstoned and connected again as a reinsert; the same vector again stays a no-op (a snapshot, a WAL replay). `src/storage/index/hnsw.rs`.
- **The fast path answered an older snapshot from the current graph.** An updated row came back with the graph's distance beside the snapshot's vector, and an inserted row the snapshot must not see was chosen and then dropped, leaving the result empty. The graph is used only when the table's index view epoch is available (no local changes, no snapshot isolation, no commit publishing), checked again after the rows are fetched; otherwise the brute-force top-k over the rows runs. A segmented table delegates the epoch to its hot store, so file tables take the graph as memory tables do. `src/executor/index_optimizer.rs`, `src/storage/volume/table.rs`.
- **An index created without USING after a seal held no sealed row.** The hot store inferred HNSW for the vector column, but the sealed rows were added only when HNSW had been asked for. The created type decides now, for both creation entry points. `src/storage/volume/table.rs`.
- **Reopening rebuilt the graph from the wrong column after a dropped column.** Recovery read each index column at its current ordinal; a volume sealed before the drop holds it elsewhere. Each volume's column mapping resolves the column now. `src/storage/mvcc/engine.rs`.
- **NULL vectors broke the ORDER BY rules.** `NULLS FIRST` was served closest first, and a graph short of LIMIT rows answered without the NULL tail. `NULLS FIRST`, and fewer graph answers than asked for, leave the order to the rows. `src/executor/index_optimizer.rs`.
- **`V` missed the index on `v`.** The single-column index lookup compared the column's spelling; it ignores ASCII case now, as column resolution does. `src/storage/mvcc/version_store.rs`.

## Review round 1 (70ecb4b3)

- **A rename before the checkpoint broke the rebuild (P1, introduced here)**: the graph was rebuilt inside the WAL replay, before the volume mappings were recomputed against the recovered schema, so a column renamed before the checkpoint resolved to no volume column and the sealed rows were left out. The rebuild runs after the mappings are current. Test: a checkpointed rename reopened with a hot row keeping the graph non-empty (answers (3, 49) with the rebuild moved back).
- **The parallel batch path skipped a changed vector (P1)**: a bulk update above five thousand rows takes `insert_batch_inner`, which skipped a live row id whatever its vector. It tombstones the node and connects the new vector as a reinsert, as the single insert does. Test: six thousand sealed vectors updated in one statement, the returned distance is the distance of the returned vector at three probes (distance 15 for a vector 99,000 away with the skip back).
- **A sealed vector set to NULL kept its node (P1)**: the value reached the index as an addition with no graph entry and was skipped. A value with no graph entry takes the row's live node out, on `add`, `add_batch` and `add_batch_slice`. Test: one row set to NULL answers the other at distance 9, and three thousand of six thousand set to NULL answer a row still holding a vector at that vector's distance (the NULL row at distance 0 with the skip back).
- The index view epoch is read only when a graph is there to use, so a brute-force query on a table without one takes no lock for it.

## Review round 2 (d653566f)

- **A refused batch took rows out (P1)**: `add_batch` and `add_batch_slice` tombstoned the rows whose value has no graph entry while collecting the batch, before the unique validation, so a batch refused for a duplicate had already dropped those nodes. The rows to take out are tombstoned only after the whole batch is validated, right before the insert. Unit test on both entry points: a refused batch leaves row 1 in the graph and its vector unique, an accepted one takes it out (fails with the tombstones back before the validation).

## Review round 3 (025b2638)

- **A departing row's vector counted as a conflict (P2)**: the batch validation checked newcomers against the graph as it was, so a batch taking row 1 to NULL and giving its vector to row 3 was refused although its final state holds no duplicate. The rows the batch takes out are excluded from the validation on both entry points; a duplicate inside the batch is still refused, and the removals still apply only after the whole batch is validated. The unit test now also accepts a batch that reuses a departing row's vector and refuses the vector afterwards (fails with the departures counted).

## Tests

`tests/hnsw_lifecycle_test.rs`, ten tests (the three of the review round listed above, and): a sealed vector's update (distance zero to its own new vector, then updated again hot and sealed again), an older snapshot after another transaction's update and after its insert (memory and file), an inferred index created after a seal (with a hot row added so an empty graph cannot hide behind the fallback), a reopen after a dropped column (same guard), NULL vectors with and without the index, and the column's spelling in EXPLAIN and in the answer. Each fails with its fix reverted: the sealed update and the sealed snapshot test with the graph change, the two snapshot tests and the NULL test with the optimizer changes, the inferred index test with the population gated on the requested type, the reopen test with the ordinal read, the spelling test with the exact comparison.

## Measurement

The write cost of the replacement: a probe (fast profile, three processes per point) seals 6,000 and 60,000 eight-dimensional vectors under an HNSW index and updates every row three times; the first update replaces sealed vectors, the next two hot ones. First update: main 7.9 to 9.1 ms against 22.3 to 24.5 on 6,000 rows, and 103 to 164 against 230 to 270 on 60,000; the two hot updates overlap on both sides (12.8 to 15.0 against 15.1 to 16.2, and 166 to 226 against 187 to 197). The difference on the first update is the reconnection of every changed vector, which main skipped and answered wrong for (its graph kept the old vectors); on the hot updates both sides already removed and re-added the node. No allocation count was taken.

## Docs

The vector search page says an index created without `USING` on a VECTOR column is HNSW and holds the sealed rows, that a column holds one index whose metric decides (the page claimed several metrics per column; the runtime refuses the second index), and when the rows decide instead of the graph (local changes, snapshot isolation, a commit publishing, `NULLS FIRST`, a LIMIT past the graph's rows).

## Not in this PR

The two PK `IN` defects the audit found (`id IN (1.0)` losing the match; `id IN (0, 1) LIMIT 1` losing the row to the absent id) are a separate change. The scalar paths still outside the side index (filtered COUNT/SUM, ORDER BY top-k, simple UPDATE/DELETE, window partition fetch, secondary JOIN/EXISTS probes on file tables) stay as listed in the audit. HNSW memory (the resident graph outside the page budgets, whole-row candidate fetch, recovery batching) is not measured here.

## Suites

Run locally at the round 3 head: `cargo nextest run --no-fail-fast` (5674 passed, 3 skipped), `cargo nextest run --features test-filedb --no-fail-fast` (5667 passed, 3 skipped), `cargo nextest run --test hnsw_lifecycle_test --test vector_test` (66 passed, and 66 with `--features test-filedb`), `cargo nextest run --lib hnsw` (8 passed), `cargo clippy --all-targets --all-features -- -D warnings` clean.


semihalev:
Round 1 fixes are in on top of 70ecb4b3, each checked in place by reverting its fix and watching the test fail.

- **P1 rename before the checkpoint**: the rebuild ran inside the WAL replay, before the volume mappings were recomputed against the recovered schema, so the renamed column resolved to no volume column. `populate_hnsw_from_segments` runs after `invalidate_mappings` now; the dropped-column guard stays and a checkpointed-rename guard joins it (a hot row keeps the graph non-empty so the fallback cannot hide it). With the rebuild moved back the test answers (3, 49) again.
- **P1 batch path**: `insert_batch_inner` skipped a live row id whatever its vector; it now tombstones the node and connects the new vector as a reinsert, exactly as the single insert does, and the same vector again is still a no-op. Test: six thousand sealed vectors updated in one statement, the returned distance is the distance of the returned vector at three probes (to f32 precision) and zero at the new vector; with the skip back the distance is 15 for a vector 99,000 away.
- **P1 vector to NULL**: a value with no graph entry (NULL, or the wrong dimension) takes the row's live node out, on `add`, `add_batch` and `add_batch_slice`; `remove` shares the helper. Test: one sealed row set to NULL answers the other row at distance 9, and three thousand of six thousand set to NULL answer the first row left at distance 3,000; with the skip back the NULL row comes at distance 0.
- **Epoch only on the graph path**: the index view epoch is read only when an HNSW index is there to use, so a brute-force query takes no lock for it.

On the write cost of the replacement: a probe (fast profile, three processes per point) seals 6,000 and 60,000 eight-dimensional vectors under an HNSW index and updates every row three times; the first update replaces sealed vectors, the next two hot ones. First update: main 7.9 to 9.1 ms against 22.3 to 24.5 on 6,000 rows, and 103 to 164 against 230 to 270 on 60,000; the two hot updates overlap on both sides (12.8 to 15.0 against 15.1 to 16.2, and 166 to 226 against 187 to 197). The difference on the first update is the reconnection of every changed vector, which main skipped and answered wrong for (its graph kept the old vectors); on the hot updates both sides already removed and re-added the node. No allocation count was taken.

The two source-only notes (a seal between candidate classification and the deferred hot fetch, and the missing old cold vector in the index undo log) are inherited and stay open; I have not touched them here.

semihalev:
Round 2 fix is in on top of d653566f, checked in place with the tombstones moved back before the validation.

- **P1 a refused batch took rows out**: both batch entry points removed the rows whose value has no graph entry while collecting the batch, before the unique validation, so a batch refused for a duplicate had already dropped those nodes and their unique entries. The rows to take out are collected during the walk and tombstoned only after the whole batch is validated, right before the insert, so a refused batch leaves the graph as it was.
- Test (unit, both `add_batch` and `add_batch_slice`): a unique index with `[1,0]` and `[2,0]`, a batch of row 1 to NULL and row 3 repeating `[2,0]` is refused; row 1 still answers nearest to `[1,0]` and a row 4 with `[1,0]` is still refused; an accepted batch then takes row 1 out. With the tombstones back before the validation the nearest is row 2.

Suites at this head: memory 5674, filedb 5667, targets 66 and 66 with filedb, lib hnsw 8, clippy clean.

semihalev:
Round 3 fix is in on top of 025b2638.

- **P2 departing vectors counted as conflicts**: the rows a batch takes out are collected before the validation and excluded from it through the `ignored_row_ids` the duplicate check already takes, on both entry points; a duplicate inside the batch is still refused, and the removals still apply only after the whole batch is validated. Your probe (row 1 to NULL, row 3 taking `[1,0]` in one batch) is accepted and answers row 3 at distance 0, a later row taking `[1,0]` is refused, and the refused-batch case of the previous round still holds; the unit test covers both entry points and fails with the departures counted.

Suites at this head: memory 5674, filedb 5667, targets 66 and 66 with filedb, lib hnsw 8, clippy clean.

semihalev:
Done on top of 246479e5: the departing set is built only when the index is unique and a row is leaving, on both entry points; the validation takes it as an option. Lib hnsw 8, the two targets 66, clippy clean; the sources are otherwise unchanged. Coverage is CI's to say.

PR #210, opened 2026-09-21
Read a primary key IN member as the comparison does, and count a LIMIT on the rows found

semihalev:
The index path audit at 61eb7d15 found two wrong answers on the primary key `IN` fast path, both older than #208. This PR fixes them with a test each, failing before its fix.

## The defects and the fixes

- **`id IN (1.0)` matched nothing while `id = 1.0` matched.** The fast path collected integer members only and dropped the rest. A float member that holds an integer is that key, one that does not matches nothing, as the comparison reads them; a member of any other type leaves the statement to the comparison itself. `src/executor/index_optimizer.rs`.
- **`id IN (0, 1) LIMIT 1` came back empty.** The id list was cut to the LIMIT before the fetch, so a member with no row used up the limit. On all three routes (the IN list, the IN subquery, the set an EXISTS becomes) every member is taken, the rows are fetched a bounded batch at a time and trimmed after.

## Review round 1 (dc50c094)

- **Large float keys still dropped (P1)**: the list route accepted integral floats below 2^53 only, and the subquery route integers only. All three routes read a float member with `lossless_float_key`, the conversion the primary key comparison already uses. Test: keys at plus and minus 2^53 as float members on the list and the subquery routes, against `=`.
- **The EXISTS set path still counted missing members toward the LIMIT (P1)**: it chose members up to the limit before fetching and cut the id list after. Every member is taken and the rows are trimmed after the fetch, as on the other two routes; the earlier claim that its members are always matching rows was wrong. Test: a file database, two hundred members without a row beside one with, the EXISTS shapes with and without a second inner predicate, five times each.
- **The whole list fetched to trim it (P2)**: the three routes fetch a bounded batch of ids at a time, doubling from the limit plus the offset, and stop once the rows found satisfy them. Test: LIMIT 1 over twenty thousand members of four sealed volumes with a long text column, after a reopen: the first volume's groups alone are decoded (12 before, at most 3 now, by `GROUP_CACHE_STATS`).

## Review round 2 (d4315fa1)

- **The batch doubled without a bound (P2)**: a run of members without a row ahead of the ones with made the batch after it as long as the rest of the list, so a LIMIT 1 read every remaining row. A batch starts from the rows still needed and never exceeds 1,024 ids, whatever the list holds. Test: 32,752 members without a row ahead of twenty thousand with, over four sealed volumes with a long text column, LIMIT 1 decodes the first volume's groups alone, at most 3 by `GROUP_CACHE_STATS` (12 without the bound).

## Review round 3 (40f5259d)

- **The two decode measurements raced under `cargo test` (P2)**: both read the process-wide decoded-group counter, and CI's single-process run let one test's decodes land in the other's difference (6 instead of 3). They take one static guard for their whole body, fixtures included; the asserted counts are unchanged.

## Tests

`tests/pk_in_lookup_test.rs`, eight tests: the four of the rounds above, a float member against `=`, a member no key equals, a LIMIT and an OFFSET over members with no row, a deleted key, the subquery shape with and without the row, and the EXISTS set shape. The list and subquery LIMIT tests fail with the early cut back; the float test fails with integer members only; the three round-one tests fail on the sources before it.

## Suites

Run locally: `cargo nextest run --no-fail-fast` (5671 passed, 3 skipped) and `cargo nextest run --features test-filedb --no-fail-fast` (5664 passed, 3 skipped) at 40f5259d, whose sources the round 3 head keeps unchanged; at the round 3 head `cargo test --test pk_in_lookup_test` three times in one process (8 passed each), `cargo nextest run --test pk_in_lookup_test` (8 passed), `cargo clippy --all-targets --all-features -- -D warnings` clean.


semihalev:
Round 1 fixes are in on top of dc50c094, each test failing on the sources before it.

- **P1 large float keys**: all three routes (list, subquery, EXISTS set) read a float member with `lossless_float_key`, the conversion the primary key comparison uses; keys at plus and minus 2^53 match with IN as with `=`, and `id IN (SELECT 1.0)` answers.
- **P1 EXISTS set path**: my earlier claim that its members are always matching rows was wrong; the path chose members up to the limit before fetching and cut the list after. Every member is taken and the rows are trimmed after the fetch. Test in a file database with two hundred members without a row beside one with, five times per shape.
- **P2 whole-list fetch**: the three routes fetch a bounded batch of ids at a time, doubling from the limit plus the offset, and stop once the rows found satisfy them. Your shape (twenty thousand members of four sealed volumes with a long text column, reopened, LIMIT 1) decodes the first volume's groups alone, at most 3 by `GROUP_CACHE_STATS` against 12 before; the test asserts it.

Suites at this head: memory 5670, filedb 5663, clippy clean.

semihalev:
Round 2 fix is in on top of d4315fa1.

- **P2 unbounded doubling**: a batch now starts from the rows still needed and never exceeds 1,024 ids, so a run of members without a row ahead of the ones with makes the fetch after it a bounded batch too. Your shape (32,752 absent ids ahead of twenty thousand present, four sealed volumes with a long text column, reopened, LIMIT 1) decodes the first volume's groups alone, at most 3 by `GROUP_CACHE_STATS` against 12 before; the test asserts it and fails without the bound.

Suites at this head: memory 5671, filedb 5664, the target 8 and 8 with filedb, clippy clean.

semihalev:
Round 3 fix is in on top of 40f5259d: the two tests that measure the process-wide decoded-group counter take one static guard for their whole body, fixtures included, so `cargo test` in one process cannot let one test's decodes land in the other's difference; under nextest nothing changes. The counts they assert are unchanged (at most 3 decoded groups each).

The sources are unchanged since 40f5259d; the target passes three times in one process under `cargo test` and under nextest, clippy clean.

PR #212, opened 2026-09-21
Let a limited join probe the index of a persistent table while its rows are still hot, the table answering each probe or ending the join

semihalev:
A join with a LIMIT on a persistent table always went to the hash path, because the executor asked for the raw index and a table that keeps sealed rows outside its hot index cannot hand that out. The benchmark's self join on `age` with LIMIT 100 built a hash table over the whole table for a hundred rows on file mode.

This is stage 1 of the plan: the join asks the table for the ids of each key instead of the index, and the table answers only while it can.

## The contract

- `Table::equality_candidates(column, key, max, out)` appends at most `max` row ids for the key and answers `Copied`, `OverCap` (nothing appended) or `None` (the table cannot answer this way). An absent key is `Copied` with nothing appended, which is a different outcome from a refusal.
- The hot table answers from its B-tree index through the capped probe, which reads the bucket's length before copying anything.
- The segmented table answers under the checks its group walk already uses: no snapshot, no volume, no destructive publication in flight, a quiet index epoch, and the same seal generation, segment set and epoch once the ids are taken. Anything else drops the ids and answers `None`. The probe runs without the seal fence, so a query never waits out a seal.
- A new `IndexLookupStrategy::TableEquality` carries the column to the streaming index join. A probe the table cannot answer ends the join and marks it; the executor drops what was joined so far and runs the whole join once on the hash path. The cap is 1024 ids per probe.
- The strategy is chosen only when the raw index is not available (the memory engine keeps its path) and only for a join whose LIMIT the executor pushes; the batch join cannot end early, so it never sees the strategy.
- EXPLAIN reports the join as Index Nested Loop, as for the raw index.

## Measurements

Harness probe (untracked, in the plan repo): the benchmark's users table, 10,000 rows, seed 42, the benchmark's self join with LIMIT 100, 300 timed runs after 20 warm runs, three interleaved rounds per binary and mode, `--profile fast`. Ranges are over the three rounds. The checksum is over the joined rows; the hash path picks a different hundred pairs than the index path, which is why the sealed checksum differs.

| mode | bin | join p50 us | join p90 us | insert p50 us | insert p99 us | checksum |
|---|---|---|---|---|---|---|
| memory | main | 16.5 to 20.7 | 18.5 to 22.9 | 1.1 to 1.5 | 2.2 to 3.3 | 20627 |
| memory | branch | 16.9 to 20.9 | 17.5 to 21.5 | 1.2 to 1.4 | 2.5 to 3.0 | 20627 |
| file, rows hot | main | 183.5 to 235.0 | 189.0 to 237.6 | 4.2 to 4.7 | 9.4 to 10.5 | 20900 |
| file, rows hot | branch | 15.2 to 17.6 | 15.7 to 18.2 | 3.8 to 3.9 | 8.0 to 8.2 | 20627 |
| file, sealed | main | 599.5 to 750.3 | 623.5 to 755.7 | 3.8 | 7.5 to 7.6 | 20900 |
| file, sealed | branch | 650.0 to 711.1 | 682.9 to 763.1 | 3.6 to 4.2 | 7.4 to 10.2 | 20900 |

The hot file join answers in the memory engine's time and returns the same rows as it. The memory path is unchanged, and so is the INSERT latency; the write path is not touched. Sealed rows stay on the hash path, within the round to round spread.

## Tests

`tests/join_equality_candidates_test.rs`: the hot file table against the memory engine before and after a seal, an outer key with no inner rows on INNER and LEFT joins, the rest of the ON clause applied to probed rows; under `test-failpoints` (new hook `after_join_probe_admitted`, added to the CI failpoint step): the probe count of a bounded and an unbounded join, an over-cap key refused before anything is copied, a join short of its limit continuing from the row its first chunk ended on, and a seal, a commit and a truncate between a probe's admission and its read. Each guard was checked to fail against a mutation of the code it covers:

- no re-check after the probe: the seal, commit and truncate guards fail
- volumes ignored before and after the probe: the sealed self join and the seal guard fail
- an over-cap refusal taken as an answer: the over-cap guard fails
- the executor keeps the rows of a join the table stopped answering: five guards fail
- the planner never chooses the strategy: the probe count guard fails, with four others
- an unbounded join handed the strategy: the no-limit guard fails on the batch join's error
- a chunk that ends on a matched row reports no boundary: the continuation guard fails

The cherry-picked first commit adds the capped B-tree probe with its boundary unit test.

## Suites

- `cargo nextest run --no-fail-fast` at the previous head: 5698 passed, 3 skipped
- `cargo nextest run --features test-filedb --no-fail-fast` at the previous head: 5689 passed, 2 failed (the two grouped join tests fixed by the grouped path's fallback and the chunk boundary), 3 skipped
- after the fix, `--features test-filedb` on `join_equality_candidates_test`, `grouped_join_stream_test`, `join_limit_shortfall_test`, `join_residual_cache_test`: 24 passed
- `cargo test --test join_equality_candidates_test --features test-failpoints -- --test-threads=1`: 10 passed
- clippy clean; the full suites at this head are left to CI

## Follow-ups

- Stage 2: the same contract answered from the side files for sealed rows.
- The planner chooses the strategy from the index alone, so a limited join on a table holding volumes fetches one outer chunk before the first probe sends it to the hash path.


semihalev:
Round 1 fixes are in as one commit on top of a0e072ef.

- P1, grouped join key: the grouped operator is built with the inner key column, so a fetched row whose key differs from the probe's is dropped, as in the streaming join. Guard: inside a transaction an order moved to another user is not charged to its old user; after commit it counts under the new one. Checked to fail without the key: the old user gets 40 and 1390. One thing the guard shows and this PR does not change: inside the transaction the moved order is not yet found under its new user, because the hot index takes the new key at commit. That is the memory engine's existing index join behaviour, not something the table probe introduced, and it is reported here rather than fixed.
- P2, allocation under the lock: the operator sizes its id buffer to the cap when the strategy is a table probe, before the first probe. No guard test; the capacity is set in the constructor.
- P2, EXPLAIN: the limit the executor pushes is computed once before the strategy is chosen (`pushed_join_limit`), the planner takes the table probe only for a bounded join, the grouped join passes its plan's bound, and EXPLAIN asks the same rule of the statement (`join_is_bounded`). The executor's later rejection is gone, so there is one rule. Guard: EXPLAIN on a hot file table names Index Nested Loop with LIMIT, and not without a limit or under ORDER BY plus LIMIT; checked to fail with the planner ignoring the bound, together with the no-limit execution guard.

Runs: `join_equality_candidates_test` 12 passed under `test-failpoints`, 5 under `test-filedb`; the touched memory targets (`grouped_join_stream_test`, `join_limit_shortfall_test`, `join_residual_cache_test`, `explain_test`, `explain_plan_test` and this file) 80 passed; clippy clean. Full suites stay with CI.

On the measurement: agreed that the alternating runs and the setup INSERT numbers do not show the concurrent writer's cost. The stage 2 PR will carry a rerun on its final head in contiguous per binary blocks with allocation counts per join and a writer running beside the join loop; this PR's numbers stay as the provisional hot join result.

semihalev:
Round 2, one commit on top of 0f002dd6: 5ef6a836.

- P2, EXPLAIN eligibility: the rule is no longer the bound alone. The grouped operator's shape checks are one function (`grouped_index_join_applies`: INNER or no right-side filter, plain aggregates the registry knows without expressions, FILTER, ORDER BY or hidden ones, a plain GROUP BY over columns, no subquery in the filters), asked by the operator before it takes the join and by EXPLAIN with the same arguments, the right-side filter and the subquery flag taken from the same partition of the WHERE clause. EXPLAIN now decides `JoinIndexUse`: the streaming join only when nothing aggregates or windows the rows, bounded or not; the grouped join only for its shape and only on the right side, since it does not swap; otherwise no index join at all, whatever the PK or index would allow.
- Guard: on a hot file table, `SUM(o.amount) ... GROUP BY u.id LIMIT 1` reports Index Nested Loop, `SUM(o.amount + 1)` under the same GROUP BY does not, and `SUM(o.amount) OVER ()` with LIMIT does not. Checked to fail with EXPLAIN taking the reduction plan for the grouped shape: the expression case reported the index join, on the PK of the swapped side.

Runs: `join_equality_candidates_test` 12 passed under `test-failpoints`; `explain_test`, `explain_plan_test`, the grouped and shortfall join targets and the lib tests 2193 passed in memory, the join and secondary targets 72 passed under `test-filedb`; clippy clean. Full suites stay with CI.

semihalev:
Round 3, one commit on top of 5ef6a836: 29b88bc8.

- P2, correlated WHERE: `join_index_use` turns the shape away first when the WHERE has a correlated subquery, as execution does, so EXPLAIN reports no index join for it. Guard: a correlated `EXISTS` with LIMIT on a hot file table is not reported as Index Nested Loop; checked to fail with the check removed.
- The allocation note: the shape rule now hands the parsed aggregates and group columns to the grouped operator instead of dropping them, so a grouped join parses its shape once.

Runs: `join_equality_candidates_test` 14 passed under `test-failpoints`, clippy clean. Full suites stay with CI.

semihalev:
Round 4, one commit on top of 29b88bc8: 327088f3.

- P2, correlated HAVING: the rejection moved from the grouped operator's entry into the shared shape rule, so EXPLAIN and execution turn the statement away at the same point. Guard: EXPLAIN on a hot file table reports no Index Nested Loop for `GROUP BY u.id HAVING EXISTS (correlated) LIMIT 1`; checked to fail with the check removed from the rule.

Runs: the EXPLAIN guards pass, the grouped join targets 22 passed, clippy clean. Full suites stay with CI.

PR #213, opened 2026-09-21
Answer a limited join's probes for sealed rows from the volumes' side files, beside the hot index, under the scan's per-volume decision

semihalev:
Stage 2 of the join plan, on top of #212: a limited join's probes are answered for sealed rows from the volumes' side files, beside the hot index for the rows not sealed yet. The base is #212's branch; this PR shows only the stage 2 commit once #212 merges.

## The contract, extended

- `SegmentedTable::equality_candidates` takes the cold view first, then asks every volume for the key through the decision the scan already makes per volume (`side_decision`): zone map and bloom prune, side file coverage under the index's current identity, the directory's estimate and the boundary pages' count, the scan share rule, and the cache admission of the reader. A volume the side file cannot serve, or a walk the ledger refuses, leaves the probe unanswered (`None`) and the join goes to the hash path once, as in stage 1.
- The ids come from the volume's resident row id column at the positions the walk names, so the probe reloads no volume; the fetch reloads what it reads, through the group cache.
- The hot index then answers for the unsealed rows under the cap left by the volumes. A seal between the cold view and the hot read moves the generation and the probe is dropped, the stage 1 rule.
- An updated row keeps its old copy in a volume until compaction, so its id can come from a volume and from the hot index; the ids are sorted and deduplicated and the fetch reads the newest copy. A deleted row's id is dropped by the fetch through its tombstone, as in the scan.
- `PRAGMA INDEX_READ_STATS` counts the probes, candidates and served rows of the join like those of a scan.
- The probe is fallible: a page that fails its check is the statement's error, as the scan's is; only a volume the side file cannot serve, a reader the ledger refuses or an index that moved across the probe answers `None`. The schema generation is checked around the cold view.
- One reader per volume per probe, built on a space the operator keeps across probes (`ProbeScratch`, `IndexFile::reader_in`, `Reader::release_into`), its window the cap left; the space carries the reservation that covers its buffers at their capacity, so the ledger sees the memory for as long as it is held, growth is admitted before it happens, and dropping the space releases both; the scan's decision is split into `side_bounds` and `side_admits`, which read no page, and `side_decision` is those two plus its own reader as before.

## Measurements

The harness probe (untracked): the benchmark's users table, 10,000 rows, seed 42, the benchmark's self join with LIMIT 100, 300 timed runs after 20 warm runs, `--profile fast`, the three binaries run in contiguous blocks of three rounds each, the three modes inside every round. Allocations are counted by a global allocator over the timed loop, per join statement. Then the same join loop runs beside a writer thread inserting new rows in autocommit; the insert latency is measured inside that window. Ranges over the three rounds. `stage1` is #212 at 0f002dd6, `main` is 7af4e585.

| mode | bin | join p50 us | join p90 us | allocs per join | join p50 beside writer | insert p50 beside join | insert p99 beside join | checksum |
|---|---|---|---|---|---|---|---|---|
| memory | main | 14.5 to 15.8 | 15.4 to 22.0 | 455 | 29.4 to 30.7 | 2.0 | 5.8 to 6.2 | 20627 |
| memory | stage1 | 14.9 to 15.9 | 15.6 to 17.3 | 455 | 29.2 to 30.6 | 1.8 to 2.0 | 5.2 to 6.0 | 20627 |
| memory | this PR | 14.5 to 15.0 | 15.1 to 15.4 | 455 | 27.7 to 32.1 | 1.8 to 2.1 | 5.6 to 6.6 | 20627 |
| file, rows hot | main | 173.6 to 181.2 | 178.9 to 199.9 | 675 | 385.0 to 438.8 | 3.2 to 3.4 | 7.8 to 10.3 | 20900 |
| file, rows hot | stage1 | 13.8 to 15.9 | 14.3 to 16.2 | 459 | 300.1 to 338.5 | 3.1 to 3.2 | 8.3 to 9.3 | 20627 |
| file, rows hot | this PR | 13.6 to 14.7 | 14.4 to 15.1 | 459 | 267.0 to 289.2 | 3.0 | 8.6 to 8.8 | 20627 |
| file, sealed | main | 597.4 to 615.6 | 615.1 to 636.9 | 10784 | 2501.5 to 2594.7 | 3.8 to 4.1 | 7.6 to 8.2 | 20900 |
| file, sealed | stage1 | 592.5 to 611.4 | 620.1 to 655.5 | 11015 | 2350.9 to 2488.8 | 3.8 to 4.0 | 7.6 to 10.0 | 20900 |
| file, sealed | this PR | 30.1 to 30.8 | 31.6 to 33.5 | 865 | 2225.7 to 2558.5 | 3.6 to 4.0 | 7.3 to 7.5 | 20627 |

- The sealed join answers in about twice the memory engine's time with the memory engine's pairs, and its allocations go from 10,784 per statement on the hash path to 865: the hot join's 459 plus about four per probe over the volume, where the review measured thirteen and two readers' buffers per probe.
- The writer's INSERT latency inside the join window is the same on the three binaries in every mode; the write path is not touched.
- Beside a writer committing every few microseconds the sealed join is the hash path's time on all three binaries: a commit landing across a probe moves the index epoch and the probe is dropped, so the join falls back on nearly every statement. That is the rule this PR keeps from stage 1, and it is the honest cost of it under a continuous writer; the hot rows' join keeps part of its gain there. A bounded retry of a probe that only the epoch moved is the follow-up that would keep the index path under a moderate writer, measured before it is taken.

## Tests

Added to `tests/join_equality_candidates_test.rs`: a sealed order moved to another user counted once under it and an order updated under the same key counted once (the exact values of the review's fixture); a side file page corrupted on disk failing the join with the side index named; a volume sealed before its index refusing the probe as ineligible with the sealed rows joined on the hash path; a sealed table against the memory engine with the read counters showing the side file's probes and rows; sealed rows updated, deleted and joined beside hot ones once each; a key holding too large a share of a volume (`cost_scans`, hash path, right rows); a sealed key past the cap (nothing served, hash path); and under `test-failpoints` a seal landing inside a probe over volumes. The stage 1 tests hold, with the sealed self join now returning the memory engine's pairs and the sealed rows probed after the seal. Mutations checked to fail their guards:

- a read error taken for a probe the table cannot answer: the corrupted side file guard fails
- the grouped join without the key check: both moved-order guards fail
- an ineligible volume skipped instead of refusing the probe: the uncovered volume guard fails
- no re-check after the probe: the commit, the two seal and the truncate guards fail
- a volume the side file cannot serve skipped instead of refusing the probe: the scan share guard fails
- no deduplication: the same-key update guard fails with the row joined twice
- the seal generation not re-checked: the two seal guards and the truncate guard fail

## Suites

- `cargo test --test join_equality_candidates_test --features test-failpoints -- --test-threads=1`: 20 passed; `--features test-filedb` on the same target: 12 passed
- `cargo nextest run --features test-filedb` on `join_equality_candidates_test`, `grouped_join_stream_test`, `join_limit_shortfall_test`, `join_residual_cache_test`, `secondary_index_query_test`, `secondary_side_file_test`, `secondary_index_backfill_test`: 70 passed; the memory run of the join, secondary and EXPLAIN targets: 85 passed
- clippy clean; the full memory and filedb suites at this head are left to CI, as for #212


semihalev:
Round 1 is in: the branch is rebased on #212's round 1 head (0f002dd6) and the stage 2 commit is 2fc97e6e; the PR body carries the new measurements.

- P1, sealed key move: the fix is #212's grouped key check, which this branch now carries. Guard added here with the review's exact fixture: 2,400 orders over 60 users, checkpoint, order 1 moved to user 2 with amount 1000, and order 2 updated under the same key; the grouped join answers (1, 39, 390), (2, 41, 1410), (3, 40, 400). Checked to fail without the key check: user 1 gets 40 and 1390.
- P2, error contract: `Table::equality_candidates` returns a `Result`; a page that fails its check is the statement's error, and only an unsupported volume, a refused reservation or an index that moved across the probe answers `None`. The schema generation is checked around the cold view as the scan checks it. Guard: a side file corrupted on disk after the seal fails the join with the side index named; checked to fail with the error taken for `None`, where the join answered from the hash path instead.
- P2, allocations: the decision reader is gone from the probe path. The scan's decision is split into `side_bounds` and `side_admits`, which read no page, and the probe reads each admitted volume through one reader built on a space the operator keeps between probes (`ReaderSpace` through `IndexFile::reader_in`, returned by `Reader::into_space`), its window the cap left rather than 4,096. The identities go into a kept buffer and the one bloom hash is on the stack. Measured per join statement with a counting allocator over 300 runs: the sealed join goes from 10,784 allocations on the hash path to 865, the hot join's 459 plus about four per probe over the volume. The rest is the cold view and the reservation; I left them as they are.
- The uncovered volume case the split could have hidden got its own guard: a volume sealed before its index refuses the probe as ineligible and the sealed rows come from the hash path; checked to fail with the volume skipped.

Measurement redone as asked: the three binaries in contiguous blocks, three rounds each, allocation counts per join, and a writer thread inserting in autocommit beside a second join loop with the insert latency taken inside that window. INSERT p50 and p99 beside the join are the same on main, #212 and this head in every mode. One honest number in there: beside a writer committing every few microseconds the sealed join runs at the hash path's time on all three binaries, because a commit across a probe moves the index epoch and the probe is dropped, statement after statement. That is the rule from stage 1 doing what it says; a bounded retry of a probe only the epoch moved is the follow-up that would keep the index path under a moderate writer, to be measured before it is taken.

Runs: `join_equality_candidates_test` 20 passed under `test-failpoints` and 12 under `test-filedb`; the join, secondary index and side file targets 70 passed under `test-filedb`, 85 in memory; clippy clean. Full suites stay with CI; the PR is retargeted to main once #212 merges.

semihalev:
Round 2: rebased on #212's 5ef6a836, one commit on top of the stage 2 commit: 0ff7fc26.

- P1, the reservation: it moves with the buffers now. `ReaderSpace` carries the reservation that covers its buffers at their capacity; `IndexFile::reader_in` takes buffers and reservation into the reader, and `Reader::release_into` hands both back, so the ledger sees the memory for as long as the operator holds it and dropping the space releases both. A reader that needs more than the space holds has the growth admitted through `try_reserve` before the buffers grow, and the charge follows what the buffers own at their capacity (`reserve_exact`, with any allocator surplus charged on top), so a smaller window later shrinks nothing. A fresh `reader` is built the same way on an empty space. The allocation reuse stays: nothing is allocated after the first probe of a pass.
- Guard, a unit test on a built side file: the charge after `release_into` equals the active reader's; a window of 1 on the same space changes neither buffers nor charge; a second space under a budget the first one fills is refused with nothing reserved; a window of 4096 grows the charge before the buffers grow; dropping the space returns the ledger to its baseline. Checked to fail with the reservation dropped at release: the charge fell to zero while the buffers stayed.

The EXPLAIN round is #212's commit, carried here by the rebase. Runs: `join_equality_candidates_test` 21 passed under `test-failpoints`, the lib unit test passes, the join and secondary targets 72 passed under `test-filedb`, 2193 in memory with the lib tests; clippy clean. Full suites stay with CI. Next I take the failed COMMIT cold row and HNSW rollback on a branch from main, then the memory join's own updates inside a transaction, then memory TRUNCATE and the HNSW seal-to-fetch check, in that order.

semihalev:
Round 3: rebased on #212's 29b88bc8, one commit on top of the reservation commit: the head is d060d34c.

- P2, repeated-probe admission: the probe's admission asks for the space the reader will still need, the window's bytes less what the operator's kept space already holds, instead of a fresh reader's worth; the scan's own decision keeps asking for a full reader as before. Guard: under a budget with room for one reader's space and its pages, a four-key join over a sealed volume is served from the side file for every key with no refusal; checked to fail with the admission asking for a whole reader, where the second key was refused.

Runs: `join_equality_candidates_test` 22 passed under `test-failpoints`, clippy clean. Full suites stay with CI.

semihalev:
Round 4: rebased on #212's 327088f3; head 8ccc8118.

- P2, the process-wide budget in the test: the tests that read side files take one lock in this file, and the lowered budget is restored by a drop guard when the test ends, whichever way it ends. No production change in this round.

Runs: `join_equality_candidates_test` 22 passed under `test-failpoints`, 13 in the plain run. Full suites stay with CI.

semihalev:
Round 5: 8ba5e988. The three tests that seal and then read a side file (the hot self join checked after its seal, the sealed order moved to another user, and the seal inside the probe) take the lock too, so every side file reader in the file is under it. The budget guard is unchanged. Test-only change; `join_equality_candidates_test` 22 passed under `test-failpoints`.

PR #215, opened 2026-09-21
Take back everything a commit applied when its marker fails: the versions, the row count, the cold tombstones and the sealed rows' index keys

semihalev:
A commit applies its versions and index updates before it writes the WAL marker, and the audit found a sealed row the failed UPDATE or DELETE had touched gone from the open database until reopen, the HNSW graph without its old vector, and a commit that failed on a later table left visible. This PR takes back everything a failed commit applied.

## The fix, as it stands at 104ef4d2

- Cold tombstones stay pending through the marker: `publish_pending_tombstones` promotes them with the commit sequence once the marker is durable, `discard_pending_tombstones` drops them when the commit fails.
- The commit records what each version's application left for its undo (`Applied`: row, created, the version displaced when the chain dropped its history), and `undo_publication` takes the index updates back in reverse, restores each row through `unpublish_version` (the version before is the head again, in the arena too; a created row leaves; the committed row count is what it was), and releases the rows the transaction claimed, whatever handles to its stores are still held.
- A commit that fails on any table aborts as if no table had applied: undo, tombstones dropped, transaction rolled back; the WAL records written so far carry no marker.
- A sealed row replaced or deleted through the hot store always carries the volume's copy as the version it replaces, marked sealed (`mark_sealed_original`): the indexes that keep sealed rows drop its old keys at commit and take them back on undo, the hot indexes treat the row as new. Every cold delete path keeps it (by id, by key, and the scan), so a vector index created between the statement and the commit, named or chosen for the column, has the old vector to take back. Cost: one row copy per sealed row replaced, which the statement had read already, and one row decode per sealed row deleted; not timed here.

A failed WAL write poisons the WAL until reopen, so the statements that follow a failure in the guards run after one.

## Tests

`tests/failed_commit_sealed_rows_test.rs` under `test-failpoints` (in the CI failpoint step), 7 guards: sealed and hot UPDATE and DELETE under a failing WAL sync keep the rows visible through the scan, the key and the aggregates, now and after reopen; a row updated twelve times then failed keeps the last update; a unique key the failed commit would have moved is still taken and the new one free; a sealed vector replaced, replaced by NULL, or deleted by id, range, another column or no predicate stays in the graph; a vector index created between the statement and the failing commit, by USING HNSW and by the column's type, keeps the old vector; a commit that fails on its second table leaves the first as it was with the counts and the freed key, before and after reopen; a failed commit frees its rows while table handles are held. A unit test walks a row through twenty-five updates each followed by a delete taken back.

Mutations checked to fail their guards: tombstones promoted before the marker; tombstones published on failure; the sealed original not recorded; the scan delete's original dropped; the displaced version ignored; the versions left at the heads; the failing table's undo left out; the claims left to the drop.

## Suites

- `cargo test --test failed_commit_sealed_rows_test --features test-failpoints -- --test-threads=1`: 7 passed; `failpoint_io_test` 35 passed
- `cargo nextest run` on `hnsw_lifecycle_test`, `vector_test`, `transaction_index_residual_test`, `seal_fence_dml_test`: 77 passed
- clippy clean; the full suites are left to CI

## Not in this PR

- A table without an integer primary key updates a sealed row under a new row id, so its old HNSW node is never removed at commit; a separate finding.


semihalev:
Round 1 is one commit on top of c7b5ee74.

- P1, quadratic batch record: both application functions return one `Applied` record per version in order (row, created, displaced), and the commit keeps that list as is; no search.
- P1, dropped history: the application hands back the version it displaced when the chain passed its limit, and `unpublish_version` restores from it when the chain holds no predecessor. Guards: the unit test walks a row through twenty-five updates each followed by a delete taken back, so every chain depth including the drops is covered; the SQL guard runs twelve hot updates and then a failing UPDATE or DELETE. Checked to fail with the displaced version ignored.
- P1, general DELETE: the scan delete path keeps the sealed original as the by-id and key paths do. The vector guard now runs the failing statement over `id = 1`, `id >= 1`, `k = 1` and no predicate for DELETE, and over `id = 1` and `k = 1` for UPDATE, checking the graph itself; checked to fail with the scan path's original dropped. On your note that the graph does not read MVCC visibility: agreed, and the PR text is corrected; the tombstone finding stands for the rows, the sealed original for the graph.
- P1, partial commit sequence: the branch reads the commit sequence before `complete_commit` and passes it into `publish_pending_tombstones`, which now takes the sequence from the caller everywhere. The publication follows the marker whatever it returns there, since the tables are already visible; the marker's error is then returned. No guard exercises this branch; it is source-checked only, as in the review.
- Cost notes: the old row is read and kept only when an index of the table keeps sealed rows (`keeps_sealed_rows_in_an_index`, HNSW today), so a table without a vector index pays no row copy on a cold update or delete; and `holds_hot_rows` no longer counts a sealed original as a hot row.

Runs: the failpoint guard file 4 passed under `test-failpoints`, the unit test passes, the six mutations each fail their guard, clippy clean. Full suites stay with CI.

semihalev:
Round 2, one commit on top of 069f8dc2.

- P1, partial application visible before a failed marker: the branch that completed a commit after a later table failed is gone. A commit that fails on any table aborts: the tables that applied take their publication back (`undo_publication`, index updates and versions), the cold tombstones are dropped, the transaction is rolled back, and the WAL records written so far carry no marker, so recovery drops them too. Guard: two sealed tables with a unique key, the transaction updates both to 999, and after the first table publishes another transaction takes 999 in the other; the commit fails, both tables show their rows as they were, COUNT and SUM agree, the key is free wherever the other transaction did not take it, and reopen shows the same. Checked to fail with the undo left out: the first table's aggregates showed the aborted update.
- P1, a vector index created between the statement and the commit: the transaction decides at the statement, by the indexes the table has then, whether to keep a sealed row's old keys, so an index created after that has nothing to take back. A vector index is now not created over a table with a volume while a transaction holds a sealed row's replacement or deletion (`TableHasActiveTransactions`, as a TRUNCATE answers); once that transaction is done the creation goes through. That is the smallest protocol that keeps the statement-time decision honest and costs the write path nothing. Guard: the creation is refused over a pending sealed update and delete, succeeds after the commit, and the graph reads as the write left it; checked to fail with the refusal removed.
- The history guard note: the unit loop covers every depth; the SQL guard stays at twelve updates as a shape check.

Runs: the guard file 6 passed under `test-failpoints` with `failpoint_io_test` 35; `transaction_index_residual_test`, `seal_fence_dml_test`, `hnsw_lifecycle_test`, `vector_test` 77 passed; clippy clean. Full suites stay with CI.

semihalev:
Round 3, one commit on top of 51ae2ea8: 104ef4d2. The PR body now describes the implementation as it stands.

- P1, the creation protocol: the refusal is gone, and with it the statement-time decision it guarded. The transaction keeps the sealed row's old copy on every replacement or deletion, so the index set at commit may differ from the set at the statement and the undo always has the old keys. That closes both cases: an index chosen for the column, and a write that starts after any check, since there is no check. Cost on the write path: one row copy per sealed row replaced (the statement had read it) and one row decode per sealed row deleted; I did not time it. Guard: the index is created between a sealed update or delete and the failing commit, with USING HNSW and by the column's type, and the graph keeps the old vector; checked to fail with the sealed original not recorded.
- P2, the claims: `undo_publication` releases the transaction's row claims itself, and the phase that fails on a table runs it for every table whether or not one had applied. Guard: two table handles held across a failed commit while the next statement updates the rows they claimed; checked to fail with the release left to the drop.

Runs: the guard file 7 passed under `test-failpoints`, `failpoint_io_test` 35, the vector and transaction targets 77; clippy clean. Full suites stay with CI.

semihalev:
Round 4, one commit on top of 104ef4d2: 17ec32ff.

- P2, repeated decodes: the scan delete path builds one reader per volume and reads every deleted row's old copy through it, so its pins keep the groups decoded across the rows; the by-id and key paths keep a reader across the rows that share a volume, the way the fetch does. Guard: 256 sealed rows of two columns deleted with no predicate and with a range under a group cache of zero bytes decode at most four groups; checked to fail with a reader per row, where it decoded 512.
- The other costs noted stand as documented: a pinned row per deleted row while the statement runs, and the old row's copy on a cold update.

Runs: the guard file 8 passed under `test-failpoints`, clippy clean. Full suites stay with CI.

semihalev:
A correction to the round 4 note, and the rebase.

- The decode guard I posted as "checked to fail with a reader per row" had not failed at that point: it lowered the group cache through `PRAGMA GROUP_CACHE_MB = 0`, which does not starve the cache the way a one-byte budget does, so the reader-per-row variant passed it too. The guard now sets the decoded group cache's budget to one byte through its public API, restored by a drop guard, and the reader-per-row variant fails it with 512 decodes for 256 rows while the branch decodes at most four. The claim stands now, with the run behind it.
- The branch is rebased on main after #212 and #213 merged: head fc5b8b2d. The conflicts were the trait and hot table methods added beside `secondary_index_identities_into` and the CI failpoint line; resolved to keep both sides.

Runs after the rebase: the guard file 8 passed under `test-failpoints`; `join_equality_candidates_test`, `hnsw_lifecycle_test`, `transaction_index_residual_test` 26 passed; clippy clean. Full suites stay with CI.

semihalev:
CI failed two unit tests of the volume table (`cold_rows_added_in_any_order_are_found_by_id`, `test_delete_tombstones_cold_row`): their plain hot store does not record a sealed original, and the trait's default answered with an error. One commit on top, e834cce5: the default takes the copy and drops it, since a store that keeps no original has no old keys for the undo to put back. The lib unit tests pass locally (2096); the guard file 8 passed; clippy clean.

semihalev:
Round 6 at the new head.

- `DELETE ... RETURNING` (and every DELETE that scans its rows first: a WHERE the storage cannot take, a referencing foreign key) now deletes its keyed rows in one `delete_by_row_ids` call instead of one `delete(pk = k)` per row, as the rows of a table without a key already did. An integer key is the row id. The originals are read through the readers that call keeps, so the decodes follow the groups touched: `DELETE FROM t WHERE id >= 1 RETURNING id` over 256 sealed rows under a one-byte group cache decodes 5 groups (the scan's 2, the originals' 2, one more), down from 515. The guard `a_sealed_delete_decodes_each_group_once` runs that statement too, bound 6; under the mutation that puts the per-row delete back it reads 515.
- The direct `delete_by_row_ids` over ids alternating between two volumes: the statement now keeps one reader per volume instead of one reader replaced on every volume change. New guard `a_delete_across_two_volumes_decodes_each_volume_once`: 256 rows over two volumes, decodes bounded by 4; under a mutation that clears the readers on every change it reads 512.
- The `expect("reader just set")` is gone with the single reader.

Not changed: a DELETE that scans first still decodes each group twice, once for the scan and once for the originals; the scan's rows are the transaction's visible versions, and the original recorded must be the volume's copy, so the second read stays.

Targets run: `failed_commit_sealed_rows_test` under `test-failpoints` 9 passed, `returning_atomicity_test`, `foreign_key_test`, `delete_without_primary_key_test`, `statement_atomicity_test`, `bugs2_regression_test` 112 passed, clippy clean. Full suites on CI. The PR body is refreshed to this head.

semihalev:
Rebased on main after #216 and #217 merged (only the CI failpoint line met them); the three failpoint targets (`failed_commit_sealed_rows_test` 9, `join_sees_own_updates_test` 4, `truncate_walk_and_seal_fetch_test` 3) pass on the rebased head, clippy clean.

semihalev:
Round 7 at the new head.

- P1: `Table::delete_scanned_rows(&mut Vec<i64>)` replaces `delete_by_row_ids` on the scan-first path. The hot store deletes the ids it still holds a visible version of and leaves the rest alone (no delete marker for an id gone since the scan); the segmented table deletes the sealed rows its statement snapshot still holds and hands the rest to the hot store the same way. On return the vector holds the ids deleted, and RETURNING returns those rows alone. Guard `a_scanned_delete_leaves_out_a_row_another_transaction_deleted_meanwhile`: a new hook, `after_delete_rows_scanned`, fires between the scan and the delete; another transaction's DELETE of a hot row (memory) and of a sealed row plus a hot row (file) commits there. RETURNING leaves them out and the table ends empty. Three mutations fail it: the scanned ids all counted as deleted, the executor returning every scanned row, the per-row delete put back (the decode guard fails too).
- P2: the ids of a scan-first or by-id delete are grouped by volume and each volume's sealed rows go through one reader dropped before the next volume's (`delete_cold_rows`), so one volume's decoded groups are pinned at a time in whatever order the ids come. The decode guards hold (5 for RETURNING, 4 across two volumes). The peak in the one-row-per-volume probe should come back to the parent's; I have no allocation guard for it, so the probe is the check.

Targets run: `failed_commit_sealed_rows_test` under `test-failpoints` 10 passed, `returning_atomicity_test`, `foreign_key_test`, `delete_without_primary_key_test`, `statement_atomicity_test`, `bugs2_regression_test` 112 passed, clippy clean. Full suites on CI. The PR body is refreshed.

semihalev:
Round 8 at the new head: an id named twice is one row. The segmented delete dedups its located sealed ids after sorting them by volume and position, and the hot store's delete skips an id it has seen in the same call, so `delete_by_row_ids([1, 1])` deletes the row once and reports 1, sealed or hot. Guard `a_row_named_twice_is_deleted_once` (a sealed row and a hot row, each named twice); the mutation that lets both duplicates through fails it. `failed_commit_sealed_rows_test` 11 passed, the DELETE targets 31, clippy clean; the PR body's Tests list names the guard.

semihalev:
Round 9 at the new head: the hot store builds the seen-id set only when the call names more than one id, so the cold DELETE's one-id calls allocate nothing for it; a batch keeps the dedup. The guard `a_row_named_twice_is_deleted_once` holds and its mutation still fails it; `failed_commit_sealed_rows_test` 11 passed, clippy clean. The allocation count per deleted row is yours to confirm with the probe.

semihalev:
Round 10: the twice-named row guard takes the shared failpoint guard before its fixture, as the file's other tests do, so another test's WAL failpoint cannot reach it under the shared-process runs. Target 11 passed, run without a thread limit.

PR #216, opened 2026-09-22
Let an index join inside a transaction find the rows the transaction moved under a key or inserted with it

semihalev:
Inside a transaction, an index join found a row under the key it had when the transaction began and not under the one the transaction gave it, and did not find a row the transaction inserted, while a plain read of the same table saw both. The shared index takes the new key at commit; the join asked only the index. The materialized CTE join read the table through a transaction of its own and did not check the fetched rows' keys.

## The fix, as it stands at bfeb5948

- `Table::local_column_values(column, out)` hands over the `column` value and id of every uncommitted row of the transaction whose value there is not the one the shared index holds for it: a row inserted, or moved under the value. A row that kept its key is the index's already. The hot store answers from its transaction-local versions and their read versions; a segmented table asks its hot store.
- The streaming and the batch index join take that once at `open`, keyed by value in one map, when the inner table has local changes and the strategy probes a shared index. A probe appends the ids under its key as they are; nothing per probe is searched or allocated. The primary key strategy takes nothing: the fetch of its row id sees the local version. Memory: one map per operator holding the transaction's changed keys and ids for the inner column, key values cloned, plus a temporary pair vector while it is built.
- The fetched row's key is checked as before, so a row moved away from a key is dropped there; the grouped index join now checks it too (the same one-line fix as in #212; it merges either way).
- The materialized CTE join reads the table through the statement's transaction, as the plain join does, and builds its operator with the inner key.

## Tests

`tests/join_sees_own_updates_test.rs`, 3 guards: inside a transaction an order is moved to another user with a new amount, one is inserted and one deleted; the grouped join charges the moved order to its new user only, the bounded and the unbounded joins agree, the inserted order is joined under its user, the materialized CTE join over the index and over the key answers the same, and the answers hold after commit; a transaction moves two thousand rows under one key and the join finds each once beside the seven the index held; a transaction updates the payload of a whole key and the join finds each row once with the new value. Mutations checked to fail them: the streaming join not asking for local rows, the batch join not asking, the CTE join on its own transaction, the grouped join not checking the key, the rows that kept their key handed over too.

## Suites

- `cargo nextest run` on `join_sees_own_updates_test`, `join_residual_cache_test`, `cte_test`, `grouped_join_stream_test`: passed; `--features test-filedb` on the first: passed
- clippy clean; the full suites are left to CI


semihalev:
Round 1, one commit on top of dfc8059a.

- P1, the CTE join: it reads the table through the statement's transaction (`join_table` on the statement snapshot, as the plain join does) instead of a transaction of its own, and builds its operator with the inner key, so the fetched row's key is checked. Guard: inside the transaction and after commit, `WITH x(uid) AS (SELECT 2)` joined on the index finds the moved order and drops the deleted one, `WITH x(oid) AS (SELECT 1)` on the key returns the moved order's new amount, and `SELECT 2` on the key returns nothing. Checked to fail with the CTE join back on its own transaction.
- P2, the primary key probe: the supplement is built only for a strategy that probes a shared index; the primary key strategy takes nothing, since the fetch of its row id sees the local version.
- P2, the quadratic dedup: `local_column_values` hands the operator the transaction's rows with their key once, at open; the operator keys them in a map, and a probe appends the ids under its key after a search of the index's ids alone, not the growing buffer. One map per operator, nothing per probe. Guard: a transaction moves two thousand rows under one key and the join finds each once beside the seven the index held.

Runs: `join_sees_own_updates_test` 2 passed, with `join_residual_cache_test`, `cte_test` and `grouped_join_stream_test` 26 passed, 2 under `test-filedb`; the four mutations (streaming supplement gone, batch supplement gone, CTE on its own transaction, grouped key check gone) each fail the guard; clippy clean. Full suites stay with CI.

semihalev:
Round 2, one commit on top of 85366e2a: bfeb5948. The PR body now describes the implementation as it stands.

- P2, the dedup: the table hands the operator only the rows whose value in the column is not the one the index holds for them (a row inserted, or moved under the value); a row that kept its key is the index's already, so the operator appends the local ids as they are, with no membership search. A probe costs its matches. Guard: a transaction updates the payload of every row under one key and the join finds each once with the new value; checked to fail with the unchanged rows handed over too. The map's memory cost is stated in the body.

Runs: the guard file 3 passed; the five mutations each fail a guard; clippy clean. Full suites stay with CI.

semihalev:
Round 3, one commit on top of round 2, and the branch rebased on main after #212 and #213 merged: head a463f3a6 (the conflicts were in the operator's fields and the grouped join, resolved to main's versions).

- P1, the snapshot: the read version is no longer taken as proof of what the index holds. The operator takes, once at open, the ids of every row the transaction wrote and its live rows by key; a probe drops the written rows from what the index names for the key and adds the transaction's rows under the probed key, so the transaction's version answers for its rows exactly once, whatever key the index holds them under. One set and one map per operator, nothing per probe beyond the retain over the index's answer. Guard: a snapshot transaction reads a row, another transaction moves it and commits, the first updates it with the moved key and, separately, with the key it read, and joins on both keys through the plain join and the CTE join; checked to fail with the index's answer kept beside the transaction's (three of four guards fail).

Runs: the guard file 4 passed, with `join_equality_candidates_test`, `cte_test`, `grouped_join_stream_test` 34 passed after the rebase; the five mutations each fail a guard; clippy clean. Full suites stay with CI.

